PR #10, opened 2020-11-22
fix: correctly expose route and middleware types

danielroe:
* allows arrays of routes/handlers to be passed in (**types**)
* this exposes Middleware (Handler with `next()`) rather than just Handler if `promisify` has not been set to false (**types**)


src/promisify.ts
```diff
@@ -12,7 +12,7 @@ function callHandle (handle: Middleware, req: IncomingMessage, res: ServerRespon
     const next = (err?: Error) => err ? reject(err) : resolve(undefined)
     try {
       const returned = handle(req, res, next)
-      if (returned !== undefined) {
+      if (returned !== undefined || handle.length !== 3) {
```
pi0, line 15: https://github.com/nuxt-contrib/h2/issues/11#issuecomment-731783698

PR #14, opened 2020-11-23
feat: add REST API helpers

danielroe:
:heavy_check_mark: Makes for very minimal API construction

:x: Downside is the slight increase in size: `0.3Kb`

Happy for this to be closed. It's also trivial for an end-user to add this to the app instance.


pi0:
Only downside is see with this is creating a wrapper function to filter out method. Also it wouldn't be easy for specify in nuxt config for example.

We would support it with layer.match() (and { method: 'get' / app.get }

pi0:
Closing since PR is stalled. Also opened #18 to investigate method filtering.

PR #15, opened 2020-12-10
feat: add request and response utils

farnabaz:
This PR 
- creates `useQuery` util to parse the query params of request.
- creates `useBody` util to parse raw and JSON payload of the request.
- Modify `createError` util to support runtime error to support runtime errors like validation errors.


src/app.ts
```diff
@@ -80,7 +80,10 @@ export function createHandle (stack: Stack): PHandle {
       }
     }
     if (!res.writableEnded) {
-      throw createError(404, 'Not Found')
+      throw createError({
+        statusCode: 404,
+        statusMessage: 'Not Found'
```
Atinux, line 85: I would prefer to see `message` instead.
I would also vote for `status` instead of `statusCode` 😄

Atinux: But it can makes sense to keep them like this, cc @pi0

src/utils.ts
```diff
@@ -68,3 +80,26 @@ export function useBase (base: string, handle: PHandle): PHandle {
     return handle(req, res)
   }
 }
+
+export function useQuery (req: IncomingMessage): URLSearchParams {
+  const url = new URL(req.url || '', 'http://localhost')
+  return url.searchParams
```
Atinux, line 86: I am not found of the `searchParams` object honesty, quite hard to loop on it, I would prefer the object syntax if possible, cc @pi0

src/app.ts
farnabaz: The idea is to be consistent with the http response object but `status` and `message` is much simpler

pi0: Indeed. Idea of h2 is keeping framework as much as consistent with node Http (https://nodejs.org/api/http.html#http_response_statuscode)

src/types.ts
```diff
@@ -47,3 +47,10 @@ export interface App {
 export interface AppOptions {
   debug?: boolean
 }
+
+export interface RuntimeError {
+  statusCode: number
+  statusMessage: string
+  body?: any
+  runtime?: boolean
```
pi0, line 55: If interface is called `RuntimeError` maybe makes sense to set type to `runtime: true` ?

src/utils.ts
```diff
@@ -26,22 +27,33 @@ export function sendError (res: ServerResponse, error: Error | string, code?: nu
...
     console.error(error) // eslint-disable-line no-console
   }
 
   // @ts-ignore
   res.statusMessage = res.statusMessage || error.statusMessage || error.statusText || 'Internal Error'
 
-  res.end(`"${res.statusMessage} (${res.statusCode})"`)
+  // @ts-ignore
+  let body = error.body || `"${res.statusMessage} (${res.statusCode})"`
+  if (typeof body === 'object') {
```
pi0, line 40: undefined
```suggestion
  if (typeof body === 'object') {
```

src/utils.ts
```diff
@@ -26,22 +27,33 @@ export function sendError (res: ServerResponse, error: Error | string, code?: nu
...
+  if (debug && !error.runtime && res.statusCode !== 404) {
     console.error(error) // eslint-disable-line no-console
   }
 
   // @ts-ignore
   res.statusMessage = res.statusMessage || error.statusMessage || error.statusText || 'Internal Error'
 
-  res.end(`"${res.statusMessage} (${res.statusCode})"`)
+  // @ts-ignore
+  let body = error.body || `"${res.statusMessage} (${res.statusCode})"`
```
pi0, line 39: suggestion:
```js
  let response = { statusMessage, statusCode, body: !debug && error.body }
```

src/utils.ts
```diff
@@ -68,3 +80,26 @@ export function useBase (base: string, handle: PHandle): PHandle {
...
+export function useBody<T> (request: IncomingMessage): Promise<T> {
+  return new Promise<T>((resolve, reject) => {
+    const body: any[] = []
+    request.on('error', (err) => {
+      reject(err)
+    }).on('data', (chunk) => {
+      body.push(chunk)
+    }).on('end', () => {
+      const bodyString = Buffer.concat(body).toString()
+      if (request.headers['content-type'] === 'application/json') {
```
pi0, line 98: I would suggest split this part to `useJSON` / `getJSON` and destr for parsing

src/utils.ts
```diff
@@ -68,3 +80,26 @@ export function useBase (base: string, handle: PHandle): PHandle {
...
+
+export function useBody<T> (request: IncomingMessage): Promise<T> {
+  return new Promise<T>((resolve, reject) => {
+    const body: any[] = []
+    request.on('error', (err) => {
+      reject(err)
+    }).on('data', (chunk) => {
+      body.push(chunk)
+    }).on('end', () => {
+      const bodyString = Buffer.concat(body).toString()
```
pi0, line 97: We might keep `req.body` for second calls of `useBody`

src/utils.ts
```diff
@@ -68,3 +80,26 @@ export function useBase (base: string, handle: PHandle): PHandle {
     return handle(req, res)
   }
 }
+
+export function useQuery (req: IncomingMessage): URLSearchParams {
+  const url = new URL(req.url || '', 'http://localhost')
+  return url.searchParams
+}
+
+export function useBody<T> (request: IncomingMessage): Promise<T> {
```
pi0, line 89: I would suggest renaming to `getBody` or `readBody`

src/utils.ts
pi0: Agreed. We might rename current utility to `useParams`/`getParams` and a new `useQuery`/`getQuery` and both may be moved to ufo (only exporting request wrapper from here)

src/types.ts
farnabaz: In this case, `createError` will be dedicated to runtime errors. For internal errors, users must use other methods?  
Should rename the interface to a more generic name, like `HttpError`

pi0: Indeed since internal errors require inspecting original error object to get status, etc. We might have `InternalError` type and both extending `HttpError`

src/utils.ts
farnabaz: I don't get the idea of `!debug && error.body`?

pi0: There might be valid use cases for client to consume error body but it might potentially leak traces in production. We might do something like `(debug || !internal) ? error : 'internal error')` at least

src/utils.ts
farnabaz: You are suggesting to move headers check to another util? or just the parsing?
In the first scenario, what is the use-case of `useJSON`?

pi0: Parsing body with destr regardless of headers (internally using useBody)

src/utils.ts
farnabaz: manipulating response structure on runtime errors is not a good idea, because applications depend on the response.  
I think this would be a good way to prevent trace leak
```js
let body: any = {
    statusCode: res.statusCode,
    statusMessage: res.statusMessage
}
if (debug || error.runtime) {
    body = error.body || body
}
``

src/utils.ts
farnabaz: Just created `getQuery` util to return object.
For the sake of consistency, I do use `get` prefix for all utils

pi0: It was just a suggestion for query or utils... Since for body we actually read body. For consistency use* sounds better

Atinux:
Would it be possible to open another PR for documenting them @farnabaz ?

Atinux:
Also, shouldn't we rename them to getBody(req) that actually use destr by default and having getRawBody() instead?

pi0:
Usage of body might be ambiguous for different responses but might rename useJSONBody to useJSON for shorter name if it is better :)

farnabaz:
@Atinux I'll create PR for docs.  
What about creating a docs project?

Atinux:
I think README is enough for now, cc @pi0

pi0:
Yeah. H2 is pretty much an experimental project now and i don't want to put energy on docs for in development utils. We may later on benefit jsdocs

danielroe:
@pi0 It may be worth starting off with jsdoc comments as these can be more easily kept up-to-date with changes to API? I am happy to do that if you agree.

pi0:
@danielroe Both yes and no. h2 is not GA public. Signature and usage of utils may change and it costs maintenance still is better than maintaining markdown...

pi0:
JSDocs seems better that would be nice if you are up to add them :)

PR #16, opened 2020-12-14
docs: add jsdoc for error, body and some utils

farnabaz:


Atinux:
Thanks @farnabaz 

While we are at this, I would like to suggest to rename `useBody` to `useBodyRaw` and having the parsing with `destr` by default for `useBody`.

Do you agree @pi0 ?

README.md
```diff
@@ -103,6 +103,74 @@ When calling `app.use`, middleware will be automatically promisified.
...
+`h2` provides multiple utility functions.
+
+### `useQuery`
+
+Parse query parametes of the request and returns equivalent object.
+
+```js
+import { useQuery } from '@nuxt/h2';
+
+app.use('/api', (req) => ({ queryParams: useQuery(req) }))
```
Atinux, line 117:
```suggestion
app.use('/api', (req) => ({ query: useQuery(req) }))
```

I will also show a quick url example:

/api?hello=world&array[]=one&array=two will return `{ hello: 'world', array: ['one', 'two] }`

pi0 (changes requested):
This is partial documentation which might be misleading. We shall include docs for all utils (like including sendError, promisify, etc)  or not. As suggested in other PR we may use JSDocs to maintain update docs with utils

Atinux:
I see, fully agree with JSDocs for automated docs 🔥

pi0:
> While we are at this, I would like to suggest to rename useBody to useBodyRaw and having the parsing with destr by default for useBody.

Makes sense

farnabaz:
I've started using jsdoc and added comments to some functions. It is not completed yet.  
The result is available here: http://h2-docs.surge.sh/

pi0:
@farnabaz Kindly please stop doing actions without consulting repo maintainer via issues!! This time I revert unnecessary changes but next time will simply close PR. Thanks.

PR #17, opened 2020-12-15
feat: add `useCookie` and `setCookie` utils

farnabaz:
I've added two utility function to work with cookies.  
- `useCookie`: Read and parse cookies from request
- `useCookieWrite`: it receives the response object and return a function to manipulate response header and set `Set-Cookie` header.


src/cookie.ts
```diff
@@ -0,0 +1,67 @@
...
+    .split(';')
+    .map((item) => {
+      const [key, value] = item.trim().split('=')
+      return [key, decodeURI(value)]
+    })
+
+  const cookieObject = Object.fromEntries(cookieEnteries)
+
+  // @ts-ignore
+  req.cookie = cookieObject
```
danielroe, line 35: I think we can add an TS assertion that extends the Request object once this function has been called...

src/cookie.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  const { cookie } = req.headers
+  const cookieEnteries = (cookie || '')
+    .split(';')
+    .map((item) => {
+      const [key, value] = item.trim().split('=')
+      return [key, decodeURI(value)]
+    })
+
+  const cookieObject = Object.fromEntries(cookieEnteries)
```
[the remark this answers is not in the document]
farnabaz: Thank you :)

src/cookie.ts
pi0: Cache value shouldn't be exposed or implicitly directly accessed.

PR #25, opened 2021-06-23
fix: createError fallback to statusMessage

antfu:
Currently, when using Nuxt, the createError does not generate meaningful error messages if only `statusMessage` is provided (which I see is more commonly been used in the Nuxt code base)

image

I would think it could be better to serve `statusMessage` if `message` is not provided. Ideally, maybe we could have better format as ``` `${input.statusCode}: ${input.statusMessage}` ``` but I will leave it for discussions

### DIFF
--- a/src/error.ts
+++ b/src/error.ts
@@ -28,7 +28,7 @@ export function createError (input: Partial<H3Error>): H3Error {
     return input
   }
 
-  const err = new H3Error(input.message)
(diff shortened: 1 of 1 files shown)


danielroe (approved):

pi0 (approved):

pi0:
Thanks!

PR #41, opened 2021-10-14
perf: minify json returned from `h3` handlers

danielroe:
resolves https://github.com/nuxt/framework/issues/1027

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -120,7 +120,7 @@ export function createHandle (stack: Stack): PHandle {
         } else if (val instanceof Error) {
           throw createError(val)
         } else {
-          return send(res, JSON.stringify(val, null, 2), MIMES.json)
+          return send(res, JSON.stringify(val), MIMES.json)
         }
       }
     }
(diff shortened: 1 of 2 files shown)


src/error.ts
```diff
@@ -83,5 +83,5 @@ export function sendError (res: ServerResponse, error: Error | H3Error, debug?:
   }
 
   res.setHeader('Content-Type', MIMES.json)
-  res.end(JSON.stringify(responseBody, null, 2))
```
pi0, line 86: Likewise, we can support the second argument for formatting and pass `debug` flag in app when calling (by default format enabled should be better IMO)

danielroe: Let's leave this one pretty as it's unlikely to make much difference with errors, then.

src/app.ts
```diff
@@ -120,7 +120,7 @@ export function createHandle (stack: Stack): PHandle {
         } else if (val instanceof Error) {
           throw createError(val)
         } else {
-          return send(res, JSON.stringify(val, null, 2), MIMES.json)
```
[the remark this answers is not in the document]
danielroe: You can tell that I use httpie by default 😆

pi0 (approved):

PR #54, opened 2021-12-10
feat: add params matching

yassilah:
Hi!
Based on that discussion #4 and on that issue [#872](https://github.com/nuxt/framework/issues/872) on the nuxt/framework repo, I thought I'd give a try to add params matching on h3 to allow for nuxt server API with params, too 🙂

Please discard it if that's not how you wanted to proceed, it's a pleasure to contribute to awesome packages anyway.

PS: I also added some tests and made sure the other tests all passed but I'm not 100% confident my implementation is the most appropriate because I'm a bit confused as to the reason why `req.url` is being reduced to `/`?

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,5 +49,8 @@
     "ts-jest": "latest",
     "typescript": "latest",
     "ufo": "latest"
+  },
+  "dependencies": {
+    "path-to-regexp": "^6.2.0"
   }
 }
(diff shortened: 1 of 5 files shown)


pi0:
Thanks so much for the PR @yassilah and sorry it took so long to move forward. Adding new routing support via #64 . See https://github.com/unjs/h3/issues/4#issuecomment-1063352360 for why not using matching system.

yassilah:
no problem at all, i'm sure you're making the right move 🙂

Issue #45, opened 2021-11-18
Respond with a stream of data


educkf:
Hello!

Is there a way already to respond a request with a stream of data?

I am planning specifically to use puppeteer [page.createPDFStream](https://pptr.dev/#?product=Puppeteer&version=v11.0.0&show=api-pagecreatepdfstreamoptions), as Firebase Functions don't give us much initial memory, and I only found a way to increase it via regular cloud function, and not using nuxt `/server/api` folder.

So, probably in this case, stream the data would be a good fit.
Thanks a lot!

harlan-zw:
related: https://github.com/unjs/h3/pull/68 ?

PR #55, opened 2021-12-20
feat: can respond to streams

Achrou:
Hi!
For issue #45, return null in the handle to prevent the default response
```
app.use('/stream', async (req, res) => {
  const result = await fetch('test.mp4')
  result.body.pipe(res)
})
```

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -108,14 +108,13 @@ export function createHandle (stack: Stack, options: AppOptions): PHandle {
         continue
       }
       const val = await layer.handle(req, res)
-      if (res.writableEnded) {
+      if (res.writableEnded || !val) {
         return
       }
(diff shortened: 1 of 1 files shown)


src/app.ts
```diff
@@ -108,14 +108,13 @@ export function createHandle (stack: Stack, options: AppOptions): PHandle {
         continue
       }
       const val = await layer.handle(req, res)
-      if (res.writableEnded) {
+      if (res.writableEnded || !val) {
```
pi0, line 111: Falsy values can be also valid json (0 and null)

pi0:
Thanks for this PR @Achrou. I believe we could find better ways to explicitly return a stream and let h3 do the pipe operation.

Issue #35, opened 2021-10-05
async handle did not wait for closing


xhebox:
Found while experimenting with nuxt3.

In my nuxt app, h3 will use a lazy/promisify/async middleware handle, which will eventually be invoked in `callHandle`:

https://github.com/unjs/h3/blob/2805d4cc42d22c22c7798a41514aca5cceeb8e19/src/handle.ts#L18-L21

But async function will return a promise instance, if async handler does not send/close the response before they return, `callHandle` resolve immediately.

The following patch is helpful on my case:

```
--- a/node_modules/h3/dist/index.mjs
+++ b/node_modules/h3/dist/index.mjs
@@ -109,7 +109,7 @@ function callHandle(handle, req, res) {
     const next = (err) => err ? reject(err) : resolve(void 0);
     try {
       const returned = handle(req, res, next);
-      if (returned !== void 0) {
+      if (returned !== void 0 && !(returned instanceof Promise)) {
         resolve(returned);
       } else {
         res.once("close", next);
```

Suspect it is root cause of #21.

pi0:
In latest (0.8+) version of h3, you should explicitly convert legacy middleware using `h3.fromNodeMiddleware((req, res) => {})`.

PR #56, opened 2021-12-22
feat: handle asyncs returning without closing res

xhebox:
refer the analysis in #35

What I am doing in my handler:

```ts
const app = fastify()

export default async function (req: unknown, res: unknown) {
    await app.ready()
    app.server.emit('request', req, res)
}
```

### DIFF
--- a/src/handle.ts
+++ b/src/handle.ts
@@ -20,7 +20,7 @@ export function callHandle (handle: Middleware, req: IncomingMessage, res: Serve
     const next = (err?: Error) => err ? reject(err) : resolve(undefined)
     try {
       const returned = handle(req, res, next)
-      if (returned !== undefined) {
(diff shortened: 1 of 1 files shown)


Atinux:
Thanks for the PR.

Would you mind adding a test for it?

pi0:
You can do the async initialization using `lazyHandle`:

```js
import { lazyHandle } from 'h3'

export default lazyHandle(async () => {
  const app = fastify()
  await app.ready()
  return (req, res) => {
    // ... 
  }
})
```

xhebox:
> ```js
> lazyHandle
> ```

I could check it that works, later. But I think h3 should handle the case by default, if possible.

xhebox:
closed since it is outdated

PR #65, opened 2022-03-10
fix: allow returning, number and boolean as well

### LINKED ISSUE #48: is it necessary?
image

1. Have been confirmed type is object
2. In my opinion，val !== void 0 is not necessary

There are other applications for this part of the code ？

Pray for god's guidance~~~

danielroe:
resolves #48

This enables us to send a response for:

* boolean
* number
* bigint

We may wrongly return a response for functions & symbols. Though likely that's a user error rather than something h3 should check for.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -114,7 +114,7 @@ export function createHandle (stack: Stack, options: AppOptions): PHandle {
       const type = typeof val
       if (type === 'string') {
         return send(res, val, MIMES.html)
-      } else if (type === 'object' && val !== undefined) {
+      } else if (val !== undefined) {
         // Return 'false' and 'null' values as JSON strings
         if (val && val.buffer) {
           return send(res, val)
--- a/src/app.ts
+++ b/src/app.ts
@@ -114,12 +114,7 @@ export function createHandle (stack: Stack, options: AppOptions): PHandle {
       const type = typeof val
       if (type === 'string') {
         return send(res, val, MIMES.html)
-      } else if (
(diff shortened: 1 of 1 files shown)


pi0:
It would be better if we check already resolved type against JSON-safe values rather than assuming all values as response

src/app.ts
```diff
@@ -114,7 +114,7 @@ export function createHandle (stack: Stack, options: AppOptions): PHandle {
       const type = typeof val
       if (type === 'string') {
         return send(res, val, MIMES.html)
-      } else if (type === 'object' && val !== undefined) {
+      } else if (['bigint', 'number', 'boolean', 'object'].includes(type)) {
```
pi0, line 117: For maximum performance, we could avoid iteration and add 4 `==` checks.

pi0:
Removed bigint. JSON.parse seems unable to handle bigint by default (https://stackoverflow.com/questions/18755125/node-js-is-there-any-proper-way-to-parse-json-with-large-numbers-long-bigin)

src/app.ts
```diff
@@ -114,7 +114,11 @@ export function createHandle (stack: Stack, options: AppOptions): PHandle {
       const type = typeof val
       if (type === 'string') {
         return send(res, val, MIMES.html)
-      } else if (type === 'object' && val !== undefined) {
+      } else if (
+        type === 'object'
```
danielroe, line 118:
```suggestion
        type === 'object' ||
```

PR #66, opened 2022-03-11
feat: add `deleteCookie` utility

danielroe:


src/utils/cookie.ts
```diff
@@ -42,3 +42,20 @@ export function setCookie (res: ServerResponse, name: string, value: string, ser
...
+/**
+ * Set a cookie value by name.
+ * @param res {ServerResponse} A ServerResponse object created by [http.Server](https://nodejs.org/api/http.html#http_class_http_server)
+ * @param name Name of the cookie to delete
+ * @param serializeOptions {CookieSerializeOptions} Cookie options
+ * ```ts
+ * deleteCookie(res, 'Authorization')
+ * ```
+ */
+export function deleteCookie (res: ServerResponse, name: string, serializeOptions?: Omit<CookieSerializeOptions, 'maxAge' | 'encode' | 'decode' | 'expires'>) {
```
pi0, line 55: It can be convenient allowing to pass these options even if not picked (to reuse same cookieOptions in different places)

src/utils/cookie.ts
```diff
@@ -42,3 +42,20 @@ export function setCookie (res: ServerResponse, name: string, value: string, ser
...
+  const cookieStr = serialize(name, '', {
```
pi0, line 56: Any reason to not reuse `setCookie` logic?

src/utils/cookie.ts
```diff
@@ -42,3 +42,19 @@ export function setCookie (res: ServerResponse, name: string, value: string, ser
...
   appendHeader(res, 'Set-Cookie', cookieStr)
 }
+
+/**
+ * Set a cookie value by name.
+ * @param res {ServerResponse} A ServerResponse object created by [http.Server](https://nodejs.org/api/http.html#http_class_http_server)
+ * @param name Name of the cookie to delete
+ * @param serializeOptions {CookieSerializeOptions} Cookie options
+ * ```ts
+ * deleteCookie(res, 'Authorization')
```
pi0, line 52: (Authorization as an example is little misleading. it is usually for a header :D) 
```suggestion
 * deleteCookie(res, 'SessionId')
```

PR #68, opened 2022-03-14
feat: add stream pipe response

stafyniaksacha:
This allows to pipe readable streams returned from middlewares!

I also set the default `content-type` header used for buffers and streams to `text/html` assuming most of time it would be used to pipe html response. We could use [`application/octet-stream`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#applicationoctet-stream) instead, but it would require to explicitly set `content-type` on each responses.


danielroe:
Very nice thought. However, it looks like this adds a hard dependency on node, which we would want to avoid. Maybe something other than `instanceof` as a test?

stafyniaksacha:
> Very nice thought. However, it looks like this adds a hard dependency on node, which we would want to avoid. Maybe something other than `instanceof` as a test?

Hum, that's true. Using deno or web workers would rely on [ReadableStream](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream) which is the new standard. 

This uses the [former node streams](https://nodejs.org/docs/latest-v16.x/api/stream.html). Instead we can rely on the new [stream implementation](https://nodejs.org/docs/latest-v16.x/api/webstreams.html), so we can check over `pipeTo` method instead of `instanceof`.

The caveat is that the feature is new and require node v16, what do you think?

src/app.ts
```diff
@@ -115,8 +115,8 @@ export function createHandle (stack: Stack, options: AppOptions): PHandle {
       if (type === 'string') {
         return send(res, val, MIMES.html)
       } else if (type === 'object' || type === 'boolean' || type === 'number' /* IS_JSON */) {
-        if (val && val.buffer) {
-          return send(res, val)
+        if (val && (val.buffer || isStream(val))) {
+          return send(res, val, MIMES.html)
```
pi0, line 119: Html mime is probably not a safe assumption. Stream responses could be binary also for sending assets

pi0: Perf: We already know data is stream when calling send. isStream check could be removed for send function

stafyniaksacha: - For the html mime, we can explicitly set to `application/octet-stream` by default (which is better than undefined), what do you think? 
- For the `isStream` call, if we don't, it will run in the else branch and the content will be stringified

src/utils/response.ts
```diff
@@ -41,3 +42,17 @@ export function appendHeader (res: ServerResponse, name: string, value: string):
...
   res.setHeader(name, current.concat(value))
 }
+
+export function isStream (data: any) {
+  return typeof data === 'object' && typeof data.pipe === 'function' && typeof data.on === 'function'
+}
+
+export function sendStream (res: ServerResponse, data: any) {
+  return new Promise((resolve, reject) => {
+    defer(() => {
```
pi0, line 52: Using `defer` was mostly a performance hack for Node.js writable. I guess with pipes, we can start without it and see if it can really improve perf. (You can try with `profile` script if interested to investigate)

stafyniaksacha: > Using `defer` was mostly a performance hack for Node.js writable. I guess with pipes, we can start without it and see if it can really improve perf. (You can try with `profile` script if interested to investigate)

It does not change anything, but I will remove it as it an unnecessary operation. Would you like I add a profile:stream script ?

PR #70, opened 2022-03-21
chore: remove jest.config.js 

harlan-zw:
h3 was migrated to vitest but there's still a jest config file that isn't used as far as I could see


Atinux (approved):

pi0:
Thanks

PR #72, opened 2022-03-21
chore(refactor): extracting handle sending logic

harlan-zw:
## **Why?**

The purpose of this PR is to expose to external packages the logic for inferring a handles return value to the send promise. Specifically for my https://github.com/harlan-zw/unrouted package I had copy + pasted the logic for this, I'd like to benefit from upstream fixes with this change (such as the stream addition).

It also provides a more testable API if further mime type inferences were to be made.

## **Side effects**

**The order in which the responses are inferred has changed.** This was for readability, tests are passing so we can assume there should be no logic issues however would be good to review. Not too sure about performance considerations.

**Edge case fix: HTML content type for primitives** Not something I'd expect to come up outside of an edge case, but if a user wanted to explicitly return a number that should be HTML, they would need to cast it as a string (see tests).

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -116,7 +116,8 @@ export function createHandle (stack: Stack, options: AppOptions): PHandle {
         return
       }
       /*
-       * An inferred send is attempted on the return value from this routes handler.
+       * Given the return value from the layer handler, we try and guess the most applicable response to send.
+       * For instance, if the response is an object, we send stringified JSON with the appropriate header.
        */
       const sendPromise = maybeSendInferredResponse(res, val, { jsonSpacing: spacing })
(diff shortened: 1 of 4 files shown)


harlan-zw:
Conflict is resolved and inline code documentation added :)

Atinux:
Do we have any way to benchmark this difference in a reliable manner?

harlan-zw:
Thanks for looking it over and good point @pi0 

I think the performance overheads and pkg size is important to keep minimal, happy to close off this PR to avoid any degradation. 

I have some other ideas on how to integrate the packages better which I think will be a better solution than this anyway, so will focus on that

PR #80, opened 2022-03-30
fix: initialise `res.req`

danielroe:
otherwise `res.req` is undefined

### DIFF
--- a/src/event.ts
+++ b/src/event.ts
@@ -95,9 +95,7 @@ export function createEvent (req: http.IncomingMessage, res: http.ServerResponse
   // @ts-ignore
   res.res = res
   // @ts-ignore
-  res.req.res = res
-  // @ts-ignore
-  res.req.req = req
+  res.req = req
 
   return event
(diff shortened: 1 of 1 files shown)


src/event.ts
```diff
@@ -95,9 +95,7 @@ export function createEvent (req: http.IncomingMessage, res: http.ServerResponse
   // @ts-ignore
   res.res = res
   // @ts-ignore
-  res.req.res = res
-  // @ts-ignore
-  res.req.req = req
+  res.req = req
```
pi0, line 98: Node.js standard `res` has a `req` property attached to which is a little bit different than the incoming `req`! 

I think this is originating from unenv mocks that `res.req` is unset.

danielroe: maybe we can just check if res.req exists then? (and possibly also resolve in unenv mock)

pi0: Or we could add `res.req` polyfill if not existing :)

PR #84, opened 2022-04-04
chore(CHANGELOG): remove link to emoji svg inside "handler"

Jclong98:
a link to the following emoji (😐) looks like it was accidentally pasted in-between the word "handler"


pi0:
Thanks 😬

Issue #93, opened 2022-04-11
How to stream files


maciej-jezierski:
Hi, I am trying to stream files from local disc (for a reason it's not served as static).

I tried to use sendStream or send it directly to pipe but each time I get 404. Documentation is not mentioning too much how to use streams. How to use it correctly with h3?

```
export default defineEventHandler((event) -> 
     const stream = fs.createReadStream(path);
     stream.pipe(event.res);
)
```

or 

```
const stream = fs.createReadStream(path);
      sendStream(event, stream);
```

maciej-jezierski:
I found solution by using playground and checking code. Solution is super simple, sendStream has to be returned directly 

```
const r = "/path/to/file
return sendStream(event, fs.createReadStream(r))
```

In my case if you have promise first it's working as well

```
return $fetch(apiUrl, { method: "GET", parseResponse: JSON.parse }).
    .then(r => sendStream(event, fs.createReadStream(r.path)))
```

KISS, but generally would be good to add it to documentation before some other people run into this issue as streaming files is common exercise.

Atinux:
Would love to see a PR to https://v3.nuxtjs.org/guide/features/server-routes#handling-requests-with-body to add an example about it 💚

maciej-jezierski:
@Atinux I've even prepared code for h3 Readme.md and Nuxt repo as well with samples, but apparently I have some access problem. Will try to push it later.

maciej-jezierski:
Nuxt PR: https://github.com/nuxt/framework/pull/4401

pi0:
@Atinux We should probably start adding examples directly to h3 docs. Otherwise, we will end up with fragmented documentation and outdated info for utils. Also, consider `sendStream` support is not universal yet.

manuel-84:
how to stream text from a variable and not in a file?

PR #101, opened 2022-04-16
Docs: add sample to playground with file stream

maciej-jezierski:
Based on #93 issue description


pi0:
Hi @maciej-jezierski. Thanks for PR and sorry it got stalled this long. We will probably add examples for utils at some point in docs but for now, `sendStream` is only listed with it's signature as others. Also need to find a way to reduce dependency on Node.js more (https://github.com/unjs/h3/issues/73).

PR #105, opened 2022-04-20
fix(types): nest arrays in handler response

### LINKED ISSUE #104: [TS][Types] Type error when returning deep nested data in defineEventHandler
https://stackblitz.com/edit/github-7pwass?file=server/api/test.ts

replygirl:
fix #104

### DIFF
--- a/src/event.ts
+++ b/src/event.ts
@@ -12,7 +12,7 @@ export interface H3Event {
 
 export type CompatibilityEvent = H3Event | IncomingMessage
 
-type _JSONValue<T=string|number|boolean> = T | T[] | Record<string, T>
+type _JSONValue<T=string|number|boolean> = T | T[] | Record<string, T | T[]>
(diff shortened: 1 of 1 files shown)


replygirl:
using this, i'm realizing nested objects also aren't accounted for

replygirl:
i got started on supporting nested objects, but it's a little more involved

the [`type-fest` `JSONValue`](https://github.com/sindresorhus/type-fest/blob/main/source/basic.d.ts) will validate against constants just find, but comparing types requires the user to wrap the compared type in [`Jsonify`](https://github.com/sindresorhus/type-fest/blob/main/source/basic.d.ts) or else their input will be rejected

`export default defineEventHandler<Jsonify<T>>(...)` is working for me, but obviously it's not desirable to make the user pass their return type

leaving this PR up so you can take a look, i imagine you'll need to refactor a bit to take `Jsonify` out of the picture

PR #119, opened 2022-05-11
feat: add `H3Event`, `H3Response` and `H3Headers`

danielroe:
https://github.com/unjs/h3/issues/81

See also https://github.com/nuxt/framework/issues/5718.

### DIFF
--- a/src/event.ts
+++ b/src/event.ts
@@ -1,22 +1,15 @@
 import type http from 'http'
 import type { IncomingMessage, ServerResponse, Handler, Middleware } from './types'
 import { callHandler } from './handler'
-
-export interface H3Event {
-  '__is_event__': true
-  event: H3Event
-  req: IncomingMessage
-  res: ServerResponse
-  context: Record<string, any>
-}
+import { MIMES } from './utils'
 
 export type CompatibilityEvent = H3Event | IncomingMessage
 
-export type H3Response<T = any> = T | Promise<T>
+export type HandlerResponse<T = any> = T | Promise<T>
 
-export interface EventHandler<T=any> {
+export interface EventHandler<T = any> {
   '__is_handler__'?: true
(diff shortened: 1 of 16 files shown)


src/event.ts
```diff
@@ -77,47 +70,182 @@ export function toEventHandler (handler: CompatibilityEventHandler): EventHandle
...
 }
+
+export class H3Headers implements Headers {
+  _headers: Record<string, string>
+
+  constructor (init?: HeadersInit) {
+    if (!init) {
+      this._headers = {}
+    } else if (Array.isArray(init)) {
+      this._headers = Object.fromEntries(init.map(([key, value]) => [key.toLowerCase(), value]))
```
SegaraRai, line 163:
```suggestion
      this._headers = init.reduce((acc, [key, value]) => {
        const lk = key.toLowerCase()
        acc[lk] = acc[lk] != null ? `${acc[lk]}, ${value}` : value
        return acc
      }, {})
```

I think it is better to use `reduce` because duplicate keys may be specified.

```ts
const headers = new Headers([['Example', 'foo'], ['Example', 'bar']])
// headers.get('Example') === 'foo, bar'
```

SegaraRai:
This is a problem with the fetch spec itself, not this PR, but the Headers API has a problem with setting multiple `Set-Cookie` headers. whatwg/fetch#973
As mentioned in the referenced issue, there is currently no fixed method to deal with this, but Cloudflare Workers, for example, has implemented a `getAll` method that accepts only `Set-Cookie` as an argument to deal with this.
https://developers.cloudflare.com/workers/runtime-apis/headers/#differences
I think it is not uncommon to specify multiple `Set-Cookie`s, so we should be able to store the individual values in `H3Headers` in some way and pass them to `ServerResponse`.

PR #124, opened 2022-05-25
feat: add `H3EventContext` for augmentation

### LINKED ISSUE #122: Allow typing event context
**Question: How to define context typescript type?**

Hey, not sure if this is an h3 question or a lack of typecript knowledge from me.

I am using Nuxt 3 and am trying to add types to my server handlers. I have a simple server handle:

```ts
export default defineEventHandler((event) => { // event here is a H3Event object
    const result = event.context.container.resolve()
    return {
      result
    }
})
```
The event here is a H3Event:

```ts
interface H3Event {
    '__is_event__': true;
    event: H3Event;
    req: IncomingMessage;
    res: ServerResponse;
    context: Record<string, any>;
}
```
And I have a middleware where I do the following:

```ts
export default defineEventHandler((event) => {
  event.context.container = {
    resolve: () => {
      return 'ok'
    }
  }
})
```
I would like to "type" the event parameter, I would like my VSCode editor to be aware that the event parameter contains a context property that contains a resolve property that is a function that returns a string.

I tried extending the H3Event using the following:

```ts
declare module 'h3' {
  interface H3Event {
    context: Record<string, any> & {
      resolve: () => string
    };
  }
}
```

But Typescript gives me the following error:

> Subsequent property declarations must have the same type. Property 'context' must be of type Record<string, any>...

Is there a way to define the context type of the event parameter in a server middleware?

danielroe:
Simply exposes an `EventContext` interface for manual typing:

```ts
declare module 'h3' {
  interface EventContext {
    myProperty?: boolean
  }
}
export {}
```

resolves #122

### DIFF
--- a/src/event.ts
+++ b/src/event.ts
@@ -2,12 +2,14 @@ import type http from 'http'
 import type { IncomingMessage, ServerResponse, Handler, Middleware } from './types'
 import { callHandler } from './handler'
 
+export interface EventContext extends Record<string, any> {}
+
 export interface H3Event {
   '__is_event__': true
   event: H3Event
   req: IncomingMessage
   res: ServerResponse
-  context: Record<string, any>
+  context: EventContext
 }
 
 export type CompatibilityEvent = H3Event | IncomingMessage
--- a/src/event.ts
+++ b/src/event.ts
@@ -9,7 +9,7 @@ export interface H3Event {
   event: H3Event
   req: IncomingMessage
   res: ServerResponse
-  context: EventContext
+  context: H3EventContext
 }
 
 export type CompatibilityEvent = H3Event | IncomingMessage
--- a/src/event.ts
+++ b/src/event.ts
@@ -2,7 +2,7 @@ import type http from 'http'
 import type { IncomingMessage, ServerResponse, Handler, Middleware } from './types'
 import { callHandler } from './handler'
 
-export interface EventContext extends Record<string, any> {}
+export interface H3EventContext extends Record<string, any> {}
 
 export interface H3Event {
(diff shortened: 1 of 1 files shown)


src/event.ts
```diff
@@ -2,12 +2,14 @@ import type http from 'http'
 import type { IncomingMessage, ServerResponse, Handler, Middleware } from './types'
 import { callHandler } from './handler'
 
+export interface EventContext extends Record<string, any> {}
```
pi0, line 5:
```suggestion
export interface H3EventContext extends Record<string, any> {}
```

src/event.ts
```diff
@@ -2,12 +2,14 @@ import type http from 'http'
...
+
 export interface H3Event {
   '__is_event__': true
   event: H3Event
   req: IncomingMessage
   res: ServerResponse
-  context: Record<string, any>
+  context: EventContext
```
pi0, line 12:
```suggestion
  context: H3EventContext
```

pi0:
(I'm not sure but kinda feeling using `H3EventContext` would be less confusing in the future about naming interface with `H3` prefix.)

wobsoriano:
Can we also add augmentation for the request object? This is for using connect/express middlewares.

An example with `express-session`:

```ts
// ~/server/middleware/session.ts
import session from 'express-session'
import type { Session } from 'express-session'

export default session(options)

declare module 'h3' {
  interface EventRequest {
    session: Session
    sessionId: string
  }
}
```

```ts
// ~/server/api/user.ts
export default defineEventHandler((event) => {
  event.req.session // typed properly!
})
```

PR #125, opened 2022-05-26
fix(pkg): add `types` to the exports

rinsuki:
in node16/nodenext module mode of TyepScript 4.7+, If package has `exports` entry on package.json, needs "types" field in `exports.*` for find typings.

ref. https://devblogs.microsoft.com/typescript/announcing-typescript-4-7/

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
   "sideEffects": false,
   "exports": {
     ".": {
+      "types": "./dist/index.d.ts",
       "import": "./dist/index.mjs",
       "require": "./dist/index.cjs"
     }


pi0:
Thanks for the PR. I agree to start adding `types` to subpath exports but wondering is top level types totally ignored in new mode? /cc @danielroe

rinsuki:
yes, when node16/nodenext mode, if `types` are not exists in `exports` fields, TypeScript treat as simple JS project (and throws compile error due to missing declaration file), like this:

image

pi0:
Thanks for confirming 👍🏼

PR #137, opened 2022-06-19
added useParams to get parameters from url

notshekhar:


src/utils/request.ts
```diff
@@ -37,3 +37,7 @@ export function assertMethod (event: CompatibilityEvent, expected: HTTPMethod |
     })
   }
 }
+
+export function useParams(event: CompatibilityEvent) {
```
pi0, line 41:
```suggestion
export function useRouteParams(event: CompatibilityEvent) {
```

PR #140, opened 2022-06-26
created a function defineEventHandlers which takes multiple handlers in events.ts 

notshekhar:


pi0:
Thanks for this pull request but as I explained in https://github.com/unjs/h3/issues/127#issuecomment-1164989158, while is possible to nest apps like this to have express-like middleware it is an anti-pattern for h3 apps. Instead, you have to split the code into reusable functions and only export one handlers. I don't think adding this utility to the main package would be the best idea.

PR #143, opened 2022-07-04
fix: ensure `useBase` returns a leading-slash-prefixed url

danielroe:
I'd prefer to resolve upstream in `ufo` but I wasn't sure what the desired behaviour of `withoutBase` should be, so this is more of a conversation starter.

Personally I think it is more intuitive that `/api` doesn't remove itself from `/api-test` but only from `/api/test` and `/api?test`, particularly as the _reverse_ behaviour (`withBase`) joins segments with `/`... If you agree, I'll implement in `ufo` directly, and also ensure that `withoutBase` returns `/`-prefixed paths.

see https://github.com/nuxt/framework/issues/5242


pi0:
Do we still need this?

danielroe:
Nope!

PR #153, opened 2022-07-26
feat: render html that will also redirect to location when static

danielroe:


src/utils/response.ts
```diff
@@ -25,7 +25,9 @@ export function defaultContentType (event: CompatibilityEvent, type?: string) {
 export function sendRedirect (event: CompatibilityEvent, location: string, code = 302) {
   event.res.statusCode = code
   event.res.setHeader('Location', location)
-  return send(event, 'Redirecting to ' + location, MIMES.html)
+  // minimal html document that redirects on client side
+  const html = `<!DOCTYPE html><html><head><meta http-equiv="refresh" content="0; url=${encodeURI(location)}"></head><body></body></html>`
```
pi0, line 29: Can you please also add a visual `<a>` link?

danielroe: I've added it within a `<noscript>` as otherwise I think it is a better experience not to have text flash on the screen and then off again. Let me know if you feel differently.

src/utils/response.ts
```diff
@@ -26,7 +26,7 @@ export function sendRedirect (event: CompatibilityEvent, location: string, code
   event.res.statusCode = code
   event.res.setHeader('Location', location)
   // minimal html document that redirects on client side
-  const html = `<!DOCTYPE html><html><head><meta http-equiv="refresh" content="0; url=${encodeURI(location)}"></head><body></body></html>`
+  const html = `<!DOCTYPE html><html><head><meta http-equiv="refresh" content="0; url=${encodeURI(location)}"></head><body><noscript>Redirecting to <a href=${JSON.stringify(location)}>${location}</a></noscript></body></html>`
```
pi0: I think meta also works in case of scripts being disabled :) What i meant was to preserve current behavior of display redirect link while navigation happens (instead of blank screen)

danielroe: yes, I think it does too. I was thinking more of a search engine. I still think it's a worse UX to have it flash on screen (because it's _meant_ to be an SSR redirect, and at least the white screen suggests it's 'still loading'), but will update.

src/utils/response.ts
pi0: I see but with browser fallback, there is anyway a flash to white...

src/utils/response.ts
```diff
@@ -25,7 +25,13 @@ export function defaultContentType (event: CompatibilityEvent, type?: string) {
 export function sendRedirect (event: CompatibilityEvent, location: string, code = 302) {
   event.res.statusCode = code
   event.res.setHeader('Location', location)
-  return send(event, 'Redirecting to ' + location, MIMES.html)
+  // minimal html document that redirects on client side
+  const html = `<!DOCTYPE html>
+<html>
+  <head><meta http-equiv="refresh" content="0; url=${encodeURI(location)}"></head>
+  <body>Redirecting to <a href=${JSON.stringify(location)}>${location}</a></body>
```
pi0, line 32: (just to ensure avoid rare XSS possibility)
```suggestion
  <body>Redirecting to <a href=${JSON.stringify(location)}>${encodeURI(location)}</a></body>
```

pi0 (approved):

PR #154, opened 2022-07-27
feat!: send 204 response if null is returned from handler

danielroe:
https://github.com/unjs/h3/issues/152

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -133,8 +133,11 @@ export function createAppEventHandler (stack: Stack, options: AppOptions) {
         return send(event, val, MIMES.html)
       } else if (isStream(val)) {
         return sendStream(event, val)
+      } else if (val === null) {
+        event.res.statusCode = 204
+        return send(event)
       } else if (type === 'object' || type === 'boolean' || type === 'number' /* IS_JSON */) {
-        if (val && (val as Buffer).buffer) {
+        if (type === 'object' && val.buffer) {
           return send(event, val)
         } else if (val instanceof Error) {
           throw createError(val)
--- a/test/app.test.ts
+++ b/test/app.test.ts
@@ -19,6 +19,15 @@ describe('app', () => {
     expect(res.body).toEqual({ url: '/' })
   })
 
+  it.only('can return a 204 response', async () => {
+    app.use('/api', () => null)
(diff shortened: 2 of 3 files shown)


src/app.ts
```diff
@@ -133,8 +133,11 @@ export function createAppEventHandler (stack: Stack, options: AppOptions) {
         return send(event, val, MIMES.html)
       } else if (isStream(val)) {
         return sendStream(event, val)
+      } else if (val === null) {
+        event.res.statusCode = 204
+        return send(event)
       } else if (type === 'object' || type === 'boolean' || type === 'number' /* IS_JSON */) {
-        if (val && (val as Buffer).buffer) {
+        if (type === 'object' && val.buffer) {
```
pi0, line 140: Type assertion above, should avoid errors (only null is exception)

```js
// [ undefined, undefined, undefined, undefined, undefined ]
[{}, true, false, 123, Infinity].map(val => val.isBuffer)
```

test/app.test.ts
```diff
@@ -19,6 +19,23 @@ describe('app', () => {
...
+    expect(res.statusCode).toBe(204)
+    expect(res.text).toEqual('')
+    expect(res.ok).toBeTruthy()
+  })
+
+  it('can return primitive values', async () => {
+    app.use('/number', () => 42)
+    expect(await request.get('/number').then(r => r.body)).toEqual(42)
+
+    app.use('/bool', () => true)
```
pi0, line 35: Is `false` handling broken?

danielroe: nope, my bad!

src/app.ts
danielroe: The type assertion wasn't doing anything as TS only narrows type if typeof is included in each test.

test/app.test.ts
danielroe: Happy to remove test if needed.

pi0: Seems good. I tought `false` as test case has more coverage than true. We can test both too...

danielroe: Ahh. Good point.

pi0 (approved):

pi0:
Marking as semver-minor as null usage is rare.

PR #156, opened 2022-08-01
refactor: deprecate utils with `use*` prefix

pi0:
`use*` prefix is usually used for composables with context. Using more semantic prefixes instead by deprecating old ones.

- `useRawBody` => `readRawBody`
- `useBody` => `readBody`
- `useCookies` => `parseCookies`
- `useCookie` => `getCookie`
- `useQuery` => `getQuery`
- `useMethod` => `getMethod`


danielroe (approved):
Agree - I think this will be much clearer.

antfu (approved):

Issue #123, opened 2022-05-24
setCookie not accept ServerResponse and got a typescript type error


csvwolf:
In the version, response is removed: https://github.com/unjs/h3/commit/b2856598e1432796ce7aadac2be1c11837f766d8

```
setCookie(res, name, value, opts?)
```

cant be called in middleware.( but if I set the type as any, it works)

What's the reason to remove ServerResponse for the api?

NozomuIkuta:
@csvwolf 

h3 has not yet reached to stable version (i.e. v1), so breaking changes are likely to be introduced even in minor updates.

As far as I see, your issue results from #73.

You also might want to check [a author's comment](https://github.com/unjs/h3/issues/149#issuecomment-1204354147), which says that `req/res` style middleware has been deprecated.

pi0:
You should pass event as first argument to all h3 utils, `setCookie(event, ...)`

Issue #73, opened 2022-03-23
Migrating to Events API


pi0:
h3 was born with the idea of being a platform-agnostic server framework for all JavaScript environments and backward compatibility with legacy server middleware format for Node.js and Express compatibility.

Initially, we started with the most backward-compatible format `(req, res)` and improved it by async support `async (req, res)` and also directly returning a response for a pleasant experience with `(req, res) => 'Hello World'` format.

[unjs/unenv](https://github.com/unjs/unenv) provides a [createCall](https://github.com/unjs/unenv/blob/a88e927dcfacc7c1e43ee434894a41c2a03f3b0c/src/runtime/fetch/call.ts#L18) and mocks of [Node.js `HTTP` built-ins](https://github.com/unjs/unenv/tree/main/src/runtime/node). Thanks to this mocking framework, h3 is able to work nicely on almost any serverless or Worker platform.

However, this kind of setup keeps h3 to be away from **natively supporting** non Node.js environments with the cost of unenv overhead for any code not using Node.js.

Additionally, `(req, res)` format is not even suitable to extend h3 supporting Node.js features like [http2](https://nodejs.org/api/http2.html) without compatibility API and `ws` support.

Migration to `(event)` format will be progressive and backward compatibility wherever necessary however we potentially have to introduce breaking changes as semver-minor until reaching `h3@1.0.0`.

### Progress

- [x] opt-in using event format using `defineEventHandler` (#74)
- [x] use Event api with utils with compatibility layer (https://github.com/unjs/h3/pull/75)
- [x] Event class (#81)
- [x] Make node-js handlers opt-in and drop `(req, res)` implicit conversion (#178)
- [x] Make `event.req` and `event.res` optional and adopt more of event native (#231)
- [ ] Mock (with warning) for req, res (#472)
- [ ] Migrate all leftover utils

PhaxeNor:
Would this potentially add support for `waitUntil` in the `fetch` listener? So it could be utilized with Cloudflare Workers lifecycle for instance. (https://developer.mozilla.org/en-US/docs/Web/API/ExtendableEvent/waitUntil)

nathanchase:
So what's the roadblock to get h3 (and Nuxt 3, as a result) to serve files via [http2](https://nodejs.org/api/http2.html) protocol?

Hebilicious:
I manually implemented something like this to play with Request/Response compatible libraries :

```ts
const getRequestFromEvent = async (event: H3Event) => {
[...]

PR #157, opened 2022-08-02
feat: add utilities for http headers

### LINKED ISSUE #147: Add helper function for headers
Since its quite common to read the header strings, it would be nice if h3 could provide a helper function for this, which parses the headers and provides a structured way to access various common headers. For example, inspired by https://hapi.dev/module/accept/api/?v=5.0.2#parseallheaders one could have
```ts
parseHeaders(event).accept.mediaTypes
//  ["text/plain", "text/html", "application/json", "*/*"]
```

NozomuIkuta:
resolves #147

I added the following 3 utility functions as @pi0 suggested:

- `getHeaders(event)` for request
- `getHeader(event, name)` for request
- `setHeader(event, name, value)`

In addition, I added a unit test for `appendHeader()` function.

---

Do we need to make `getHeaders` and `getHeader` available for response?

For your reference, Express.js provides such functions as methods on `req` and `res` objects (i.e. `req.get` and `res.get`), so it's obvious from what object the functions get headers.

There are 2 ways to provide same functionalities with normal functions.

One is to let users pass `event.req` or `event.res`:

- `getHeaders(event.req)` / `getHeader(event.req, name)`
- `getHeaders(event.res)` / `getHeader(event.res, name)`

But `event` may be an `IncomingMessage` so developers must know which to pass to the function.

- `getHeaders(event)` (event is an `IncomingMessage`)

Another way is to provide functions dedicated to `req` and `res`:

- `getReqHeaders(event)` / `getReqHeader(event, name)`
- `getResHeaders(event)` / `getResHeader(event, name)`

### DIFF
--- a/src/utils/request.ts
+++ b/src/utils/request.ts
@@ -1,3 +1,5 @@
+import { IncomingMessage } from 'http'
+import type { IncomingHttpHeaders } from 'http'
 import { getQuery as _getQuery } from 'ufo'
 import { createError } from '../error'
 import type { HTTPMethod } from '../types'
@@ -42,4 +44,17 @@ export function assertMethod (event: CompatibilityEvent, expected: HTTPMethod | 
       statusMessage: 'HTTP method is not allowed.'
     })
   }
+}
+
+export function getHeaders (event: CompatibilityEvent): IncomingHttpHeaders {
+  return event instanceof IncomingMessage
+    ? event.headers
+    : event.req.headers
+}
+
+export function getHeader (event: CompatibilityEvent, name: string): IncomingHttpHeaders[string] {
+  const headers = getHeaders(event)
+  const value = headers[name.toLowerCase()]
+
+  return value
 }
--- a/src/utils/request.ts
+++ b/src/utils/request.ts
@@ -1,5 +1,3 @@
-import { IncomingMessage } from 'http'
-import type { IncomingHttpHeaders } from 'http'
 import { getQuery as _getQuery } from 'ufo'
 import { createError } from '../error'
 import type { HTTPMethod } from '../types'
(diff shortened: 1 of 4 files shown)


NozomuIkuta:
I'm not sure how it is going about #73 (related: #123).
So, if you tell me the details, I will fix this PR as needed.

src/utils/request.ts
```diff
@@ -43,3 +45,16 @@ export function assertMethod (event: CompatibilityEvent, expected: HTTPMethod |
     })
   }
 }
+
+export function getHeaders (event: CompatibilityEvent): IncomingHttpHeaders {
+  return event instanceof IncomingMessage
```
pi0, line 50: CompatibilityEvent always has `.req` and `.res` props. We do not need to handle Node.js `req` directly being passed.

pi0:
Thanks for PR @NozomuIkuta <3 I think you are right. We might better have explicit names for utils operating on request or response. What do you think of these names:

- `getRequestHeaders`
- `getRequestHeader`
- `setResponseHeader`

this way we can also add `getReponseHeader`/`getResponseHeaders`

NozomuIkuta:
@pi0 

Thank you for quick review!

---

> What do you think of these names:
>
> - `getRequestHeaders`
> - `getRequestHeader`
> - `setResponseHeader`
>
> this way we can also add `getReponseHeader`/`getResponseHeaders`

I agree with you, I will add these methods in `src/utils/request.ts` and `src/utils/response.ts`, respectively.

One question is, should we rename `appendHeader` to `appendResponseHeader` for more clarity and consistency?
It's only response object to which we can append headers, so it might be a little bit too obvious and verbose.

If you prefer the longer name, I will deprecated the old name as you did for `useQuery` and so on.

pi0:
`appendResponseHeader` also makes sense. I think we could keep other utils simple as is because header is on the few cases is shared between request and response. 

We might even have shortcuts for common usage (`getHeader` => shortcut for `getRequestHeader`), (`setHeader` => `setResponseHeader`) too.

NozomuIkuta:
@pi0 

I reorganized header utility methods like below:

- `getRequestHeaders(event, headers)` (alias: `getHeaders`)
- `getRequestHeader(event, name)` (alias: `getHeader`)
- `setResponseHeaders(event, headers)` (alias: `setHeaders`)
- `setResponseHeader(event, name, value)` (alias: `setHeader`)
- `appendResponseHeaders(event, headers)` (alias: `appendHeaders`)
- `appendResponseHeader(event, name, value)` (alias: `appendHeader`)

(Note: `headers` is an object of header names and their values)

Node.js native `IncomingMessage` has both `getHeaders` and `getHeader`, so I applied this relationship to the other methods as well.

pi0 (approved):
Looks perfect. Thanks!

Issue #149, opened 2022-07-21
req.context not available for app.use()


renhiyama:
This code works:
```js
const router = createRouter()
 .get('/', () => 'Hello World!')
 .get('/hello/:name', req => `Hello ${req.context.params.name}!`)

app.use(router)
```

But this wont:
```js
app.use("/",(req,res)=>  `Hello ${req.context.params.name}!`)
```

And `app.get` doesnt even exist like expressjs has...
I dont want to create a new router, but straight away use app.use if possible. Thanks.

pi0:
Hi dear @renhiyama. You should use `eventHandler` wrapper to create new handlers. Legacy `(req, res)` is going to be unsupported soon.

```js
import { eventHandler } from 'h3'
app.use("/", eventHandler((event)=>  `Hello ${event.context.params.name}!`))
```

Also @NozomuIkuta made a some new utils to access params. Check #158.

PR #158, opened 2022-08-03
feat: add utility for router params (#120)

### LINKED ISSUE #120: Add useParam(event) like useQuery(event)
To access parameters need to do `event.context.params.name`.
Why not exists `useParam(event)`?

NozomuIkuta:
Resolves #120 

I added `getRouterParams(event)` utility function.
I also added `useRouterParams(event)` for it to be aligned with other API until the next braking change (may be v1?).

### DIFF
--- a/src/utils/request.ts
+++ b/src/utils/request.ts
@@ -10,6 +10,14 @@ export function getQuery (event: CompatibilityEvent) {
 /** @deprecated Use `h3.getQuery` */
 export const useQuery = getQuery
 
+export function getRouterParams (event: CompatibilityEvent): CompatibilityEvent['context'] {
+  // Fallback object needs to be returned in case router is not used (#149)
+  return event.context.params || {}
+}
+
+/** @deprecated Use `h3.getRouterParams` */
+export const useRouterParams = getRouterParams
+
 export function getMethod (event: CompatibilityEvent, defaultMethod: HTTPMethod = 'GET'): HTTPMethod {
   return (event.req.method || defaultMethod).toUpperCase() as HTTPMethod
 }
--- a/src/utils/request.ts
+++ b/src/utils/request.ts
@@ -18,6 +18,15 @@ export function getRouterParams (event: CompatibilityEvent): CompatibilityEvent[
 /** @deprecated Use `h3.getRouterParams` */
 export const useRouterParams = getRouterParams
 
+export function getRouterParam (event: CompatibilityEvent, name: string): CompatibilityEvent['context'][string] {
+  const params = getRouterParams(event)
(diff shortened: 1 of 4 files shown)


src/utils/request.ts
```diff
@@ -10,6 +10,14 @@ export function getQuery (event: CompatibilityEvent) {
 /** @deprecated Use `h3.getQuery` */
 export const useQuery = getQuery
 
+export function getRouterParams (event: CompatibilityEvent): CompatibilityEvent['context'] {
+  // Fallback object needs to be returned in case router is not used (#149)
+  return event.context.params || {}
+}
```
NozomuIkuta, line 16: This is a comment for GitHub to add a reference link to #149.

src/utils/request.ts
```diff
@@ -10,6 +10,14 @@ export function getQuery (event: CompatibilityEvent) {
 /** @deprecated Use `h3.getQuery` */
 export const useQuery = getQuery
 
+export function getRouterParams (event: CompatibilityEvent): CompatibilityEvent['context'] {
```
pi0, line 13: Can we please export it from `src/router.ts` or `src/utils/router.ts`?

test/utils.test.ts
```diff
@@ -53,6 +53,56 @@ describe('', () => {
     })
   })
 
+  describe('getRouterParams', () => {
```
pi0, line 56: Can we please move it to `test/router.test.ts` ?

pi0:
Nice addition! Would be nice to also introduce `getRouterParam(event, name)` :)

NozomuIkuta:
@pi0 

I added `getRouterParam(event, name)` and `useRouterParam(event, name)`, and reorganized unit tests.

src/utils/request.ts
```diff
@@ -10,6 +10,23 @@ export function getQuery (event: CompatibilityEvent) {
...
+export const useRouterParams = getRouterParams
+
+export function getRouterParam (event: CompatibilityEvent, name: string): CompatibilityEvent['context'][string] {
+  const params = getRouterParams(event)
+
+  return params[name]
+}
+
+/** @deprecated Use `h3.getRouterParam` */
+export const useRouterParam = getRouterParam
```
pi0, line 28: We never introduce `useRouter*` utils. Then don't need to add and deprecate. deprecation of other utils was that they could be used in existing projects.

NozomuIkuta: Deleted! 🙋‍♀️ 

819d133

pi0 (approved):
Amazing! Thank you!

PR #162, opened 2022-08-10
fix: don not log errors when `onError` is provided

danielroe:
Currently we get two copies of error logs if we are logging them in a user-provided `onError` handler. Presumably it makes sense for user to log (or not) if they are providing their own error handler

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -64,13 +64,12 @@ export function createApp (options: AppOptions = {}): App {
         error.unhandled = true
       }
 
-      if (error.unhandled || error.fatal) {
-        console.error('[h3]', error.fatal ? '[fatal]' : '[unhandled]', error) // eslint-disable-line no-console
-      }
-
       if (options.onError) {
         await options.onError(error, event)
       } else {
+        if (error.unhandled || error.fatal) {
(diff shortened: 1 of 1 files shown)


pi0:
This change makes sense but is slighly breaking. We need to update two places at least:

- https://github.com/nuxt/framework/blob/97f0fa511a10c25bcea0618799fe7319639c2c20/packages/nuxt/src/core/runtime/nitro/error.ts#L28 (fatal handler)
-  https://github.com/unjs/nitro/blob/e10f1307d765b0ac35e5492c14f804086569ac91/src/runtime/error.ts#L15 (currently nitro uses no console logs)

danielroe:
I don't think we need to update the framework code. That line is already logging the error to the console.

And I'm probably misunderstanding, but don't we already log in https://github.com/unjs/nitro/blob/e10f1307d765b0ac35e5492c14f804086569ac91/src/runtime/error.ts#L29-L31 ?

pi0:
I missed nitro one but neither are handling `.fatal` flag (nuxt ignores it i guess even)

danielroe:
Good point! In that case it's not breaking, but just incremental improvement to suppressing logging for non-fatal, non-unhandled errors. I'll follow up with PRs to framework + nitro.

pi0 (approved):

PR #173, opened 2022-09-02
fix: use named type imports from `http`

### LINKED ISSUE #169: IntelliSense error in H3Event
### Environment

"nitropack": "0.4.24"

### Reproduction

https://stackblitz.com/edit/github-xukg4p-mpwena?file=routes%2Findex.ts,package.json

### Describe the bug

```ts
export default defineEventHandler((event) => {
  event.res.statusCode = 204;
  event.res.end();
});
```

IntelliSense error:
```
Property 'statusCode' does not exist on type 'ServerResponse'.(2339)
```

Seems like the error is in:

```ts
// node_modules/h3/dist/index.d.ts:1
import http, { OutgoingMessage } from 'http';
```

should be this:

```ts
import * as http from 'http';
import { OutgoingMessage } from 'http';
```

### Additional context

_No response_

### Logs

_No response_

danielroe:
resolves #169

**Note**: By disabling `esModuleInterop`, we will now get the same errors as others do who are using this library without it enabled.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -1,4 +1,4 @@
-import type http from 'http'
+import type { IncomingMessage as NodeIncomingMessage, ServerResponse as NodeServerResponse } from 'http'
 import { withoutTrailingSlash } from 'ufo'
 import { lazyEventHandler, toEventHandler, createEvent, isEventHandler, eventHandler } from './event'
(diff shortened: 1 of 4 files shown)


eduardosada:
I really appreciate this. Keep up the good work!

Issue #170, opened 2022-09-01
Reading body after reading body in seperate module hangs request


chmking:
While developing a module to add csrf support for h3, I've discovered that reading the body after reading it in the module middleware does not pass the symbol conditional and attempts to parse the body a second time which causes the request to hang.

```js
const csrf = require("@chmking/h3-csrf");
const h3 = require("h3");
const http = require("http");

const app = h3.createApp();
app.use(csrf.csrf());

app.use("/login", async (event) => {
  await h3.readBody(event);
  return "success";
});

app.use("/", (event) => {
  const token = event.req.csrfToken();
  return `<form action="/login" method="POST"><input name="_csrf" type="hidden" value="${token}"></input><input type="submit"></form>`;
});

http.createServer(app).listen(8080);
```

Visit localhost:8080 to get the csrf cookie and then click the submit button.

Throwing a breakpoint on `await h3.readBody` and inspecting the call shows the symbol in the request object but the conditional fails and does not early out as expected.

This same behavior does not occur when read twice inside the same module. There is test demonstrating this in [@chmking/h3-csrf](https://github.com/chmking/h3-csrf/blob/v0.2.1/src/index.spec.ts#L215)

chmking:
I've tried switching `@chmking/h3-csrf` to use h3 as a peerDependency with no improvement. After walking the code in the debugger it is very clear that it is not returning the parsed body from the previous call:
```js
if (ParsedBodySymbol in event.req) {
  return event.req[ParsedBodySymbol];
}
```
![Screen Shot 2022-09-03 at 7 38 02 AM](https://user-images.githubusercontent.com/12450799/188273016-c0370e96-cb80-46a8-98b1-7b4cf141ac49.png)

Additionally, should raw and parsed body be using the same Symbol key?

```js
const RawBodySymbol = Symbol("h3RawBody");
const ParsedBodySymbol = Symbol("h3RawBody");
```

chmking:
I'm fairly new to JS and Symbols, would this be fixed by using the global symbol registry for the symbols?
```js
const RawBodySymbol = Symbol.for("h3RawBody");
const ParsedBodySymbol = Symbol.for("h3ParsedBody");
```
This should mean that future calls receive the same Symbol, correct?

PR #174, opened 2022-09-03
fix(readBody, readRawBody): use global symbol

chmking:
This addresses the issues of #170 where a second read from the body in a different library causes the request to hang.

Changing this to use `Symbol.for()` will first check the global registry for the symbol before generating a new one. I assume this means that different registries will share the same Symbol for the body and correctly early out once processed.


pi0 (approved):
Thanks!

PR #184, opened 2022-09-28
feat: add `writeEarlyHints` utility

danielroe:
https://github.com/unjs/nitro/issues/439


pi0:
Do you mind to update readme?

PR #201, opened 2022-10-16
feat: useBody takes content-type header into account

tobiasdiez:
Improves the body parsing methods in the following ways:
- readBody: only parse body as JSON if content-type indicates it is JSON
- improve internal and public typing information
- add tests, especially for empty string handling (refs https://github.com/unjs/h3/issues/197)

Depends on https://github.com/unjs/destr/pull/10.

### DIFF
--- a/src/utils/body.ts
+++ b/src/utils/body.ts
@@ -1,3 +1,4 @@
+import type { IncomingMessage } from 'http'
 import destr from 'destr'
 import type { Encoding, HTTPMethod } from '../types'
 import type { H3Event } from '../event'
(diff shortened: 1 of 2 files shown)


tobiasdiez:
I think it would also make sense that `readBody` actually throws an error if the content type says its json but there are parse errors. Currently, this is hidden and devs don't get the chance to react to this (e.g. return an error). Any idea how to make this happen? A boolean object to destr that disables the validation checks?

pi0:
@tobiasdiez Would you please open a new PR that adds `{ validateJSON: true }` support JSON validation to `useBody`?

tobiasdiez:
I can do this, but what should be the semantics of `validateJSON`?

pi0:
Maybe as you mentioned in this PR, validate both headers and also we can throw an error on failed parsing with `readBody(event, { validateJSON: true })` ?

(We can add an option for destr `{ strict: true }` [here](https://github.com/unjs/destr/blob/2f6612533dd798d2ff5786274f2244558f86d09c/src/index.ts#L38) that throws error instead of silent fallback)

tobiasdiez:
Okay, first step is at https://github.com/unjs/destr/pull/11.

PR #202, opened 2022-10-17
feat(writeEarlyHints): handle object-format early hints

danielroe:
subsequent to previous implementation, https://github.com/nodejs/node/pull/44820 made format object-based. this pr aligns h3 with node's current implementation

before https://github.com/nodejs/node/pull/44874 is released, we also have to manually replace crossorigin attribute 🤦

### DIFF
--- a/src/utils/response.ts
+++ b/src/utils/response.ts
@@ -86,35 +86,31 @@ export function sendStream (event: H3Event, data: any): Promise<void> {
   })
 }
 
-export function writeEarlyHints (event: H3Event, hints: Record<string, string | string[]>, callback?: () => void) {
+const noop = () => {}
+export function writeEarlyHints (event: H3Event, hints: string | string[] | Record<string, string | string[]>, cb: () => void = noop) {
   if (!event.res.socket && !('writeEarlyHints' in event.res)) {
-    if (callback) {
-      callback()
-    }
+    cb()
     return
   }
 
-  // Normalize if string is provided
-  if (typeof hints === 'string') {
+  // Normalize if string or string[] is provided
+  if (typeof hints === 'string' || Array.isArray(hints)) {
     hints = { link: hints }
   }
 
   if ('writeEarlyHints' in event.res) {
-    return event.res.writeEarlyHints(hints, callback)
+    return event.res.writeEarlyHints(hints, cb)
   }
 
   const headers: [string, string | string[]][] = Object.entries(hints)
   if (!headers.length) {
-    if (callback) {
-      callback()
-    }
+    cb()
     return
   }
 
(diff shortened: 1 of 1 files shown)


src/utils/response.ts
```diff
@@ -86,19 +86,39 @@ export function sendStream (event: H3Event, data: any): Promise<void> {
...
+    cb()
+    return
+  }
+
+  let hint = 'HTTP/1.1 103 Early Hints'
+  if (hints.link) {
+    const links = Array.isArray(hints.link) ? hints.link : Array.from(hints.link)
+    hint += `\r\nLink: ${links.join('\r\n')
+      // TODO: remove when https://github.com/nodejs/node/pull/44874 is released
+      .replace(/; crossorigin/g, '').split(', ')}`
```
pi0, line 116: Shouldn't we join it back before stringification?

danielroe: no. only string is supported for these headers

src/utils/response.ts
```diff
@@ -86,35 +86,31 @@ export function sendStream (event: H3Event, data: any): Promise<void> {
...
-    }
+    cb()
     return
   }
 
   let hint = 'HTTP/1.1 103 Early Hints'
-  const [, link] = headers.find(([header]) => header.toLowerCase() === 'link') || []
-  if (link) {
-    const links = Array.isArray(link) ? link : [link]
+  if (hints.link) {
```
danielroe, line 112: I did it previously as users can pass `{ Link }` presumably?

src/utils/response.ts
danielroe: oh, yes. we shouldn't split it here

pi0: split() retuns an array... Array[].toString format joins string by `,` itself but why we needed split first?

src/utils/response.ts
pi0: Check last commet fixed by lowercasing

src/utils/response.ts
```diff
@@ -86,35 +86,31 @@ export function sendStream (event: H3Event, data: any): Promise<void> {
...
+    const links = Array.isArray(hints.link) ? hints.link : Array.from(hints.link)
```
danielroe, line 113:
```bash
> Array.from('test')
[ 't', 'e', 's', 't' ]
```

pi0: ahh my bad! Added fix

src/utils/response.ts
danielroe: we shouldn' tsplit it here at all - it needs to be earlier. see last commit

pi0 (approved):
tried locally on nuxt3

PR #207, opened 2022-10-19
feat(readBody): validate requests with `application/json` content type

tobiasdiez:
Follow-up from https://github.com/unjs/h3/pull/201. Changes are:
- Throw error if content-type says the request body is json but the body is not valid json; uses the newly introduced strict mode from https://github.com/unjs/destr/pull/11.
- Doesn't try to parse the body (as json) if content-type says its plain text
- Improve internal and external typing information. Fixes https://github.com/unjs/h3/issues/223.
- Streamline codeflow in `readBodyRaw`


tobiasdiez:
@pi0 this is now ready for review

nopeless:
@tobiasdiez Nice pr; I ended up fixing the encoding signature too just to realize that you had a fix haha

tobiasdiez:
@pi0 merged in the latest version from main, so this PR is ready for review again

tobiasdiez:
@pi0 would appreciate if you could have a look at this PR. Thanks!

Hebilicious:
@pi0 I think this is overall a good change, would like to know what you think before approving.

pi0:
Hi, dear @tobiasdiez, and sorry for checking this late.

Before merging, I made a few changes to fix the behavior matching PR description (we strictly validate json body only when `application/json` is passing otherwise using default `defu` behavior with a try-parse strategy which also ensures this is a non-breaking change). 

I will be merging the type fix issue for `readRawBody` in a separate commit into the main (on your behalf). In the future please consider splitting fixes to allow iterating faster and having a clear changelog 🙏🏼 

Cheers!

PR #219, opened 2022-10-31
fix: import type from correct location

### LINKED ISSUE #216: Generated d.ts causes Type Error
I am using h3 from TypeScript.
But I get a Type Error.

The generated d.ts references `src/types`.
But src/types doesn't exist.
image

image

nmathew98:
resolves #216


danielroe (approved):

pi0 (approved):
Thanks

PR #226, opened 2022-11-03
feat: add `proxyRequest` util

### LINKED ISSUE #224: Proxy request `method`, `headers` and `body` with  `sendProxy()`
I would expect `sendProxy` to copy `method`, `headers` and `body` from the original request.

```ts
// Copy request method
const method = event.req.method

// Copy request headers
const ignoredHeaders = [
	// https://github.com/nodejs/undici/blob/cd78eb57203ce85a26312290c1bef4017e30cf74/lib/core/request.js#L196-L220
	'transfer-encoding',
	'connection',
	'keep-alive',
	'upgrade',
	'expect',
]
const headers: HeadersInit = {}
for (const [header, value] of Object.entries(getRequestHeaders(event))) {
	if (value === undefined || ignoredHeaders.includes(header)) continue
	headers[header] = value
}
// TODO: Set x-forwarder-* headers?

// Copy request body
const payloadMethods = ['PATCH', 'POST', 'PUT', 'DELETE']
const body = payloadMethods.includes(method)
	? await (readRawBody(event, false) as unknown as Promise<Buffer>) // https://github.com/unjs/h3/issues/223
	: undefined

sendProxy(event, target.toString(), {
	fetchOptions: { method, headers, body },
})
```

pi0:
Resolves #224 (Thanks @P4sca1 for suggestion)

This PR adds a new `proxyRequest` utility built on top of `sendProxy` that proxies incoming request headers, body and method to target. (path is not rewritten)

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,7 @@ H3 has concept of compasable utilities that accept `event` (from `eventHandler((
 - `assertMethod(event, expected, allowHead?)`
 - `createError({ statusCode, statusMessage, data? })`
 - `sendProxy(event, { target, headers?, fetchOptions?, fetch?, sendStream? })`
+- `proxyRequest(event, { target, headers?, fetchOptions?, fetch?, sendStream? })`
 
 👉 You can learn more about usage in [JSDocs Documentation](https://www.jsdocs.io/package/h3#package-functions).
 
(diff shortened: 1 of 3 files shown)


scripness:
@pi0 is it supposed not to rewrite the path? I still get the path rewritten - proxying in a nuxt server middleware to an internal route, just trying to hide some path in the url.

PR #231, opened 2022-11-15
refactor: move `event.req` and `event.res` to `event.node.*`

pi0:
Ref #73

Using soft deprecation to use `event.node.req` and `event.node.res` instead of `event.req` and `event.res` preparing users for next iterations.

A new `event.path` getter also added to make accessing to path easier. (used `path` instead of `url` to avoid confusing hostname inclusion)


n4an:
>  preparing users for next iterations.
poor users, why are you so with them?

pi0:
Please see https://github.com/unjs/h3/issues/73 for motives and reasons for this refactor :)

PR #240, opened 2022-11-18
docs: add example use nested router usage

nopeless:
It was hard to figure out


nopeless:
@pi0 By the way, I realized later on that it gives a deprecation warning. I simply could not find a fix though. Would you mind amending my PR?

nopeless:
#253 is the tracker

Atinux:
Could you please check with latest release of `h3` if: `app.use('/api', router)` works?

nopeless:
> Could you please check with latest release of `h3` if: `app.use('/api', router)` works?

sure

nopeless:
@Atinux It appears to not work. In fact, I cannot get nested router to work at all on the main branch

nopeless:
@pi0 This needs investigation

Atinux:
Do you mind sharing a Stackblitz or GitHub repo @nopeless ?

nopeless:
> Do you mind sharing a Stackblitz or GitHub repo @nopeless ?

Well, would you mind doing that? There are currently no tests that prove that this works so it would be better to prove that it works

nopeless:
> Could we use middleware to nested route, if it works again?

There is no big distinction between middleware and nested router in h3. I am assuming that a middleware pattern can temporary stub a different baseurl in order for the radix3 router to work properly but that is just a very bad idea in general. There are so many undocumented things going on that I am not sure what the actual root cause is. Best thing to do is to wait until maintainers resolve this

pi0:
Here is an example using nested routers: https://stackblitz.com/edit/github-2bmusk?file=app.ts&startScript=dev

PR #250, opened 2022-11-26
feat: add`sendNoContent`  utility to create a 204 response

tobiasdiez:
Adds a little helper method to immediately end the response without sending any data. 

Note sure about the naming, but otherwise ready to be review.


src/utils/response.ts
```diff
@@ -17,6 +17,22 @@ export function send (event: H3Event, data?: any, type?: string): Promise<void>
...
 }
 
+/**
+* Respond with an empty payload.<br>
+* Note that calling this function will close the connection and no other data can be sent to the client afterwards.
+*
+* @param event H3 event
+* @param code status code to be send. By default, it is `204 No Content`.
+*/
+export function sendEmpty (event: H3Event, code: number = 204) {
```
stafyniaksacha, line 27: Is `sendNoContent` a more standard name?

pi0 (approved):
Thanks!

PR #253, opened 2022-11-28
[WIP] Code improvement and TS fixes

nopeless:


nopeless:
@pi0 I pushed a temporary commit so you can see the work here. If you don't like something or think that I'm going in the wrong direction, leave a review

Also, after you are done submitting all of your reviews and I'm finished with the fixes, I'm going to reset the head 1 back and do a rebase so commits are semantic

nopeless:
Also quick question, should I define some getters/setters for `res`, `req` so that users know how to set up an express middleware that uses non-public api?

ex)

```ts
res._implicitHeader
// ^ Throws (Wrap it in an express instance. Guide: <link>)
```

nopeless:
Another question: does this library support error handlers (4 parameter function) of express? I'm currently not seeing a code that does that

nopeless:
I think the `use` function is unnecessarily complicated in both implementation and type that it's hard to test and make it reliable. We should do a refactoring

nopeless:
The lazy api is fundamentally fragile. I'm ignoring it for now

nopeless:
The above commit also includes a feature update which allows users to directly pass in a router instead of having to do `eventHandler(router.handler)`

pi0:
Hi dear @nopeless. Thanks for all your efforts on this PR 💚  There are several possible improvements I can see from the changes. But they should be splitted into smaller branches with the corresponding issue and PRs each. 

Do you mind to help on start by creating issues for lazy handler and router implementations?

As for express compact, I think we can mainly improve docs, or possibly introduce a new utility like `defineExpressHandler` that tries to add polyfills. Would be happy to discuss on this too once having an issue.

nopeless:
> Hi dear @nopeless. Thanks for all your efforts on this PR 💚  There are several possible improvements I can see from the changes. But they should be splitted into smaller branches with the corresponding issue and PRs each. 
> 
> Do you mind to help on start by creating issues for lazy handler and router implementations?
> 
> As for express compact, I think we can mainly improve docs, or possibly introduce a new utility like `defineExpressHandler` that tries to add polyfills. Would be happy to discuss on this too once having an issue.
> 
> 

Ok, give a rough priority list

PR #256, opened 2022-11-29
refactor(createRouter): deprecate misspelled `preemptive` option

### LINKED ISSUE #255: CreateRouterOptions misspelling of "preemptive"
https://github.com/unjs/h3/blob/4d3b596837313fe16f156458694843dced39d14b/src/router.ts#L23

`CreateRouterOptions` has a `preemtive` property that is currently misspelled. It should be `preemptive` with a `p` instead.

NozomuIkuta:
Resolves #255.

cc: @AaronSmithX

### DIFF
--- a/src/router.ts
+++ b/src/router.ts
@@ -20,7 +20,9 @@ interface RouteNode {
 }
 
 export interface CreateRouterOptions {
+  /** @deprecated Please use `preemptive` instead. **/
   preemtive?: boolean
+  preemptive?: boolean
 }
 
 export function createRouter (opts: CreateRouterOptions = {}): Router {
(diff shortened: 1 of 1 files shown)


src/router.ts
```diff
@@ -61,7 +63,7 @@ export function createRouter (opts: CreateRouterOptions = {}): Router {
     // Match route
     const matched = _router.lookup(path);
     if (!matched || !matched.handlers) {
-      if (opts.preemtive) {
+      if (opts.preemptive || opts.preemtive) {
```
nopeless, line 66: Do not taint the logic code with hacks like this

A better solution would be preprocessing the options
`opts.preemptive ||= opts.premtive`
right when the function starts

nopeless (changes requested):

src/router.ts
pi0: This is good. We also avoid using recent syntaxes (logical or supported for Node 15+).

pi0 (approved):
Thanks!

PR #278, opened 2022-12-13
fix(readBody): handle top-level arrays in url-encoded data

danielroe:
### 🔗 Linked issue

resolves https://github.com/unjs/h3/issues/275

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation or readme)

### 📚 Description

Previously a valid url-encoded string with multiple entries for the same value was not properly parsed. This creates arrays in such cases; it still doesn't support advanced nested arrays like `qs` but handles this specification as expected.

### 📝 Checklist

### DIFF
--- a/src/utils/body.ts
+++ b/src/utils/body.ts
@@ -67,14 +67,14 @@ export async function readBody<T=any> (event: H3Event): Promise<T> {
     const form = new URLSearchParams(body);
     const parsedForm: Record<string, any> = Object.create(null);
     for (const [key, value] of form.entries()) {
-      if (!(key in parsedForm)) {
+      if ((key in parsedForm)) {
+        if (!Array.isArray(parsedForm[key])) {
+          parsedForm[key] = [parsedForm[key]];
+        }
+        parsedForm[key].push(value);
+      } else {
         parsedForm[key] = value;
-        continue;
       }
-      if (!Array.isArray(parsedForm[key])) {
-        parsedForm[key] = [parsedForm[key]];
-      }
-      parsedForm[key].push(value);
     }
     return parsedForm as unknown as T;
   }
--- a/src/utils/body.ts
+++ b/src/utils/body.ts
@@ -64,7 +64,16 @@ export async function readBody<T=any> (event: H3Event): Promise<T> {
   const body = await readRawBody(event) as string;
 
   if (event.node.req.headers["content-type"] === "application/x-www-form-urlencoded") {
-    const parsedForm = Object.fromEntries(new URLSearchParams(body));
+    const form = new URLSearchParams(body);
+    const parsedForm: Record<string, any> = {};
+    for (const [key, value] of form.entries()) {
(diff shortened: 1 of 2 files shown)


src/utils/body.ts
```diff
@@ -64,7 +64,16 @@ export async function readBody<T=any> (event: H3Event): Promise<T> {
   const body = await readRawBody(event) as string;
 
   if (event.node.req.headers["content-type"] === "application/x-www-form-urlencoded") {
-    const parsedForm = Object.fromEntries(new URLSearchParams(body));
+    const form = new URLSearchParams(body);
+    const parsedForm: Record<string, any> = {};
```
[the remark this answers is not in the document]
danielroe: Out of curiosity, why?

src/utils/body.ts
```diff
@@ -64,7 +64,16 @@ export async function readBody<T=any> (event: H3Event): Promise<T> {
...
+    for (const [key, value] of form.entries()) {
+      if (key in parsedForm && !Array.isArray(parsedForm[key])) {
+        parsedForm[key] = [parsedForm[key]];
+        parsedForm[key].push(value);
```
pi0, line 72: I think we need to handle when key is there and value is already an array (to test, adding another prop with same name)

src/utils/body.ts
```diff
@@ -64,7 +64,16 @@ export async function readBody<T=any> (event: H3Event): Promise<T> {
...
+      } else {
+        parsedForm[key] = value;
```
pi0, line 74: Style: This behavior can be short circuit before array check.

src/utils/body.ts
danielroe: I don't think we support arrays at all, at the moment. So I think this is safe.

danielroe:
```js
[...new URLSearchParams('a[]=j').entries()]
// [ "a[]", "j" ]
[...new URLSearchParams('a=[j]').entries()] 
// // [ "a", "[j]" ]
```

src/utils/body.ts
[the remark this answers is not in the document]
pi0: Possible prototype pollution. This makes constructor null (we might have missed some other places)

src/utils/body.ts
danielroe: if you wish we can use `continue` but I'm not sure it's clearer

pi0: indeed without seems better

pi0 (approved):

alphaelf:
on [this line](https://github.com/unjs/h3/blob/03243bab1c5db92ec615fff073101114e104103c/src/utils/body.ts#L98) and [this line](https://github.com/unjs/h3/blob/03243bab1c5db92ec615fff073101114e104103c/src/utils/body.ts#L100) can we use `destr` to store typed value instedof string value?, for now the `undefined, true, false, null, ...` values will returned as string

PR #295, opened 2023-01-08
feat: add `eventHandlers` to chain handlers

### LINKED ISSUE #192: Allow to chain event handlers
It would be nice if event handlers could be chained to create a new event handler that invokes each handler after each other. Something like:

```
chain(defineCorsHandler(...), eventHandler(event => ...)) // this is an event handler again
```

This is especially relevant for usage in nuxt, where one usually declares one event handler for a certain route and then in the declaration would like to specify cors options (using h3-cors) in parallel to invoking the actual handler.

Related: https://github.com/unjs/h3/issues/177

notshekhar:
Resolves #192


Atinux:
Could you better explain your use case?

NozomuIkuta:
@Atinux 

This PR seems to be related to #192.

Atinux:
Thank you for pointing that out @NozomuIkuta 

I believe it also need to add `defineEventHandlers` but support mot things, somehow it's not only because it returns nothing that we should keep calling the next handler.

It should check is the response has been ended somehow.

notshekhar:
> Thank you for pointing that out @NozomuIkuta
> 
> I believe it also need to add `defineEventHandlers` but support more things, somehow it's not only because it returns nothing that we should keep calling the next handler.
> 
> It should check is the response has been ended somehow.

Detecting whether a function returns something or not can be challenging in JavaScript, mainly because JavaScript is a dynamic language and a function's behaviour can change during runtime.

notshekhar:
@pi0

Hebilicious:
@notshekhar Thanks for the PR. People who wants this behaviour can use it as a reference to add this helpers in their projects, as it doesn't require any extra changes. I'm gonna tag this as pending until we come up with an implementation that fits @Atinux requirements.

Hebilicious (commented):
@notshekhar Thanks for the PR. People who wants this behaviour can use it as a reference to add this helpers in their projects, as it doesn't require any extra changes. I'm gonna tag this as pending until we come up with an implementation that fits @Atinux requirements.

pi0:
Hi dear @notshekhar and sorry your PR didn't get responses in a long time. 

I still belive that while you can always prefer to use a pattern like this, offering it as a core utility is not a good practice to chance actual event handlers. We are thinking about an alternative object syntax (see #424 to get some ideas) which allows `before`/`after` hooks instead.

PR #315, opened 2023-02-06
feat: session support

### LINKED ISSUE #58: How implement Session
Is there a way to implement session in h3?
```js
req.session => object
```

pi0:
Resolves #58

This PR adds basic session support. Sessions are sealed (encrypted and signed with a private key) in cookies using [iron-webcrypto](https://github.com/brc-dd/iron-webcrypto) (thanks @brc-dd ❤️) and [uncrypto](https://github.com/unjs/uncrypto) to work out of with Node.js and other environments.

The encrypted session is carried over cookies, therefore no server storage is required for session support.

Multiple session support is also possible with this implementation and custom `name` option (default is `h3` and used for cookie name as well )

Default cookie options are secure, httpOnly with path of `/` (can be configured)

Exposed utils:
- `useSession`: Initializes session and returns an `{ data, id, update(val), clear() }` interface to manage session. Wrapping `getSession`, `upsteSession` and `clearSession`
- `getSession`: Treis to reuse or otherwise init an empty session for user
- `updateSession`: Updates session both in cookie and context. Also initializes if not initialized yet
- `clearSession`: Clears session from both cookie and context

Simple usage:

```js
import { useSession } from 'h3'

const sessionPassword = "secretsecretsecretsecretsecretsecretsecret"

export default eventHandler(async (event) => {
      const session = await useSession<{ ctr: number }>(event, { password });
      await session.update((data) => ({ ctr: Number(data.ctr || 0) + 2 }));
      await session.update({ ctr: Number(session.data.ctr || 0) - 1 });
      return `Hello! You visited this page ${session.data.ctr} times. session id: ${session.id})`;
})
```

Known issues:
 - each updates appends a new cookie. fixing in second PR

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,9 +1,8 @@
 {
-  "extends": [
-    "eslint-config-unjs"
-  ],
+  "extends": ["eslint-config-unjs"],
   "rules": {
     "unicorn/no-null": "off",
-    "unicorn/number-literal-case": "off"
+    "unicorn/number-literal-case": "off",
+    "@typescript-eslint/no-non-null-assertion": "off"
   }
 }
--- a/README.md
+++ b/README.md
@@ -153,6 +153,10 @@ H3 has a concept of composable utilities that accept `event` (from `eventHandler
 - `getResponseStatus(event)`
 - `getResponseStatusText(event)`
 - `readMultipartFormData(event)`
+- `useSession(event, { password, name?, cookie?, seal?, crypto? })`
+- `getSession(event, { password, name?, cookie?, seal?, crypto? })`
+- `updateSession(event, { password, name?, cookie?, seal?, crypto? }), update)`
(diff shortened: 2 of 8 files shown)


BananaAcid:
Seems to me, there is an export missing for one of the types [`SessionUpdate`](https://github.com/unjs/h3/blob/main/src/utils/session.ts#L122) - which makes referencing the type in a wrapper fn harder. Is that actually on purpose?

peerreynders:
> Is that actually on purpose?

My impression is that the TS community likes to keep their types implicit in order to minimize the potential maintenance overhead of explicit (and published) types. If you want to explicitly capture/name the type:

```ts
import { updateSession } from "h3";

type Holder = {
  counter: number;
};

type HolderUpdate = (Parameters<typeof updateSession<Holder>>)[2];
```
TypeScript Handbook: [`Parameters<type>`](https://www.typescriptlang.org/docs/handbook/utility-types.html#parameterstype)

PR #319, opened 2023-02-07
separate multiple cookie headers

meixger:
re: https://github.com/unjs/h3/issues/318 proxying multiple cookies and split by comma


pi0 (approved):
Thanks ❤️

oleghalin:
I have problem with this PR, Second cookie from header not applied and found next in the web

Origin servers SHOULD NOT fold multiple Set-Cookie header fields into a single header field. The usual mechanism for folding HTTP headers fields (i.e., as defined in [RFC2616]) might change the semantics of the Set-Cookie header field because the %x2C (",") character is used by Set-Cookie in a way that conflicts with such folding.

What you think @pi0 ?

PR #328, opened 2023-02-14
fix(clearSession): accept partial session config

zhiyuanzmj:
``` ts
// before need `password`
await clearSession(event, {password: '...'})

//  after
await clearSession(event)
```

### DIFF
--- a/src/utils/session.ts
+++ b/src/utils/session.ts
@@ -202,7 +202,7 @@ export async function unsealSession(
   return unsealed;
 }
 
-export async function clearSession(event: H3Event, config: SessionConfig) {
+export async function clearSession(event: H3Event, config: Partial<SessionConfig> = {}) {
   const sessionName = config.name || DEFAULT_NAME;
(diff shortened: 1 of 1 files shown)


src/utils/session.ts
```diff
@@ -202,7 +202,7 @@ export async function unsealSession(
   return unsealed;
 }
 
-export async function clearSession(event: H3Event, config: SessionConfig) {
+export async function clearSession(event: H3Event, config: Partial<SessionConfig> = {}) {
```
pi0, line 205: Making _name_ also optional, can be both good DX but also lead to programmer mistakes forgetting to provide custom name. (also might be needed to clear cookie from subpath)

pi0: Maybe only add Partial to remove password requirement?

zhiyuanzmj (commented):
feat(clearSession): password optional

pi0 (approved):
Thanks!

PR #329, opened 2023-02-14
chore: improve `lint` npm script

NozomuIkuta:
I applied the latest unjs template, where `lint:fix` npm script is available with which we can fix/format files.
I also enable ESLint cache to speed it up.


pi0:
thanks!

PR #357, opened 2023-03-15
fix: sanitize `statusMessage` of disallowed chars

danielroe:
### ❓ Type of change
- [ ] 📖 Documentation (updates to the documentation or readme)

### 📚 Description

It will break the response if we set any of these characters by accident (e.g. if a user includes in an error message).

Related: https://github.com/nuxt/nuxt/issues/14688

### 📝 Checklist

### DIFF
--- a/src/utils/response.ts
+++ b/src/utils/response.ts
@@ -42,7 +42,12 @@ export function setResponseStatus(
 ): void {
   event.node.res.statusCode = code;
   if (text) {
-    event.node.res.statusMessage = text;
+    // Allowed characters: horizontal tabs, spaces or visible ascii characters: https://www.rfc-editor.org/rfc/rfc7230#section-3.1.2
+    event.node.res.statusMessage = text.replace(
+      // eslint-disable-next-line no-control-regex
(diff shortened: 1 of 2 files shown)


src/utils/response.ts
```diff
@@ -40,9 +40,16 @@ export function setResponseStatus(
...
-  event.node.res.statusCode = code;
+  if (code) {
+    event.node.res.statusCode = code;
+  }
   if (text) {
-    event.node.res.statusMessage = text;
+    // Allowed characters: horizontal tabs, spaces or visible ascii characters: https://www.rfc-editor.org/rfc/rfc7230#section-3.1.2
+    event.node.res.statusMessage = text.replace(
+      // eslint-disable-next-line no-control-regex
+      /[^\u0009\u0020-\u007E]/g,
```
pi0, line 50: Just to mention for future we might extend the scaped for XSS as well (we had report of similar issues with ipx fixed by JSON.stringify and removing more chars)

pi0 (approved):
Thanks ❤️

PR #358, opened 2023-03-15
fix: sanitize utils and sanitize all response code and messages

pi0:
Following up #357, this PR fixes all setters of `statusCode` and `statusMessage` with two new `sanetizeStatusCode` and `sanetizeStatusMessage` utils


src/error.ts
```diff
@@ -20,12 +24,12 @@ export class H3Error extends Error {
       H3Error,
       "message" | "statusCode" | "statusMessage" | "data"
     > = {
-      message: this.message,
-      statusCode: this.statusCode,
+      message: sanetizeStatusMessage(this.message),
```
danielroe, line 27: Should be sanitize.

src/error.ts
```diff
@@ -83,14 +87,14 @@ export function createError(
...
   if (input.statusCode) {
-    err.statusCode = input.statusCode;
+    err.statusCode = sanetizeStatusCode(input.statusCode, err.statusCode);
   } else if (input.status) {
-    err.statusCode = input.status;
+    err.statusCode = sanetizeStatusCode(input.status, err.statusCode);
   }
   if (input.statusMessage) {
-    err.statusMessage = input.statusMessage;
+    err.statusMessage = sanetizeStatusMessage(input.statusMessage);
```
danielroe, line 95: I didn't think we needed to sanitize the error object itself, only when it was set to the response object.

src/utils/sanetize.ts
```diff
@@ -0,0 +1,23 @@
+// Allowed characters: horizontal tabs, spaces or visible ascii characters: https://www.rfc-editor.org/rfc/rfc7230#section-3.1.2
+// eslint-disable-next-line no-control-regex
+const UNALLOWEED_STATUS_CHARS = /[^\u0009\u0020-\u007E]/g;
```
danielroe, line 3:
```suggestion
const DISALLOWED_STATUS_CHARS = /[^\u0009\u0020-\u007E]/g;
```

src/error.ts
pi0: Inddeed. Doing this as extra caution in case someone is picking statusMessage for custom setter.

danielroe: I think we should leave it up to them. Maybe their error message is in Arabic (see example here: https://github.com/nuxt/nuxt/issues/14688). They might want to be able to read that error message elsewhere, or display it (as, for example in Nuxt we would do in the page).

pi0: I think this is the purpose of `message` right? `status{code,message}` are purposed for HTTP headers.

danielroe: I think it still makes sense to leave it up to them. If they are using a custom setter, then they might want to sanitise differently - e.g. replace with `-`.

pi0: If the input is already santized (and includes valid chars) we wouldn't remove them. 

In general `statusMessage` should probably be used less IMO. HTTP clients (browser, node, curl, etc) already have defaults matching the code.

For anything else (display purpose), `message` is there which we also need to sanetize from xss later.

danielroe: I get that probably people _should_ do that. But this is very much an edge case, as we will always sanitise it _ourselves_ in h3, nitro or nuxt before setting the response status.

The only remaining situation is where the _user_ sets the header based on `error.statusMessage`. And for that purpose we export utilities they can use to sanitise the header. Sanitising it at the creation of the error itself is a breaking change that may affect people more than we expect, including breaking error pages in Nuxt if people were using `statusMessage` rather than `message`.

pi0: Looking at linked nuxt issue, throw error with unsupported characters was always broken (please correct me if i am wrong ) 

message is standard error property also it is nothing new.

i mainly insist on doing this to avoid breaking changes from now on that we are normalizing errorMessage field, it should be consistent and universally :) Also that thinking to support a standard response object to h3 (similar to ipx generic response object) for that we also have similar concern and if not fixing a standard normalization now, it would be lots of inconsistency on a field that only in one case is not normalized.

danielroe: I'm not only thinking about throwing the error on server side, but also on client. Or throwing an error only to have it bubble up and be handled elsewhere in the app.

pi0: Hmm i see it is only for showerror. Well we need to fix it anyway. Status message was never meant for that evne if wrongly used.

I think we can make it a warning only for nuxt wdyt? (When normalized and input status message are not the same warn to use message field instead )

danielroe: It just feels like extra work to me - we are normalising statusMessage twice. Once when we create error and once when we set statusMessage. And the error doesn't _need_ normalising separately if it is being used for the envisioned purpose. The only reason it needs to be normalised if it is _not_ being used for the normal purpose, such as if the user is going to directly set the response statusMessage from it.

pi0: The status* fields on error object specify exist because of http response so it is not wrong to normalize them.

it comes to consistency and external usage of error object. 

If it is not normalized today, breaking change surface of normalizing this field will extend in the future where we cannot easily do this anymore. Today only issue is with (wrong) usages only and we dan soft warn for a while and ensure normalization in long term. I agree with your concern of util currently being used in places out of http response and warning seems a better way to move usage to the correct message field.

pi0: Updated with warn.

src/error.ts
```diff
@@ -83,15 +88,24 @@ export function createError(
...
     err.statusMessage = input.statusMessage;
   } else if (input.statusText) {
     err.statusMessage = input.statusText;
   }
+  if (err.statusMessage) {
+    const originalMessage = err.statusMessage;
+    err.statusMessage = sanitizeStatusMessage(err.statusMessage);
+    if (err.statusMessage !== originalMessage) {
+      console.warn(
+        "[h3] `statusMessage` has been stripped from error. Please prefer using `message` for longer messages."
```
danielroe, line 105:
```suggestion
        "[h3] `statusMessage` has been sanitized. Please prefer using `message` for longer messages."
```

src/error.ts
```diff
@@ -20,12 +25,12 @@ export class H3Error extends Error {
       H3Error,
       "message" | "statusCode" | "statusMessage" | "data"
     > = {
-      message: this.message,
-      statusCode: this.statusCode,
+      message: sanitizeStatusMessage(this.message),
```
danielroe, line 28: why are we sanitizing the message as well?

danielroe:
```suggestion
      message: this.message,
```

src/error.ts
pi0: Also made soft warn for now.

src/error.ts
```diff
@@ -83,15 +88,25 @@ export function createError(
...
   }
+  if (err.statusMessage) {
+    // TODO: Always sanitize status message in the next major releases
+    const originalMessage = err.statusMessage;
+    const sanitizedMessage = sanitizeStatusMessage(err.statusMessage);
+    if (sanitizedMessage !== originalMessage) {
+      console.warn(
+        "[h3] Please prefer using `message` for longer error messages instead of `statusMessage`. In the future `statusMessage` will be sanitized by default."
+      );
+    }
```
danielroe, line 108: this looks great to me 👌

danielroe (approved):

PR #366, opened 2023-03-24
fix(readBody, readRawBody): handle raw body as buffer

### LINKED ISSUE #363: `readBody` needs to handle `Buffer`
https://github.com/unjs/h3/blob/e199df3e5d4c9709aa8ff24e20af57e63df4eb63/src/utils/body.ts#L83

johannschopplich:
- Closes #363

I'm not sure if this implementation is enough. For my use-case in [nuxt-api-party](https://nuxt-api-party.jhnn.dev/):

```ts
// https://github.com/johannschopplich/nuxt-api-party/blob/main/src/runtime/server.ts#L26
let _body = await readBody<EndpointFetchOptions>(event)

// Inconsistent `readBody` behavior in some Nitro presets
// https://github.com/unjs/nitro/issues/912
if (Buffer.isBuffer(_body))
  _body = destr((_body as Buffer).toString())
```

… this change fixed handling `Buffer` requests, e.g. when sending POST requests in a Cloudflare Nitro worker.

### DIFF
--- a/src/utils/body.ts
+++ b/src/utils/body.ts
@@ -79,8 +79,12 @@ export async function readBody<T = any>(event: H3Event): Promise<T> {
     return (event.node.req as any)[ParsedBodySymbol];
   }
 
-  // TODO: Handle buffer
-  const body = (await readRawBody(event)) as string;
+  let body = (await readRawBody(event)) as string;
+
+  // Handle inconsistent `readBody` behavior in some Nitro presets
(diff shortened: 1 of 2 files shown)


pi0:
Thanks for the PR. I have pushed couple of changes to handle it also for readRawBody consistently + tests

pi0:
Related fix https://github.com/unjs/h3/commit/19d133d6a33ffb742713f4d83f00f555aeb83224

PR #369, opened 2023-03-30
fix: parse rawBody buffer when isBuffer

dvanmali:
Flip order of operation.

Fixes issue:
```
[request error] [unhandled] [500] The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received an instance of Promise
```

### DIFF
--- a/src/utils/body.ts
+++ b/src/utils/body.ts
@@ -31,7 +31,7 @@ export function readRawBody<E extends Encoding = "utf8">(
     (event.node.req as any).body; /* unjs/unenv #8 */
   if (_rawBody) {
     const promise = Promise.resolve(
-      Buffer.isBuffer(_rawBody) ? _rawBody : Buffer.from(_rawBody)
+      Buffer.isBuffer(_rawBody) ? Buffer.from(_rawBody) : _rawBody
     );
     return encoding
       ? promise.then((buff) => buff.toString(encoding))


dvanmali:
@pi0 Ready for merge

chrisspiegl:
I just ran into this error message while deploying deploying Nuxt 3 on Firebase and trying to get VueFire auth sessions working. I hope the fix will be merged soon and working 👍.

dvanmali:
Link issue #371

sawa-ko:
Any progress?
I need this fix please 🙏.

chrisspiegl:
@sawa-ko You can use the workaround I mentioned in the issue https://github.com/unjs/h3/issues/371#issuecomment-1500129738

Basically downgrading to version `v1.6.2` will do the trick for now.

pi0:
Hi and sorry for the late review. I think the logic is correct. rawBody util is supposed to return a buffer or convert it to buffer. 

Is this issue only reproducable with nitro and firebase preset?

mao-shonen:
> Hi and sorry for the late review. I think the logic is correct. rawBody util is supposed to return a buffer or convert it to buffer.
> 
> Is this issue only reproducable with nitro and firebase preset?

This problem also happens on nodejs preset,
The problem seems to be that I used readBody once in the middleware and handler respectively,
The solution proposed above also worked for me,

pi0:
Sorry again for delay on this issue. I could just locally reproduce it.

The problem was type checking on unresolved promises fixed by https://github.com/unjs/h3/commit/2e472e8ea2eb0cad16b0e1fbf9da8173aed1df11

Please remove any `resolutions` or version fields and update lockfile to the latest to receive the fix.

---

Side note: Please consider opening an issue with reproduction in such cases. It could help to resolve much earlier 🙏🏼

PR #374, opened 2023-04-11
feat(proxy): stream request body with `streamRequest` option

iainsproat:
Example of problem and roughly-sketched implementation for https://github.com/unjs/h3/issues/373

This PR amends the behaviour of `proxyRequest` to stream the proxied request to the upstream target.

Note that this PR is a rough hack and doesn't represent a polished solution.

### DIFF
--- a/test/proxy.test.ts
+++ b/test/proxy.test.ts
@@ -165,22 +165,25 @@ describe("", () => {
         })
       );
 
-      const stream = new ReadableStream({
-        start(controller) {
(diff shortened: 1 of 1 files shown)


pi0:
Thanks for the nice PR! With recently introduced (experimental) `event.body` that returns a readable stream. Based on this, i think introducing this enhancement at least behind an option flag (streamRequest) would be a nice idea. I have made some refactors to do this.

PR #387, opened 2023-05-16
Set default return type of `readBody` to `unknown`

### LINKED ISSUE #386: `readBody` has a default returnType of  `Promise<any>` instead of `Promise<unknown>`
### Environment

Node v20.2.0
h3 1.6.6
typescript 5.0.4

### Reproduction

```js
import { eventHandler, readBody, sendNoContent } from "h3";

export const routeExample = eventHandler(async (event) => {
	const body = await readBody(event);
	// => const body: any
});
```

### Describe the bug

When using `readBody`, the output type is typed as `Promise<any>`. This in turn means it's very easy to accidentally lose type safety from the response body, as TypeScript will treat any use of the response as valid (where it's most likely not going to be). By returning `Promise<unknown>` instead, h3 can explicitly communicate that the return type should be handled with care, instead of treating every use of it as valid. You can find a super basic example of the type of any "trap" in this playground: https://www.typescriptlang.org/play?#code/MYewdgzgLgBAhgEwQFQO4hgXhgCjAVwFsAuGAwgIwFMAnASiwD4yiYBqGAJgG4BYAKAGhIsCiAQBPAIJgJpOLKwwARAAsqAGw0Z0NDQmV9+w6DDGSAqmADWYEKjCl8Nuw6VrN2mLv2GBAxBR0HHNpWTojQLQQEPEJK1t7MAigA. Note how the unknown body use throws a helpful error in this mistaken usage example.

### Additional context

You can workaround the problem by providing the explicit type you expect in the override:

```js
interface BodySchema {
  action: 'comment';
  message: string;
}

await readBody<BodySchema>(event);
```

or manually set `unknown` if you can't rely on the input body:

```js
await readBody<unknown>(event);
```

### Logs

_No response_

rijkvanzanten:
Fixes #386

### DIFF
--- a/src/utils/body.ts
+++ b/src/utils/body.ts
@@ -1,6 +1,6 @@
 import destr from "destr";
-import type { Encoding, HTTPMethod } from "../types";
 import type { H3Event } from "../event";
+import type { Encoding, HTTPMethod } from "../types";
 import { parse as parseMultipartData } from "./internal/multipart";
 import { assertMethod, getRequestHeader } from "./request";
 
@@ -77,7 +77,7 @@ export function readRawBody<E extends Encoding = "utf8">(
  * const body = await readBody(req)
  * ```
  */
-export async function readBody<T = any>(event: H3Event): Promise<T> {
+export async function readBody<T = unknown>(event: H3Event): Promise<T> {
   if (ParsedBodySymbol in event.node.req) {
     return (event.node.req as any)[ParsedBodySymbol];
   }


danielroe (approved):

Atinux (approved):

pi0:
The return type of `readBody` is consistent with `JSON.parse` and `destr` types with are both `any` as well. It is ultimately your responsibility to make sure body is both **type validated** and **runtimeValidated**.

This change is also breaking for types in existing projects. I would be happy to discuss about possible breaking change in next major versions of h3/nitro or alternatively have a more specific util like `readValidatedBody` with more strict types.

danielroe:
I don't believe replacing with `unknown` is a breaking change; I would consider it fixing a bug with type safety.

Users who do not care about type safety in this kind of way can always set `strict: false` in order for TS to treat the `unknown` as `any`.

rijkvanzanten:
> The return type of readBody is consistent with JSON.parse [...]

I'm pretty sure JSON.parse is any because it was typed before `unknown` even existed. There's also a long running suggestion around replacing those `any`s with `unknown`s for the same reason.

> [...] and destr types [...]

Which is your design decision! 😄 IMO that should be unknown as well

> It is ultimately your responsibility to make sure body is both type validated and runtimeValidated. However i agree we need better type safety..

I fully agree, which is why I proposed this change. Right now, by default, `h3` effectively disables type safety by setting it to `any` instead of making it my responsibility with an `unknown`.

pi0:
[destr v2 is out](https://github.com/unjs/destr/releases/tag/v2.0.0) with default `unknown` and genric. Next major version of h3 is planned soon so rest assured this change will happen soon and thanks for the idea.

pi0:
We have added todos in the code to change from `any` to `unknown` in next future versions. Also any new utils that are being introduced will prefer `unknown` by default. (on a related note, other than `JSON.parse`, i noticed even `response.json()` with web standards used `any` by default -- nothing that stops us from this change!)

#459 adds safe and validated (type + runtime) alternative utils and #417 a much better way to type events input via the Event handler itself.

Let's keep tracking it by #386 (closing this PR to clear up open PRs) and thanks again for pushing forward this important change in typings.

PR #390, opened 2023-05-19
feat(useSession): support custom session id generator

maou-shonen:
Sometimes it may be necessary to customize the session ID.
for example, using something like `crypto.randomBytes(20).toString('base64url')`.
having a method available in the session configuration would be convenient.

### DIFF
--- a/src/utils/session.ts
+++ b/src/utils/session.ts
@@ -28,7 +28,7 @@ export interface SessionConfig {
   seal?: SealOptions;
   crypto?: Crypto;
   /** Default is Crypto.randomUUID */
-  idGenerator?: () => string;
+  generateId?: () => string;
 }
 
 const DEFAULT_NAME = "h3";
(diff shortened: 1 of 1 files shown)


maou-shonen:
Do I need to create a session.test.ts?
Should I open another PR?

Atinux:
What would be the use case to customize the ID generation?

pi0 (approved):
Thanks!

pi0:
>  What would be the use case to customize the ID generation?

I have no usecase but makes sense to make id generation customizable (shorter, compatible with backend, more secure, crypto incompatibility, etc could be some reasons)

maou-shonen:
> What would be the use case to customize the ID generation?

Obviously, I'm replying late,😅
One of my real-world use cases is that I can set the session ID to be the same as the ID from external login (usually OAuth2),
This way, I don't need to use two DB fields or `session.data.id` to associate them.

PR #400, opened 2023-06-07
fix(types): return type for get router parameter utils

zeferinix:
Improve the return type of `getRouterParam` and `getRouterParams`. The latter's return type in particular is misleading especially when destructured:

image

And the implementation is simply returning the `event.context.params` with a fallback.

https://github.com/unjs/h3/blob/3e5f42748ab73102c1b9c2059fe8fa110e734363/src/utils/request.ts#L10-L13

Unless `event.context.params`, for some reason, can possibly have `sessions` and `params` in it?

### DIFF
--- a/src/utils/request.ts
+++ b/src/utils/request.ts
@@ -7,7 +7,9 @@ export function getQuery(event: H3Event) {
   return _getQuery(event.node.req.url || "");
 }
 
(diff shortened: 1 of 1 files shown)


pi0 (approved):
Thanks 👍🏼

PR #403, opened 2023-06-09
fix(readRawBody): handle body as object

### LINKED ISSUE #381: [nuxt] [request error] [unhandled] [500] The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received an instance of Promise
### Environment

node: v16.17
nuxt: 3.4.2
nitro: 2.3.3

### Reproduction

Nuxt API requests using useAsyncData + $fetch in frontend and server api uses `await readBody`

### Describe the bug

frontend request
```js
const body = { params: { ...route.query } }

const { data, error } = await useAsyncData(
    'server-api',
    () => $fetch('/api/server-api', { method: 'post', body: body }),
    {
        initialCache: false
    }
).catch((e) => console.error(e))
```

server API composable
```js

    const body = await readBody(event)
}
```

### Additional context

Same issue as described here https://github.com/unjs/h3/commit/19d133d by many others. 
This latest release `1.4.0` seems to break things.

There is a workaround but it doesn't feel like something we'd actually want to be doing all over projects.
https://github.com/unjs/h3/commit/19d133d#r109674536

### Logs

```sh
[nuxt] [request error] [unhandled] [500] The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received an instance of Promise
```

igorjacauna:
- Closes #381 
- Related to 19d133d

Based in on [this](https://github.com/unjs/h3/commit/19d133d6a33ffb742713f4d83f00f555aeb83224#r109674536) where seems `unenv` already parsed the `body` to `Object` and when try to create a `Buffer` from it, in `nitro` Firebase preset, throw a error mentioned [here](https://github.com/unjs/h3/issues/381#issuecomment-1568943830)

```
TypeError [ERR_INVALID_ARG_TYPE]: The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received an instance of Object
```

This change check if `body` is a `Object` and converts to string before create a Buffer from it

### DIFF
--- a/test/body.test.ts
+++ b/test/body.test.ts
@@ -194,7 +194,7 @@ describe("", () => {
       expect(result.text).toBe("200");
     });
 
-    it("handle readBody with Oject type (unenv)", async () => {
+    it("handle readBody with Object type (unenv)", async () => {
       app.use(
         "/",
         eventHandler(async (event) => {
--- a/test/body.test.ts
+++ b/test/body.test.ts
@@ -194,6 +194,26 @@ describe("", () => {
       expect(result.text).toBe("200");
     });
 
+    it("handle readBody with Oject type (unenv)", async () => {
+      app.use(
+        "/",
+        eventHandler(async (event) => {
+          // Emulate unenv
+          // @ts-ignore
+          event.node.req.body = { test: 1 };
+
+          const body = await readBody(event);
+          expect(body).toMatchObject({ test: 1 });
+
+          return "200";
+        })
+      );
+
+      const result = await request.post("/api/test").send();
+
+      expect(result.text).toBe("200");
+    });
+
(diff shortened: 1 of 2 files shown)


test/body.test.ts
```diff
@@ -194,6 +194,26 @@ describe("", () => {
       expect(result.text).toBe("200");
     });
 
+    it("handle readBody with Oject type (unenv)", async () => {
```
Atinux, line 197:
```suggestion
    it("handle readBody with Object type (unenv)", async () => {
```

pi0 (approved):
Thanks LGTM!

BTW this is something we probably have to resolve in nitro presets to make sure provided body is a stream...

PR #414, opened 2023-06-23
fix(fetchWithEvent): allow customizing fetch impl type

danielroe:
Downstream we are wrongly typing `event.$fetch`. This is part of the fix.

### DIFF
--- a/src/utils/proxy.ts
+++ b/src/utils/proxy.ts
@@ -151,7 +151,9 @@ export function getProxyRequestHeaders(event: H3Event) {
   return headers;
 }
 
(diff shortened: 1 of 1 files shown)


src/utils/proxy.ts
```diff
@@ -151,12 +151,14 @@ export function getProxyRequestHeaders(event: H3Event) {
   return headers;
 }
 
-export function fetchWithEvent(
+export function fetchWithEvent<
+  T extends (req: RequestInfo | URL, opts?: any) => any = typeof fetch
```
pi0, line 155: I think we might reserve first generic argument for fetch response type (same as $fetch, also reserving second and use this as third..)

pi0 (approved):

PR #416, opened 2023-06-24
fix(readMultipartFormData): handle utf8 encoding for `name` and `filename`

zhiyuanzmj:
After I uploaded a file with a Chinese name:
```
20xx年xx月xx日一号楼电表数据模版.xlsx
```
the filename was parsed as ISO-8859-1 encoding:
```
20xxå¹´xxæ__xxæ_¥ä¸_å_·æ¥¼ç_µè¡¨æ_°æ_®æ¨¡ç__.xlsx
```

So I used Buffer to convert it to UTF-8.
``` js
let filename = "20xxå¹´xxæxxæ¥ä¸å·æ¥¼çµè¡¨æ°æ®æ¨¡ç.xlsx";
let buffer = Buffer.from(filename, 'latin1');
filename = buffer.toString('utf8');
console.log(filename);

```


Hebilicious (approved):
This looks fine to me @pi0

pi0 (approved):
Thanks!

PR #423, opened 2023-07-04
missing helpers

Hebilicious:
- feat: add header helpers and extract into single file
- feat: add response helpers and tests
- feat: add url helpers and extract to url file
- test: run response tests on node 18+

### 🔗 Linked issue

Resolves #420

### ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme, or JSdoc annotations)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [x] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

This PR adds utility helpers to help slowly deprecating all `event.node` usage, as a part of the migration to events API #73 
### 📝 Checklist

- [x] I have linked an issue or discussion.
- [ ] I have updated the documentation accordingly.

Issue #420, opened 2023-07-02
[helpers] add missing helpers 


Hebilicious:
### Describe the feature

Some helpers that are missing from h3, and require users to explicitly rely on `event.node`. 

We should provide the following helpers as a part of the migration to  #73 

Once everything is covered, we can start deprecating `event.node` usage for end-users, as they should not rely on node agnostic API for their needs.

Header : 
- removeResponseHeaders
- removeResponseHeader
- getRequestRawHeader
- getResponseHeaders
- getResponseHeader

Response : 
- sendResponse

URL : 
- setOriginalUrlPath
- getOriginalUrlPath
- setUrlPath
- getUrlPath
- getRequestPath

Note specific :
- nodeResWritableEnded (event.node.res.writableEnded)

Reference for implementation and test https://github.com/Hebilicious/h3/pull/2/files

### Additional information

- [X] Would you be willing to help implement this feature?

pi0:
I think now we have most missing utils as `event.[]` props. Let's see what is missing in the future.

PR #427, opened 2023-07-06
feat: add `removeResponseHeader` and `clearResponseHeaders` utils

Hebilicious:
- feat: add request and response helpers
- feat: add url utilities

### 🔗 Linked issue

#420 
#423

### ❓ Type of change

### 📚 Description

This adds 7 new helpers. 

### 📝 Checklist

### DIFF
--- a/src/utils/request.ts
+++ b/src/utils/request.ts
@@ -87,6 +87,10 @@ export function getRequestHeader(
 
 export const getHeader = getRequestHeader;
 
+export function getRequestRawHeader(event: H3Event, name: string) {
+  return event.node.req.headers[name];
+}
+
 export function getRequestHost(
   event: H3Event,
   opts: { xForwardedHost?: boolean } = {}
--- a/src/utils/request.ts
+++ b/src/utils/request.ts
@@ -87,10 +87,6 @@ export function getRequestHeader(
 
 export const getHeader = getRequestHeader;
 
-export function getRequestRawHeader(event: H3Event, name: string) {
-  return event.node.req.headers[name];
-}
-
 export function getRequestHost(
   event: H3Event,
   opts: { xForwardedHost?: boolean } = {}
--- a/src/utils/request.ts
+++ b/src/utils/request.ts
@@ -132,15 +132,6 @@ export function getRequestURL(
   return new URL(path, `${protocol}://${host}`);
 }
 
-export function setOriginalUrlPath(event: H3Event, url: string) {
-  // eslint-disable-next-line @typescript-eslint/no-extra-semi
-  (event.node.req as any).originalUrlPath = url;
-}
-
-export function getOriginalUrlPath(event: H3Event) {
-  return (event.node.req as any).originalUrlPath as string;
-}
-
 export function setUrlPath(event: H3Event, url: string) {
   event.node.req.url = url;
 }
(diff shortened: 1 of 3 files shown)


src/utils/request.ts
```diff
@@ -87,6 +87,10 @@ export function getRequestHeader(
 
 export const getHeader = getRequestHeader;
 
+export function getRequestRawHeader(event: H3Event, name: string) {
+  return event.node.req.headers[name];
```
pi0, line 91: Why can't we depend on `event.headers`?

src/utils/request.ts
```diff
@@ -129,3 +133,20 @@ export function getRequestURL(
...
   return new URL(path, `${protocol}://${host}`);
 }
+
+export function setOriginalUrlPath(event: H3Event, url: string) {
+  // eslint-disable-next-line @typescript-eslint/no-extra-semi
+  (event.node.req as any).originalUrlPath = url;
+}
+
+export function getOriginalUrlPath(event: H3Event) {
+  return (event.node.req as any).originalUrlPath as string;
```
pi0, line 143: I think the original URL could be simply an internal prop for event `event._originalPath` since it is mainly for advanced/internal usecases.

src/utils/request.ts
```diff
@@ -129,3 +133,20 @@ export function getRequestURL(
...
+}
+
+export function setUrlPath(event: H3Event, url: string) {
+  event.node.req.url = url;
+}
+
+export function getUrlPath(event: H3Event) {
+  return event.node.req.url || "/";
```
pi0, line 151: Why we need this and setter?

Hebilicious: It is used in many places and end-users might be tempted to use it: https://github.com/unjs/h3/blob/091f32602c8133d355d9c9ed784e55ca53099dec/src/utils/route.ts#L12

src/utils/request.ts
Hebilicious: Will remove as I've created #435

src/utils/request.ts
pi0: Why not advice `event.path` (which is essentially `req.url` + some security fixes)

Hebilicious: The API is helper based, `getPath(event)`  would be more consistent than `event.path` for end users.
The setter is probably internal use only,  and it was needed for the fetch adapter.

pi0: We can directly set `_*` internals from within other internals such as fetch adapter if there is no need for a reusable normalization util.

Hebilicious: What should we do then ? Remove the setter and expose `getPath(event)` ?

src/utils/response.ts
```diff
@@ -148,6 +148,16 @@ export function appendResponseHeader(
 
 export const appendHeader = appendResponseHeader;
 
+export function removeResponseHeaders(event: H3Event): void {
```
pi0, line 151: I think this should support a `string[]` no? What is usecase of removing all headers?

pi0: I still don't see the reason. Why they should come from many places and need to be overridden like this? Do you have a real-world example? (If there is, we might also call it `clearResponseHeaders` for clarity)

pi0: Supporting `string[]` for `removeResponseHeader` would be a bonus 💯

Hebilicious: For a Nuxt use case, you could have middlewares/modules that adds headers, and in a specific handler, you'd want to override all of them ?

pi0 (approved):

PR #429, opened 2023-07-06
feat: add `event.method` and `event.headers`

Hebilicious:
### 🔗 Linked issue

#423, #73 

### ❓ Type of change

### 📚 Description

This PR adds `event.method` and `event.headers` support. 

### 📝 Checklist

- [ ]  I have updated the documentation accordingly.

### DIFF
--- a/src/event/event.ts
+++ b/src/event/event.ts
@@ -9,17 +9,6 @@ import {
 } from "../utils";
 import { H3Response } from "./response";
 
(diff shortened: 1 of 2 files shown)


pi0:
Thanks for the PR! I've made a couple of updates:

- Use a getter with lazy evaluation and keep the public interface read-only.
  - Similar to `_handled` we have an internal way to override them. Not to be advised.
- Fix tests for Node 16 not having global Headers. In H3 we have a working polyfill but thinking to not include it in base bundle.

PR #421, opened 2023-07-02
feat: add `readFormData` util

Hebilicious:
Resolves #419 

### ❓ Type of change

- [x] 📖 Documentation (updates to the documentation, readme or JSdoc annotations)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ ] 👌 Enhancement (improving an existing functionality like performance)
- [x] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

This PR adds support for Request and FormData in event handler.

2 new helpers are introduced : 

- `getRequestFromEvent`
- `getFormData`

This helpers relies on Request and FormData to be available in the global scope as per the Fetch specification, which means the tests are disabled for node < 18 (which is fine as 18 reaches EOL in 3 months)

```ts
// With a JSON body ...
const handler = eventHandler(async (event) => {
	const nativeRequest = await getRequestFromEvent(event);
	expect(nativeRequest instanceof Request).toBe(true);
	expect(nativeRequest.method).toBe("POST");
	expect(nativeRequest.headers.get("hello")).toBe("world");
	const body = await nativeRequest.json();
	expect(body).toMatchObject({
	  user: "john",
	});
	return "ok";
})
```  
        
```ts
// With x-www-form-urlencoded data
const handler = eventHandler(async (event) => {
	const formData = await getFormData(event);
	const user = formData.get("user");
	expect(formData instanceof FormData).toBe(true);
	expect(user).toBe("john");
	return { user };
})
``` 

### 📝 Checklist

- [x] I have linked an issue or discussion.
- [x] I have updated the documentation accordingly.

PR #431, opened 2023-07-08
feat: add runtime typesafe body and query parameters helpers

### LINKED ISSUE #402: Type and Runtime validated request utils
Currently, we have utils such as `readBody(event)` and `getParams(event)` which both have any or unspecific types and have no runtime validation.

The idea is to expose request utils that are both **runtime** and **type** safe.

We might expose some core request validation utils (that require a custom validator function and have default strict `unknown` type) + `h3/zod` integration that provides them.

Current solutions:
- [h3-zod](https://github.com/wobsoriano/h3-zod) by @wobsoriano ❤️ 
- [h3-typebox](https://github.com/kevinmarrec/h3-typebox) by @kevinmarrec ❤️

Related discussions:
- #129
- #386 / #387

### LINKED ISSUE #419: [helpers] Native Request and FormData helpers
### Describe the feature

I've implemented the following in https://github.com/Hebilicious/form-actions-nuxt

```ts
export async function getRequestFromEvent(event: H3Event) {
  const url = new URL(getRequestURL(event))
  const method = getMethod(event)
  const body = method === "POST" ? await readRawBody(event) : undefined
  return new Request(url, { headers: getRequestHeaders(event) as any, method, body })
}

export async function getFormData(event: H3Event) {
  return (await getRequestFromEvent(event)).formData()
}
```

Which directly enables great DX around forms : 

```ts
const eventHandler = event => {
	const formData = await getFormData(event)
	const email = formData.get("email") as string
	const password = formData.get("password") as string
	return { email }
}
```

### Additional information

Hebilicious:
- feat: add formdata and request helpers
- chore: rename to eventToRequest and readFormData
- feat: add readBodySafe and getQuerySafe helpers

### 🔗 Linked issue

Resolves #419
Resolves #402 

### ❓ Type of change

### 📚 Description

(This PR includes https://github.com/unjs/h3/pull/421)

This PR ads two new helpers to H3 

- readBodySafe
- getQuerySafe

Both of these helpers accept a https://github.com/decs/typeschema schema as the 2nd argument, which is compatible with zod, yup, joi, superstruct, typia, runtypes, arktype and custom validation function.
`readBodySafe` implementation will use FormData for `x-www-form-urlencoded`, and then rely on `safeDestr` for body parsing before passing it to the schema validation. By design, both these function will throw on runtime validation errors.

It would also be possible to take https://github.com/decs/typeschema and extract its logic into our own internal helpers or UnJS, as it's a very tiny and clever library (50 LOC). We could also look into creating a unjs version of this library that fits our needs in the future.

I'm looking for feedback for alternative APIs.

### 📚 Usage

Here's an example with a zod schema.

```ts
  eventHandler(async (event) => {
    const schema = z.object({
      firstName: z.string(),
      lastName: z.string(),
    });
    const data = await readBodySafe(event, schema);
    expectTypeOf(data).toMatchTypeOf<{
      firstName: string;
      lastName: string;
    }>();
    return { ...data };
  })
```

### 📝 Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,11 +168,13 @@ H3 has a concept of composable utilities that accept `event` (from `eventHandler
 
 - `readRawBody(event, encoding?)`
 - `readBody(event)`
+- `readBodySafe(event, schema, onError?)`
 - `readMultipartFormData(event)`
 
 #### Request
 
 - `getQuery(event)`
+- `getQuerySafe(event, schema, onError?)`
 - `getRouterParams(event)`
 - `getMethod(event, default?)`
 - `isMethod(event, expected, allowHead?)`
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "test": "pnpm lint && vitest run --coverage"
   },
   "dependencies": {
+    "@decs/typeschema": "^0.5.2",
     "cookie-es": "^1.0.0",
     "defu": "^6.1.2",
     "destr": "^2.0.0",
@@ -58,7 +59,8 @@
     "supertest": "^6.3.3",
     "typescript": "^5.1.6",
     "unbuild": "^1.2.1",
-    "vitest": "^0.33.0"
+    "vitest": "^0.33.0",
+    "zod": "^3.21.4"
   },
   "packageManager": "pnpm@8.6.9"
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -8,6 +8,9 @@ importers:
 
   .:
     dependencies:
+      '@decs/typeschema':
+        specifier: ^0.5.2
+        version: 0.5.2(zod@3.21.4)
       cookie-es:
         specifier: ^1.0.0
         version: 1.0.0
@@ -90,6 +93,9 @@ importers:
       vitest:
         specifier: ^0.33.0
         version: 0.33.0
+      zod:
+        specifier: ^3.21.4
+        version: 3.21.4
 
   playground:
     dependencies:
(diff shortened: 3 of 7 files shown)


pi0:
> (This PR includes https://github.com/unjs/h3/pull/421)

Please always target PRs against `main` 🙏🏼

Hebilicious:
> > (This PR includes #421)
> 
> Please always target PRs against `main` 🙏🏼

I'm gonna put this as pending until #421  is merged

Hebilicious:
@pi0 I've un-drafted this now that readFormData has landed 👍🏽

pi0:
While I love the type schema idea, it is beyond 50 LOC when bundled as it registers adapters. I agree with you, we might make an unjs similar library in order to achieve this because we also need validation fo db0, etc. 

I think for this to move forward, we need to fully decouple the validation function from the underlying library. So the `readValidatedBody` **requires** a validate function which can be wrapped into an adaptor via a subpath `h3/zod`, `h3/typeschema` or an external `h3-zod`, `h3-typeschema` to integrate with those libs.

Hebilicious:
> While I love the typeschema lib idea, it is beyond 50 LOC when bundled as it registers adapters. I agree with you, we might make an unjs similar library in order to achieve this because we also need validation fo db0, etc. (It might be a major version of unjs/untyped ~> [unjs/untyped#101](https://github.com/unjs/untyped/issues/101))
> 
> I think for this to move forward, we need to fully decouple the validation function from the underlying library. So the `readValidatedBody` **requires** a validate function which can be wrapped into an adaptor via a subpath `h3/zod`, `h3/typeschema` or an external `h3-zod`, `h3-typeschema` to integrate with those libs.

The first version that I tried had a very simple implementation that worked great : https://github.com/decs/typeschema/blob/v0.1.3/src/assert.ts

Maybe we could just inline that logic for now and support slightly less schemas ?

pi0:
I prefer to do it step by step and with zero dependencies added now. Such checks need proper test suite also.

decs:
hey, folks! `typeschema` maintainer here

we started with a simple logic, inspired by tRPC, but quickly noticed that it wouldn't support other libraries where the validation function lives outside of the schema object (like `io-ts`, `ajv`). so we pivoted to the adapter + optional peer dependencies model, and now we're on track to support all major libs this week. and indeed there's a tradeoff: the architecture change increased the LOC.

I can think of 2 alternatives to adding `typeschema` as a direct dependency, if the LOC is a concern:
1. making it an optional peer dependency. try dynamically importing `typeschema` (falling back to no schema validation if not installed) and instructing users to install it on their own
2. or supporting custom validators with this format: `<T>(data: unknown): Promise<T> | T`, then asking users to independently install `typeschema` and use `createAssert(schema)` (which generates a function on this format)

pi0:
Hi dear @decs thanks for your genius idea for simple util and nice library and explanations.

Indeed the schemas with decoupled validator util is another concern we have to take into account.

I think we still want to move forward with a direct option for custom validators and make wrapper composables in order to provide integrations in next strps (if unified, even better!). 

I have made a generic discussion in https://github.com/unjs/untyped/issues/101 to brainstorm how to make it a generic utility so we can use it across unjs ecosystem, if you like to weight in there and put any ideas how to do this.

pi0:
#459

PR #395, opened 2023-05-29
feat: Request/Response support, adapter cloudflare and streams

Hebilicious:
## 🔗 Linked issue

fix #394 
fix #401
fix #144

## ❓ Type of change

- [ ] 📖 Documentation (updates to the documentation, readme or JSdoc annotations)
- [ ] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [x] 👌 Enhancement (improving an existing functionality like performance)
- [x] ✨ New feature (a non-breaking change that adds functionality)
- [ ] 🧹 Chore (updates to the build process or auxiliary tools and libraries)
- [ ] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

## 📚 Description

Why ? 

- Supporting streaming on all platforms. 
- Allowing users to conveniently use h3 in environment that aren't node
- Returning Response objects from event handler in node environments 

I needed these features on one of my project, and I'm contributing them back here. 

### `event.request` alternative API 

Can be used when h3 is initialised with `createEvent(undefined, undefined, request)`.
This API allows you to return native Response objects.
This is more natural in environment that aren't node, such as cloudflare, lagon or deno.

```ts
  const app = createApp();
  const router = createRouter();
  router.get(
    "/hello",
    eventHandler((event) => {
      const request = event.request; // Request object, from the fetch API
      return new Response(`hello ${request.method}`);
    })
  );
  app.use(router);
  const response = await app.handler(
    createEvent(
      undefined,
      undefined,
      new Request(new URL("http://localhost/hello"))
    )
  );
  console.log(await response.text()); // hello GET
```
    
### Event handler Response support 

Supports Response and RawResponse to override setters values.
Helpers functions are usable instead: `sendResponse` and `sendResponseRaw`.
This works in node too, the Response() will be converted and sent with `res.write` and `res.end`.

```ts
      eventHandler(() => {
        setHeader(event, "hello", "yes"); // will take priority 
        setResponseStatus(event, 207, "hello-status"); // will take priority 
        return new Response("valid", {
          status: 208,
          statusText: "bye-status",
          headers: { hello: "no" },
        });
      })
```

```ts
      eventHandler((event) => {
        setHeader(event, "hello", "yes");
        setResponseStatus(event, 208, "bye-status");
        return new RawResponse("valid", {
          status: 207, // will take priority 
          statusText: "hello-status", // will take priority 
          headers: { hello: "world" }, // will take priority 
        });
      })
```
###  Streaming support 
[...]

PR #436, opened 2023-07-10
feat: web responses with streaming support

pi0:
### 🔗 Linked issue

#73

### ❓ Type of change

### 📚 Description

This PR adds support for directly returning a web [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) object from event handlers, leveraging new streaming support for body! (#432)

It is also possible to explicitly use the newly exposed `sendWebResponse` utility.

**Note:** The rest is pretty much simple and implementation is already used for `event.respondWith` (by @danielroe ❤️). I am thinking in next steps, reuse same utility for `respondWith` as well and deprecating pollyfills we initially made due to runtime incompatibilities. (https://github.com/unjs/h3/pull/119) 

Also thanks to @Hebilicious for initial efforts in #395 to push for supporting native Responses ❤️ 

### 📝 Checklist

### DIFF
--- a/test/app.test.ts
+++ b/test/app.test.ts
@@ -13,6 +13,7 @@ import {
 // Node.js 16 limitations
 const readableStreamSupported = typeof ReadableStream !== "undefined";
 const blobSupported = typeof Blob !== "undefined";
+const responseSupported = typeof Response !== "undefined";
 
 describe("app", () => {
   let app: App;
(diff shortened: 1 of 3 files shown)


Hebilicious (commented):
Thank you for adding this ❤️ 
I have a few questions : 

What happens with the response 2nd argument ?
```ts
        setResponseStatus(event, 208, "hello-status");
        return new Response("valid", {
          status: 207,
          statusText: "bye-status"
        });
```
Does it works with Transform Streams ?
```ts
        const response = new Response(`Hello world !`);
        const { readable, writable } = new TransformStream();
        response.body?.pipeTo(writable);
        return new Response(readable, response);
      })
```

It would be useful to have tests for theses.

pi0:
> What happens with the response 2nd argument ?

Status code from `Response` is expected to be picked as it is last operation. Normally we should set it in one place only :) 

> Does it works with Transform Streams?

Readable streams should be normally chainable. I think we can improve tests for `sendStream` utility instead as this feature reuses sendStream. PR Welcome to improve tests 💯

test/app.test.ts
```diff
@@ -33,6 +34,23 @@ describe("app", () => {
     expect(res.body).toEqual({ url: "/" });
   });
 
+  it.runIf(responseSupported)("can return Response directly", async () => {
+    app.use(
+      "/",
+      eventHandler(
+        () =>
+          new Response("Hello World!", {
```
image72, line 42: please document this feature demo in readme.

Issue #413, opened 2023-06-23
Corrupted binary file when using proxyRequest


Ngob:
### Environment

I am using nuxt.
In my project, I am using `h3: "1.7.0"`

Relevant packageJson (in the stackblitz)

```
  "dependencies": {
    "nuxt": "^3.5.2"
  }
```

### Reproduction

The issue is due to `proxyRequest` loading the request and converting it to utf8 before sending it to the proxified server.

It leads to a corrupted binary file.

A sample of the issue is available here

[https://stackblitz.com/edit/h3-nuxt-proxyrequest?file=server%2Fapi%2F%5B...%5D.ts](https://stackblitz.com/edit/h3-nuxt-proxyrequest?file=server%2Fapi%2F%5B...%5D.ts)

Or (on the branch main)

https://github.com/Ngob/h3-nuxt-proxyRequest-issue/

Step to reproduce the issue:

In the stackblitz, select a pdf file for the input "Using the proxy".
Then access to /dummy-sample.pdf

The file here is "empty"

To demonstrate the expected behavior, select a pdf file for the input "Without the proxy", then access to /dummy-sample.pdf. Here the pdf is displayed, that means it's not corrupted

I commited a `dummy.pdf` file at the root of the repository for easier testing

### Describe the bug

The issue looks similar to [373](https://github.com/unjs/h3/issues/373) and to [385](https://github.com/unjs/h3/issues/385)

The proxyRequest method read the body and use utf8 encoding leading to corrupted binary file when using the proxy
[relavant code](https://github.com/unjs/h3/blob/main/src/utils/proxy.ts#L39)
```
body = await readRawBody(event).catch(() => undefined);
```

and [here](https://github.com/unjs/h3/blob/main/src/utils/body.ts#L68C11-L68C11)
```
  const result = encoding
    ? promise.then((buff) => buff.toString(encoding))
    : promise;
```

In this code, casting the buffer to string using the default encoding (utf8) leads the proxy to convert the binary file and corrupting it

What leads me to link [373](https://github.com/unjs/h3/issues/373) is that using a large request (like uploading a video), the proxy loads the full content of the binary file in memory before sending it, leading to a very inefficient memory usage.

### Additional context

My current fix is to use the server request's stream as part of the body of the request to the backend server.

It allows to resolve my bug and it improve memory usage

Let me know if I can open a pull request with the fix.

[...]

PR #440, opened 2023-07-13
fix(proxy): avoid decoding request body as utf8

Ngob:
### 🔗 Linked issue
[#413](https://github.com/unjs/h3/issues/413)

### ❓ Type of change

### 📚 Description
Resolves [#413](https://github.com/unjs/h3/issues/413)

This pull-request add:
 - A test to test if the proxy preserve the PDF file when sending the file to the upstream service. I added a PDF file into the test/ressource directory
 - A fix that stream the request body to the upstream service. It allow better memory usage and prevent any conversion on the body

This PR consists into 2 commit: 
 - The first commit does add the test. In this commit you can see the test failing
 - The second commit does add the fix. In this commit, you can see the test not failing anymore

Let me know if I need to update the PR.
Any comments would be appreciated.

### 📝 Checklist

- [ ] I have updated the documentation accordingly (does not apply here).


Ngob:
Hello,

I added another test to improve the coverage that test if additionnals headers are being proxifyed along with the request header
I also tested on node version 16, 18 and 19 and added the "duplex: 'half'" to make it work using version 18/19. The pipeline should be fixed now.

Thanks for the review

pi0:
Hi and thanks for the PR. I think it is a good idea to support incoming request body streaming but it only works on Node.js and we have an ongoing draft for it (https://github.com/unjs/h3/pull/374). Is there any chance we might be able to fix `readRawBody` directly? (it is supposed to properly read "binary" request body into a Buffer if there is any possible bug with it..)

Ngob:
> Hi and thanks for the PR. I think it is a good idea to support incoming request body streaming but it only works on Node.js and we have an ongoing draft for it (#374). Is there any chance we might be able to fix `readRawBody` directly? (it is supposed to properly read "binary" request body into a Buffer if there is any possible bug with it..)

Yes I can do
So what I need to do is to restore the usage of readRawBody and fix the method when using binaries?
So I will update tests to check that readRawBody has the correct behavior

pi0:
Thanks for helping on this by providing reproduction, explanations and fixtures ❤️ 

I have made a refactor to simplify fix (We need to pass `false` as second argument to `readRawBody` to make it reading body as a stream).

Using proxy for incoming requests seems a good idea too! Leveraging `event.request` introduced in #454 we can implement this feature in #374 (an experimental flag at first)

Ngob:
Thanks a lot for the followup!

PR #441, opened 2023-07-14
fix(proxy): respect `fetchOptions.method` over incoming request method

emilsgulbis:
### ❓ Type of change

### 📚 Description
When `proxyRequest` is used with different `method` (e.g. GET) defined in `fetchOptions`, the `body` is still filled and returns `Request with GET/HEAD method cannot have body.`

### 📝 Checklist

### DIFF
--- a/src/utils/proxy.ts
+++ b/src/utils/proxy.ts
@@ -31,7 +31,7 @@ export async function proxyRequest(
   opts: ProxyOptions = {}
 ) {
   // Method
-  const method = getMethod(event);
+  const method = opts.fetchOptions?.method ?? getMethod(event);
 
   // Body
   let body;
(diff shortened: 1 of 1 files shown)


NozomuIkuta:
@emilsgulbis 

Would you add unit test?
https://github.com/unjs/h3/blob/main/test/proxy.test.ts

src/utils/proxy.ts
```diff
@@ -31,7 +31,7 @@ export async function proxyRequest(
   opts: ProxyOptions = {}
 ) {
   // Method
-  const method = getMethod(event);
+  const method = opts.fetchOptions?.method ?? getMethod(event);
```
pi0, line 34: Just in case to handle edge case of method: `""`
```suggestion
  const method = opts.fetchOptions?.method || getMethod(event);
```

pi0 (approved):
Thanks!

pi0:
Adding an additional test would be nice btw (however is not necessary)

PR #461, opened 2023-07-25
fix(router): fallback for method-shadowed routes

pi0:
### ❓ Type of change

### 📝 Checklist

### DIFF
--- a/test/router.test.ts
+++ b/test/router.test.ts
@@ -97,6 +97,27 @@ describe("router", () => {
     const res = await request.get("/404");
     expect(res.status).toEqual(404);
   });
+
+  it("Handle shadowed route", async () => {
+    router.post(
+      "/test/123",
+      eventHandler((event) => `[${event.method}] ${event.path}`)
+    );
+
+    router.use(
+      "/test/**",
+      eventHandler((event) => `[${event.method}] ${event.path}`)
+    );
+
+    const postRed = await request.post("/test/123");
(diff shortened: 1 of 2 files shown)


test/router.test.ts
```diff
@@ -97,6 +97,29 @@ describe("router", () => {
...
   });
+
+  it("Handle shadowed route", async () => {
+    router.post(
+      "/test/123",
+      eventHandler((event) => `[${event.method}] ${event.path}`)
+    );
+
+    router.use(
+      "/test/**",
```
Hebilicious, line 108: Should we test with "**" as its closer to the Nuxt behaviour ?

Hebilicious (commented):
This looks great !
I was thinking if the name catchAll wouldn't be more fitting than shadow, but maybe there's a use case that doesn't involve `**` that I'm not aware of ?

test/router.test.ts
pi0: I wanted to also make sure base URL works. Notice that "both" have test in prefix. If base works, without certainly works too :)

pi0:
> I was thinking if the name catchAll wouldn't be more fitting than shadow, but maybe there's a use case that doesn't involve ** that I'm not aware of ?

All nitro route rules use this feature to catch matching paths :)

---

Marking PR as draft since I'm not still sure if this is best approach for fixing the issue

PR #462, opened 2023-07-26
fix(event): do not remove double slashes from query

Aareksio:
### 🔗 Linked issue

https://github.com/unjs/h3/issues/463
(also see added test case)

### ❓ Type of change

### 📚 Description

Currently the regex replaces all occurrences of multiple `/` with a single `/`. This is desired, but only for the `path`, not `query` part of the URL. This PR ensures the replacement is only done to the part of path coming before `?`. There is no `ufo` helper for extracting just the path nor for detecting if path includes query - therefore I have used good old `.includes('?')` and `.split('?')`.

There is [cleanDoubleSlashes](https://github.com/unjs/ufo/blob/main/src/utils.ts#L92-L97) available in ufo, but it seems to suffer from the same issue. May be worth fixing there and updating h3 to use ufo helper.

Resolves https://github.com/unjs/h3/issues/463

### 📝 Checklist

### DIFF
--- a/src/event/event.ts
+++ b/src/event/event.ts
@@ -61,11 +61,11 @@ export class H3Event implements Pick<FetchEvent, "respondWith"> {
 
   get path() {
     if (!this._path) {
-      const hasQuery = this._originalPath.includes('?');
+      const hasQuery = this._originalPath.includes("?");
 
       if (hasQuery) {
-        const [basePath, query] = this._originalPath.split('?');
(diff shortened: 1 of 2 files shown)


pi0 (approved):
Thanks!

pi0:
> There is [cleanDoubleSlashes](https://github.com/unjs/ufo/blob/main/src/utils.ts#L92-L97) available in ufo, but it seems to suffer from the same issue. May be worth fixing there and updating h3 to use ufo helper.

Upstream PR welcome for ufo too 💯

PR #466, opened 2023-07-27
fix(proxy): transparently forward errors when passing ofetch

Aareksio:
### 🔗 Linked issue

https://github.com/unjs/h3/issues/464

### ❓ Type of change

- [x] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

As discussed in https://github.com/unjs/h3/issues/464 and https://github.com/unjs/nitro/pull/1489.

`ProxyOptions` type updated to allow overriding the default value (`true`).

---

Is it **breaking change**?

I would say so, users may depend on error masking. This is not strictly a bug. Mandatory xkcd: https://xkcd.com/1172/

---

I personally dislike this solution a lot; if we allow freedom of passing fetch implementation, it should be up to the callee to configure the behavior of their implementation exactly as they like. If it's required to pass option to do so, they already can do so with `fetchOptions`.

Forcing `ignoreResponseError: true` is implicitly re-configuring non-default user-passed implementation. I think better approach is to allow the freedom along with passing the responsibility. Another edge-case is when user passes fetch implementation which also supports `ignoreResponseError` or throws on unknown parameters - both sound plausible to me.

### 📝 Checklist

### DIFF
--- a/src/utils/proxy.ts
+++ b/src/utils/proxy.ts
@@ -9,7 +9,7 @@ export type Duplex = "half" | "full";
 
 export interface ProxyOptions {
   headers?: RequestHeaders | HeadersInit;
-  fetchOptions?: RequestInit & { duplex?: Duplex };
+  fetchOptions?: RequestInit & { duplex?: Duplex } & { ignoreResponseError?: boolean };
   fetch?: typeof fetch;
   sendStream?: boolean;
   streamRequest?: boolean;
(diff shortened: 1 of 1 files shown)


Aareksio:
Either solution solves the problem. If you like it, that's good enough! Much better title too! ❤️

I am still a little worried about this one guy who uses error masking to hide secrets from consumers, but if you consider it non breaking, be it 😂

pi0:
You can try it on nightly channel now (check little new section in readme!)

PR #483, opened 2023-08-02
feat: plain and web adapters

### LINKED ISSUE #458: Fetch Adapter
The fetch adapter is the last piece of the puzzle for #401 .
For now `createEvent` expect a node.js style event. For platforms that are not node.js based, this creates un-necessary overhead, and a fetch adapter would allow to construct an H3 event from a Request.

pi0:
### 🔗 Linked issue

Resolves #458

### ❓ Type of change

### 📚 Description

This PR adds two new plain and web adapters to use h3 directly outside of Node.js environments!

Until now, it was only possible to use h3 in workers and lambda environments by either using Nitropack or custom code. With two new adapters, integration will be much more straightforward.

Two new utils depend on [unenv](https://github.com/unjs/unenv) framework to keep backward compatibility with Node.js API. This will be fully optional in the next major version of H3 once all h3 internal utils were decoupled.

### 📝 Checklist


pi0:
Let's try it!

src/adapters/plain.ts
```diff
@@ -0,0 +1,132 @@
...
+// --- Internal ---
+
+export async function _handlePlainRequest(app: App, request: PlainRequest) {
+  // Normalize request
+  const path = request.path;
+  const method = (request.method || "GET").toUpperCase() as HTTPMethod;
+  const headers = new Headers(request.headers);
+
+  // Shim for Node.js request and response objects
+  // TODO: Remove in next major version
```
magne4000, line 45: What does this actually mean? I did not find any other information regarding this comment.
Does this mean that internal representation as `IncomingMessage/ServerResponse` will change at some point? If yes, is there discussions somewhere on the subject? If no, what was the idea?

For the context, I'm [experimenting on generic handlers](https://github.com/hattipjs/hattip/discussions/45#discussioncomment-9342511) (and middlewares), and currently listing how each popular server implements handlers and middlewares.

magne4000: Some more context on a dedicated repo https://github.com/magne4000/universal-handler

PR #485, opened 2023-08-03
feat: object-syntax event handlers

danielroe:
### 🔗 Linked issue

https://github.com/unjs/h3/pull/417
resolves https://github.com/unjs/h3/issues/424

### ❓ Type of change

### 📚 Description

This supports object-syntax event handlers with the following syntax:

```ts
defineEventHandler({
  handler: () => {},
  before: [],
  after: []
})
```

Helper validate functions need to be drafted that type-hint `event` and infer a return type. (I would suggest we also suggest `validate: { body: () => {}, query: () => {} }` which I think will be easier to directly get a type from as user can just return the validated body/query, which we can then flow through into rest of handlers.)

### 📝 Checklist

### DIFF
--- a/src/event/utils.ts
+++ b/src/event/utils.ts
@@ -21,7 +21,7 @@ export function defineEventHandler<
   Request = EventHandlerRequest,
   Response = EventHandlerResponse,
 >(
-  def: EventHandler<
+  handler: EventHandler<
     Request extends EventHandlerRequest ? Request : any,
     Request extends EventHandlerRequest ? Response : Request
   >,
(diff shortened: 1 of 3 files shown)


src/event/utils.ts
```diff
@@ -25,10 +36,34 @@ export function defineEventHandler<
...
+    : EventHandlerRequest,
+>(
+  handler: EventHandlerInput<Request, Response, Validator, ValidatedRequest>,
+): EventHandler<Request, Response> {
+  if (typeof handler === "function") {
+    return Object.assign(handler, { __is_handler__: true });
+  }
+  const wrapper: EventHandler<Request, any> = async (event) => {
+    if (handler.validate) {
+      await validateData(event, handler.validate);
```
pi0, line 54: Let's not include validate in first iteration. I am worried (even if small, this adds to the bundle when not validation is needed). We can rethinking about adding more proper in next iterations 👍🏼

src/event/utils.ts
```diff
@@ -25,10 +36,34 @@ export function defineEventHandler<
...
+    }
+    for (const hook of handler.before || []) {
+      await hook(event);
```
pi0, line 57: We need to stop search when first handler sets `event.handled`

src/event/utils.ts
```diff
@@ -25,10 +36,34 @@ export function defineEventHandler<
...
+  handler: EventHandlerInput<Request, Response, Validator, ValidatedRequest>,
+): EventHandler<Request, Response> {
+  if (typeof handler === "function") {
+    return Object.assign(handler, { __is_handler__: true });
+  }
+  const wrapper: EventHandler<Request, any> = async (event) => {
+    if (handler.validate) {
+      await validateData(event, handler.validate);
+    }
+    for (const hook of handler.before || []) {
```
pi0, line 56: Tiny perf: Maybe we can initiate iterator when `handler.before` is defined?

src/event/utils.ts
```diff
@@ -25,10 +36,34 @@ export function defineEventHandler<
...
+      await hook(event);
+    }
+    const result = await handler.handler?.(event);
+    for (const hook of handler.after || []) {
```
pi0, line 60: Tiny perf: We can initiate `after` also when defined

src/event/utils.ts
```diff
@@ -25,10 +36,34 @@ export function defineEventHandler<
...
+      await hook(event);
```
pi0, line 61: Would be nice to also pass `, { body }` to the hooks to intercept the response. (body, to be consistent with global hooks)

src/event/utils.ts
danielroe: Makes sense! Mainly I wanted to confirm that the types would work with inferring typed request _from_ a provided validator function.

pi0 (approved):

PR #496, opened 2023-08-07
feat: request validation using handler object syntax `validate` and `validateEvent(event)`

danielroe:
### ❓ Type of change

### 📚 Description

This allows inferring type of event by passing a validate function which needs to return the event (as validated). It's simple and allows any kind of validation, throwing an error if invalid input is provided.

### 📝 Checklist


pi0:
I am wondering if we can alternatively support `before:[validate()]` to allow multiple validate functions and avoid adding new hook.

danielroe:
The issue is the typing. It might be possible but will be a lot more complex, to create chainable validators. (So output of one becomes input of another.)

pi0:
It makes sense for type complexity

I am thinking it could give some overhead for when we want to do body normalization, we actually read the body (and an external library can apply schema normalization). Only returning a boolean + typed Event will cause to loose this information (unless the utils support a context caching mechanism for when we later call `useValidatedBody`).

Also, it would be nice if we expose a generic utility `async validateEvent` to extract this functionality and also make it available for non object syntax usages.

src/event/event.ts
```diff
@@ -19,13 +19,16 @@ export interface NodeEventContext {
 export class H3Event<
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
   _RequestT extends EventHandlerRequest = EventHandlerRequest,
+  Context = unknown,
```
pi0, line 22: I am thinking we might support `context` in `EventHandlerRequest` to avoid introducing new generic and increase portability of types

danielroe: I considered that initially. The problem is that this means that the _incoming_ request is typed as requiring or having a context, and will also make typing `$fetch` request considerably more complex.

For example, an `EventHandlerRequest<{ body: { id: string } }>` would not be assignable to `EventHandlerRequest<{ body: { id: string }, context: { other: true } }>` even though you can't _control_ the context from the thing initiating the _request_.

src/event/utils.ts
```diff
@@ -57,6 +81,23 @@ async function _callHandler(event: H3Event, handler: EventHandlerObject) {
...
   }
+  if (handler.validate) {
+    const res = (await validateEvent(event, handler.validate)) as Record<
+      string,
+      any
+    >;
+    if (res && typeof res === "object") {
+      for (const property in res) {
+        if (property === "body") {
+          (event.node.req as any)[ParsedBodySymbol] = res.body;
```
pi0, line 92: We might keep readBody utils decoupled from this new feature of context injection. (the read options could be different)

src/event/event.ts
pi0: I see (this also when locally trying to refactor now). I think once we have a chance before 1.8 release to introduce typed event, we might merge `body` and `query` types to the context so that `EventHandlerRequest<T>` the `T` generic is describing a validated event context (instead of two places)

danielroe: What do you mean by:
> we might merge `body` and `query` types to the context so that `EventHandlerRequest<T>` the `T` generic is describing a validated event context (instead of two places)

If it means what I think, then I would suggest we do not do this. I think it's important to separate out the incoming from outgoing. So the flow is:

* request comes in (requiring certain inputs)
* new 'validate' function runs, and _also_ returns values to context
* event is now typed with body/query and also context

As far as the request is concerned, the context typing is irrelevant, and it would make inference harder in Nitro and any other framework using EventHandler types if the context typing is merged with body/query.

pi0: (context, is also part of incoming) i will explain better in out call :)

pi0:
Update: Both me and @danielroe need to rethink about whole typing and validation and how it works with downstream (Nitro and Nuxt) types and best type/runtime safely+perf balance which is tricky at the moment. Making it draft until after 1.8 to iterate.

pi0:
(closing as object syntax removed, we need to build it based on new middleware format)

PR #503, opened 2023-08-10
feat: util `getRequestIP` 

harlan-zw:
### 🔗 Linked issue

https://github.com/unjs/h3/issues/272

### ❓ Type of change

### 📚 Description

We should make it convenient for users to get the client request ip while being able to opt-in to trust the x-forwarded-for header.

Currently, users are implementing themselves:
- https://github.com/Baroshem/nuxt-security/blob/main/src/runtime/server/middleware/rateLimiter.ts#L7C14-L7C30
- https://github.com/timb-103/nuxt-rate-limit/blob/master/src/runtime/server/utils/rate-limit.ts#L81

I'm also open to naming this `getClientIp` or `getRequestClientIp`.

This utility does open the door for trusting spoofable headers (see nuxt-security), sorting that is out of the scope of this utility, I'll open this as a seperate issue.

### 📝 Checklist

### DIFF
--- a/src/utils/request.ts
+++ b/src/utils/request.ts
@@ -158,10 +158,10 @@ export function getRequestIP(
      *
      * Note: This is easily spoofed, make sure that this header can be trusted before enabling.
      */
-    xForwardedFor?: boolean
+    xForwardedFor?: boolean;
   } = {},
 ) {
-  const nonProxyIp = event.context.clientAddress
+  const nonProxyIp = event.context.clientAddress;
 
   if (!opts.xForwardedFor) {
     return nonProxyIp;
--- a/src/utils/request.ts
+++ b/src/utils/request.ts
@@ -160,15 +160,21 @@ export function getRequestIP(
      */
     xForwardedFor?: boolean;
   } = {},
-) {
-  const nonProxyIp = event.context.clientAddress;
-
-  if (!opts.xForwardedFor) {
-    return nonProxyIp;
+): string | undefined {
+  if (event.context.clientAddress) {
+    return event.context.clientAddress;
   }
 
-  const xForwardedFor = getRequestHeader(event, "x-forwarded-for")
-    ?.split(",")
-    ?.pop();
-  return xForwardedFor || nonProxyIp;
+  if (opts.xForwardedFor) {
+    const xForwardedFor = getRequestHeader(event, "x-forwarded-for")
+      ?.split(",")
+      ?.pop();
+    if (xForwardedFor) {
+      return xForwardedFor;
+    }
+  }
+
+  if (event.node.req.socket.remoteAddress) {
+    return event.node.req.socket.remoteAddress;
+  }
 }
--- a/src/utils/request.ts
+++ b/src/utils/request.ts
@@ -154,14 +154,15 @@ export function getRequestIP(
   event: H3Event,
   opts: { xForwardedFor?: boolean } = {},
 ) {
-  const ip = (event.node.req.connection as any).remoteAddress;
+  const nonProxyIp = (event.node.req.connection as any).remoteAddress;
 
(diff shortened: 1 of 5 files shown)


src/utils/request.ts
```diff
@@ -149,3 +149,20 @@ export function getRequestURL(
...
   return new URL(path, `${protocol}://${host}`);
 }
+
+export function getRequestIP(
+  event: H3Event,
+  opts: { xForwardedFor?: boolean } = {},
+) {
+  const nonProxyIp = (event.node.req.connection as any).remoteAddress;
+
+  if (!opts.xForwardedFor) {
```
pi0, line 159: I think we should make it opt-in strictly to avoid issues.

src/utils/request.ts
```diff
@@ -149,3 +149,20 @@ export function getRequestURL(
   const path = getRequestPath(event);
   return new URL(path, `${protocol}://${host}`);
 }
+
+export function getRequestIP(
+  event: H3Event,
+  opts: { xForwardedFor?: boolean } = {},
+) {
+  const nonProxyIp = (event.node.req.connection as any).remoteAddress;
```
pi0, line 157: We shall support (and prefer + type) `event.context.clientAddress` to support integration with nodeless runtimes.

pi0:
Thanks for PR! I love it it is a must have util!

/cc @danielroe @Atinux @antfu about naming ideas. I like the current `getRequestI^P` btw :)

Atinux:
Nice!

I like `getRequestIP()`

danielroe:
Agreed on `getRequestIP` :+1:

And nice work.

src/utils/request.ts
```diff
@@ -149,3 +149,20 @@ export function getRequestURL(
...
+
+  if (!opts.xForwardedFor) {
+    return nonProxyIp;
+  }
+
+  const xForwardedFor = getRequestHeader(event, "x-forwarded-for")
+    ?.split(",")
+    ?.pop();
+  const ip = xForwardedFor || nonProxyIp;
+  return ip.split(":")[0];
```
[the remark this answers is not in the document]
harlan-zw: Good point, I was reading another implementation that was handling ports so included this. Not sure if there's a simple way to handle ports with the ipv6 though, any ideas?

src/utils/request.ts
harlan-zw: Good point, can you see if I set this up correctly.

src/utils/request.ts
harlan-zw: This is opt-in isn't it

harlan-zw:
With the introduction of this utility, it's important to consider this issue https://github.com/unjs/h3/issues/504

As you can see in the issue, it's common for module authors to be trusting easily spoofed headers atm.

src/utils/request.ts
```diff
@@ -149,3 +149,32 @@ export function getRequestURL(
...
+    const xForwardedFor = getRequestHeader(event, "x-forwarded-for")
+      ?.split(",")
+      ?.pop();
+    if (xForwardedFor) {
+      return xForwardedFor;
+    }
+  }
+
+  if (event.node.req.socket.remoteAddress) {
+    return event.node.req.socket.remoteAddress;
```
pi0, line 178: I liked the idea that you moved it to node adopter but not sure at this stage specially because it should be last in priority (for Node.js source) and `clientAddress` set by custom user middleware, first. We might later use it as `event.context._incomingClientAdress` or something to fully decouple Node.js

pi0 (approved):

src/utils/request.ts
harlan-zw: Thanks, this looks like a better solution for now :+1:

PR #507, opened 2023-08-14
feat: `defineRequestMidleware`, `defineResponseMiddleware` and rename object synctax hooks

pi0:
### 🔗 Linked issue

Context: Find most of DX issues while working on a example for https://github.com/unjs/h3/issues/501

### ❓ Type of change

### 📚 Description

This PR introduces two new define utils `defineRequestMidleware` and `defineResponseMiddleware` (with `RequestMiddleware` and `ResponseMiddleware`) types.

These utils are then used for event handlers with object syntax but can also used independently to define such hooks outside of the event handler in a composable way (imagine `compressionMiddleware` as util)

Also in order to keep consistency with global hook options and readability, the hooks are renamed from `before`/`after` to `onRequest`/`beforeResponse`.

Also we support non Array format for most consistency with top level

### 📝 Checklist

### DIFF
--- a/src/event/utils.ts
+++ b/src/event/utils.ts
@@ -4,8 +4,15 @@ import type {
   EventHandlerRequest,
   EventHandlerResponse,
   EventHandlerObject,
+  RequestMiddleware,
+  ResponseMiddleware,
 } from "../types";
 import type { H3Event } from "./event";
+
+type _EventHandlerHooks = {
+  onRequest?: RequestMiddleware[];
+  beforeResponse?: ResponseMiddleware[];
+};
 
 export function defineEventHandler<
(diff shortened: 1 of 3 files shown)


danielroe (commented):
This looks good to me, from an API perspective.

Note that we will probably need to change the signature of these types before merging any kind of validation support. You could already go ahead and add the appropriate generics (even if unused) in the hook types so we don't break anything when implementing later.

pi0:
These middleware are same as current before/after. Do you mind to push any required placeholder generic?

danielroe (approved):

danielroe:
I think prefixing with underscore would be enough to mark them as internal types for now.

PR #518, opened 2023-08-22
fix: send original Error to custom error handlers instead of H3Error

### LINKED ISSUE #517: Can the original exception be passed to the onError handler instead of H3Error?
### Environment

1.8.0

### Reproduction

Throw any custom error inside of an eventHandler(). Example:
```javascript
class MyCystomError extends Error {
  public myCustomParam
  constructor(someParam){
     super("My custom error")
     this.myCustomParam = someParam
  }
}

  throw new MyCystomError({myData:true})
})

```

### Describe the bug

The [app option "onError" handler](https://github.com/unjs/h3/blob/164f68e9cb89e011cd304dbe07b328611e3428a3/src/app.ts#L52) is not passed [the original error "_error"](https://github.com/unjs/h3/blob/164f68e9cb89e011cd304dbe07b328611e3428a3/src/adapters/node.ts#L59) but instead a [H3Error "error"](https://github.com/unjs/h3/blob/164f68e9cb89e011cd304dbe07b328611e3428a3/src/adapters/node.ts#L66C10-L66C46). 
This results in me not being able to access `error.myCustomParam` from my custom ErrorHandler.

Would it be OK to pass the oringal error? Since we could always wrap the original error into a H3Error if we want to? 

### Additional context

_No response_

### Logs

_No response_

MarijnMensinga:
### 🔗 Linked issue

#517

### ❓ Type of change

- [X] 👌 Enhancement (improving an existing functionality like performance)
- [X] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

### 📚 Description

The custom errorHandler option currently gets passed a `H3Error` instead of the original `Error` in order for the
custom errorHandler to handle specific errors it should get passed the original `error`. This PR fixes this.
Resolves #517

This could break currently existing errorHandlers that expect an H3Error to be passed. Those can be fixed very easily by
calling `createError(error)`

### 📝 Checklist

- [X] I have linked an issue or discussion.
- [X] I have updated the documentation accordingly. _(not applicable)_

### DIFF
--- a/src/adapters/node.ts
+++ b/src/adapters/node.ts
@@ -63,7 +63,7 @@ export function toNodeListener(app: App): NodeListener {
       }
 
       if (app.options.onError) {
-        await app.options.onError(error, event);
+        await app.options.onError(_error, event);
       }
       if (event.handled) {
         return;
--- a/src/adapters/plain.ts
+++ b/src/adapters/plain.ts
@@ -101,7 +101,7 @@ export async function _handlePlainRequest(app: App, request: PlainRequest) {
       error.unhandled = true;
     }
     if (app.options.onError) {
-      await app.options.onError(error, event);
+      await app.options.onError(_error, event);
     }
     if (!event.handled) {
       if (error.unhandled || error.fatal) {
--- a/src/app.ts
+++ b/src/app.ts
@@ -49,7 +49,7 @@ export interface AppUse {
 
 export interface AppOptions {
   debug?: boolean;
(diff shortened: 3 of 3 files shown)


pi0:
Hi dear @MarijnMensinga sorry for checking on this late. As I have explained in https://github.com/unjs/h3/issues/517#issuecomment-1694486230, you can always use `error.cause` to access the original error(s). 

It was an intentional behavior but would be happy to rediscuss this if you think is still a wrong behavior.

PR #544, opened 2023-09-24
feat: add websocket support utils, and app handling

nksaraf:
depends on https://github.com/unjs/listhen/pull/128

adds websocket support by giving a few helpers, highlighted here. This can be added at the top of an event handler, and its a web socket serevr ready to do with all handling logic in your hand.

```ts

if (isWebSocketEvent(event)) {
      const wsEvent = toWebSocketEvent(event);
      if (wsEvent.type === "connection") {
        wsEvent.connection.send("pong");
      } else if (wsEvent.type === "message") {
        console.log("got", new TextDecoder().decode(wsEvent.message));
      }
    }
    if (isWebSocketUpgradeRequest(event)) {
      return upgradeWebSocket(event);
    }

```

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "test:types": "vitest typecheck"
   },
   "dependencies": {
+    "@types/ws": "^8.5.5",
     "cookie-es": "^1.0.0",
     "defu": "^6.1.2",
     "destr": "^2.0.1",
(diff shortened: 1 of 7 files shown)


TheEhsanSarshar:
@pi0 is there any chance to merge this PR

pi0:
Thanks for PR. ws is a Node.js dependency and we aim for cross platfom/runtime. Closing for triage.

PR #548, opened 2023-09-29
feat: add `getRequestFingerprint` util

nandi95:
closes https://github.com/unjs/h3/issues/536

### DIFF
--- a/README.md
+++ b/README.md
@@ -312,6 +312,15 @@ PRs are welcome to add your packages.
 - [h3-valibot](https://github.com/intevel/h3-valibot)
   - `useValidateBody(event, schema)`
   - `useValidateParams(event, schema)`
+- [@intlify/h3](https://github.com/intlify/h3)
+  - `defineI18nMiddleware(options)`
+  - `useTranslation(event)`
+  - `getHeaderLocale(event, options)`
+  - `getHeaderLocales(event, options)`
+  - `getCookieLocale(event, options)`
+  - `setCookieLocale(event, options)`
+  - `getPathLocale(event, options)`
+  - `getQueryLocale(event, options)`
 
 ## License
 


src/utils/request.ts
```diff
@@ -190,3 +191,22 @@ export function getRequestIP(
     return event.node.req.socket.remoteAddress;
   }
 }
+
+export async function getFingerprint(event: H3Event): Promise<string> {
+  let fingerprint = event.toString();
+  const ip = getRequestIP(event, { xForwardedFor: !!getHeader(event, 'x-forwarded-for') });
```
pi0, line 197:
```suggestion
  const ip = getRequestIP(event, { xForwardedFor: true });
```

pi0 (approved):

nandi95:
This seems good to me. I don't get why have the path and method opt in. Unlike userAgent, they cannot be spoofed?

Side note, perhaps this or another utility can account for the whole request as is, including the query string, body and headers. Such a thing would be useful if someone had to de-duplicate requests (if that is even a thing) or perhaps some anonymised server side analytics.

src/utils/request.ts
```diff
@@ -195,3 +196,68 @@ export function getRequestIP(
     return event.node.req.socket.remoteAddress;
   }
 }
+
+export interface RequestFingerprintOptions {
+  /** @default SHA-1 */
+  hash?: false | "SHA-1";
```
nandi95, line 202:
```suggestion
  hash?: false | AlgorithmIdentifier;
```

src/utils/request.ts
```diff
@@ -195,3 +196,68 @@ export function getRequestIP(
...
+    fingerprint.push(event.path);
+  }
+
+  if (opts.userAgent === true) {
+    fingerprint.push(getRequestHeader(event, "user-agent"));
+  }
+
+  const fingerprintString =
+    fingerprint.filter(Boolean).join("|") ||
+    Math.random().toString(36).slice(2);
```
nandi95, line 247:
```suggestion
  if (!fingerprint.length) {
    throw new RangeError('At least one fingerprinting factor must be enabled.')
  }

  const fingerprintString = fingerprint.filter(Boolean).join("|");
```

nandi95: I would rather have an error thrown here, warning the developer. Otherwise, they might end up with different fingerprint for the same request.
(Range error is somewhat applicable, but it can be an error too)

pi0: We might instead disable random behavior and return null allowing usage to handle it for custom error wdty? (usually h3 utils are errorless)

pi0:
> Side note, perhaps this or another utility can account for the whole request as is, including the query string, body and headers. Such a thing would be useful if someone had to de-duplicate requests (if that is even a thing) or perhaps some anonymised server side analytics.

For an analytics usecase, uniqueness of visitors is important yes, but i guess it might need to keep path unhashed otherwise it would be useless.

Mainly enabling these is dangerous for main purpose i could assume (security) because for example someone can change identity by iterating user agent or appending a query param to the end of url.

Once we had better vision (util is still in experimental state) we might improve defaults or have an option that enables all/group of defaults

src/utils/request.ts
nandi95: That's better too than a random string. In this case, the return value should be dependent on the options' argument type. Just so the user doesn't have to assert that the fingerprint is in fact a string.

danielroe (approved):

nandi95:
Opened a new PR from a branch that isn't named main as pull bot keeps nullifying the work.
https://github.com/unjs/h3/pull/564
I think this can be closed.

src/utils/request.ts
nandi95: shall I make this change?

PR #554, opened 2023-10-07
feat: add middleware wrapper for defineEventHandler

### LINKED ISSUE #553: defineEventHandler with middleware
### Describe the feature

I know h3 has the `defineRequestMiddleware` and `defineResponseMiddleware` utils, but I was thinking about maybe creating an overload for defineEventHandler that would accept middleware to run before/after the request. I have the main structure in mind already:
`defineEventHandler(event /*this is the regular event that is used in regular endpoints*/, requestMiddleware[]?, responseMiddleware[]?)`.
This would be like useFetch in [Nuxt](https://nuxt.com/docs/api/composables/use-fetch#usefetch), and could serve as a convenient wrapper around 
```ts
defineEventHandler({
  onRequest:defineRequestMiddleware(()=>{}),
  handler:defineEventHandler(()=>{}),
  onBeforeResponse:defineResponseMiddleware(()=>{})
})
```
I was about to open a PR but I wanted to first discuss it and whether or not it is possible/useful. I will gladly help implement this feature if it's deemed fit to add.
NOTE: There is also a possibility to only add ...middleware[] as the second argument and in them provide an arg that determines whether it's a request middleware or a response middleware.
Would love to hear your opinion on this!

### Additional information

GalacticHypernova:
### 🔗 Linked issue

Resolves #553 

### ❓ Type of change

### 📚 Description

This change extends the functionality of defineEventHandler and allows to utility different middleware in routes using vararg syntax, offering a convenient alternative to the object syntax. This resolves #553 
NOTE: This is still a prototype, and is in no way ready to review just yet. I'm still figuring out how to integrate it correctly.
### 📝 Checklist

### DIFF
--- a/src/event/utils.ts
+++ b/src/event/utils.ts
@@ -45,7 +45,7 @@ export function defineEventHandler<
   Response = EventHandlerResponse,
 >(
   handler: EventHandler<Request, Response>,
-  ...middleware: Array<_RequestMiddleware[]>
+  ...middleware: Array<_RequestMiddleware>
 ): EventHandler<Request, Response>;
 
 export function defineEventHandler<
--- a/src/event/utils.ts
+++ b/src/event/utils.ts
@@ -45,7 +45,7 @@ export function defineEventHandler<
   Response = EventHandlerResponse,
 >(
   handler: EventHandler<Request, Response>,
-  ...middleware: Array<_RequestMiddleware>
+  ...middleware: Array<_RequestMiddleware|_ResponseMiddleware>
 ): EventHandler<Request, Response>;
 
 export function defineEventHandler<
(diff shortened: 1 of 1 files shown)


GalacticHypernova:
@pi0 I wanted to ask for your opinion on whether or not it's a reasonable approach.
According to Nuxt's [documentation](https://nuxt.com/docs/guide/directory-structure/server), server middleware are "global" by default (as if you wrote a `.global.ts` middleware in `~/middleware`). How about we change nitro's behavior (or maybe it's a downstream thing we need to configure in nuxt) to allow (and respectively read) .global.ts in ~/server/middleware so that they could be global, whereas those that aren't prefixed with .global would run whenever they're called? That would help keep the server structure organized as there won't be a need to move non-global middleware outside of the middleware directory.

pi0:
Hi dear @GalacticHypernova thanks for issue and PR and sorry for late reply.

I have briefly explained some of my comments in https://github.com/unjs/h3/issues/553#issuecomment-1845200074

Nitro middleware pattern is also something we might deprecate at some point in favor of better controlled per request route rules and lifecycle hooks using object syntax.

GalacticHypernova:
@pi0 does that mean all function syntax should be replaced with object syntax? What would be the performance differences?

PR #572, opened 2023-11-08
test(proxy): Add failing test for forwarding cookies with redirects

Aareksio:
### ❓ Type of change

### 📚 Description

Stumbled upon this testing redirects against https://httpbin.org/#/Cookies/.

The issue is two fold:
- the cookie is not forwarded to the client at all
- the cookie is not forwarded when fetching new location

Tbh I am not sure if such complex and niche scenario should be handled by the lib at all. One could use `{ redirect: 'manual' }` when `set-cookie` + redirect is expected and handle it in userland.

### 📝 Checklist

### DIFF
--- a/test/proxy.test.ts
+++ b/test/proxy.test.ts
@@ -13,6 +13,8 @@ import {
   readRawBody,
   setCookie,
   setResponseHeader,
+  getCookie,
+  sendRedirect,
 } from "../src";
 import { sendProxy, proxyRequest } from "../src/utils/proxy";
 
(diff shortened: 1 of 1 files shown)


pi0:
Hi dear @Aareksio thanks for spending time on making PR with tests.

It is kinda of expected behavior since in server fetch, redirects are followed but `fetch` is stateless (unlike browsers, it does not store cookies before redirection for example)

As a result, you are right we need `fetchOptions: { redirect: "manual" }` in `proxyReqeust` to manually access the first encounter's response (which includes cookie and manual redirect).

If you feel we might improve DX somehow please feel free to open a new discussion issue 👍🏼

PR #582, opened 2023-11-22
types: add generics to `isError` and update `DataT` default generic param

DamianGlowala:
### ❓ Type of change

- [x] 🧹 Chore (updates to the build process or auxiliary tools and libraries)

### 📚 Description

Makes `isError` generic, hence removing the need to explictly annotate the return type of `createError`.

@pi0, shall we consider amending the default generic param for `DataT` from `any` to `unknown`, as per [this discussion](https://github.com/nuxt/nuxt/pull/24396#discussion_r1401272551)?

### 📝 Checklist

### DIFF
--- a/src/error.ts
+++ b/src/error.ts
@@ -19,7 +19,7 @@ import { hasProp } from "./utils/internal/object";
  *                         This can be used to pass additional information about the error.
  * @property {boolean} internal - Setting this property to `true` will mark the error as an internal error.
  */
-export class H3Error<DataT = any> extends Error {
+export class H3Error<DataT = unknown> extends Error {
   static __h3_error__ = true;
   statusCode = 500;
   fatal = false;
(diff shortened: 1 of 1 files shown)


pi0:
> @pi0, shall we consider amending the default generic param for DataT from any to unknown, as per https://github.com/nuxt/nuxt/pull/24396#discussion_r1401272551?

Would it be necessary for generic support? If yes, sure LGTM as breaking change surface for error data type seems low. (in general, i prefer any => unknown migrations in major versions but this seems small)

src/error.ts
```diff
@@ -68,12 +68,12 @@ export function createError<DataT = any>(
   input:
     | string
     | (Partial<H3Error<DataT>> & { status?: number; statusText?: string }),
-): H3Error {
+) {
```
danielroe, line 71: Do we need to add an explicit return type here to ensure the generic propagates?

danielroe (commented):
This looks good to me. I'd recommend migrating to `unknown` here as @pi0 is happy to do it in a non major version. ❤️

src/error.ts
DamianGlowala: With the addition of generics to `isError`, this is now inferred correctly as `H3Error<DataT>`.

pi0 (approved):

pi0:
Probably doing release by next week but you can try using nightly channels 👍🏼 (And thanks!)

PR #594, opened 2023-12-07
feat(validate): provide validate error in `data`

Atinux:
### ❓ Type of change

### 📚 Description

We were not forwarding the error from zod to the `data` property, making it impossible to get back the errors/issues from it.

Before:
CleanShot 2023-12-07 at 11 50 19

After:
CleanShot 2023-12-07 at 11 50 41

### DIFF
--- a/src/utils/internal/validate.ts
+++ b/src/utils/internal/validate.ts
@@ -40,8 +40,6 @@ function createValidationError(validateError?: any) {
     status: 400,
     statusMessage: "Validation Error",
     message: validateError.message || "Validation Error",
-    data: {
-      ...validateError,
-    },
+    data: validateError,
   });
 }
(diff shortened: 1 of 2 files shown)


Atinux:
I guess we could also improve the error message but might be a breaking change so I kept it this way.

src/utils/internal/validate.ts
```diff
@@ -38,7 +38,10 @@ export async function validateData<T>(
 function createValidationError(validateError?: any) {
   throw createError({
     status: 400,
-    message: validateError.message || "Validation Failed",
-    ...validateError,
+    statusMessage: "Bad Request",
+    message: validateError.message || "Bad Request",
+    data: {
+      ...validateError,
```
pi0, line 44: We might directly pass error to `data` or `cause` without destructure or use a known specific key as it might be unsafe on artibrary error objects

Atinux: As you prefer, as long as we have the data :)

pi0 (approved):
💯

Issue #375, opened 2023-04-12
An error occurred when using the proxyRequest method to proxy a DELETE request.


WongSaang:
### Environment

Nuxt

### Reproduction

This is the code for the server middleware in my Nuxt project:

```
export default defineEventHandler((event) => {
    if (event.node.req.url.startsWith('/api/')) {
        return proxyRequest(
            event,
            (process.env.SERVER_DOMAIN || 'http://localhost:8000') + event.node.req.url
        )
    }
})
```

### Describe the bug

```
[nuxt] [request error] [unhandled] [500] fetch failed                                                                                         16:59:26
  at Object.fetch (node:internal/deps/undici/undici:14062:11)
  at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
  at async sendProxy (./node_modules/h3/dist/index.mjs:506:20)
  at async Object.handler (./node_modules/h3/dist/index.mjs:1247:19)
  at async Server.toNodeHandle (./node_modules/h3/dist/index.mjs:1322:7)
```

lenqwang:
I had the same problem

tjhiggins:
Found a fix from https://github.com/unjs/nitro/issues/1137

`export NODE_OPTIONS='--no-experimental-fetch'`

It seems the experimental fetch doesn't like `content-length: 0` for deletes.

migueabellan:
My workaround for this issue was to add an empty body in the DELETE method.

```
fetch(url, { method: 'DELETE', body: [] }),
```

pi0:
(note for myself) some nice investigation by @jasenmichael  https://github.com/unjs/h3/pull/605#issuecomment-1910787078 most intrestingly that it seems to be reproducable only with Nuxt3 (maybe because of browser request?)

eljass:
We had this problem and the problem to proxy `OPTIONS` requests. It seems that from Node v20.11.0 onwards, this issue has been fixed within the `fetch` as upgrading from `v18.7.1` to `v20.11.0` removes the error.

The problem here was originated from undici's fetch which resulted the error `UND_ERR_REQ_CONTENT_LENGTH_MISMATCH`. For some reason, the error the h3 server prints does not include the cause. To print that we needed to modify these lines to catch original error: https://github.com/unjs/h3/blob/c8aa150b59d12f1a25ef154c00eeae132da59a32/src/utils/proxy.ts#L84C3-L88C6

```ts
const response = await _getFetch(opts.fetch)(target, {
[...]

PR #605, opened 2024-01-04
fix(proxy): workaround undici issue with `DELETE` and content-length of 0

jasenmichael:
###  Linked issue

- #375
- https://github.com/nuxt/nuxt/issues/19325#issuecomment-1511440436_

### ❓ Type of change

###  Description

This pull request addresses the issue of proxy DELETE requests failing when no body or "content-length" header is present.

example error (this PR solves) when making a proxy DELETE request with nuxt3/nitro using h3:
``` typescript (nuxt.config)
export default defineNuxtConfig({
  routeRules: {
    '/api/**': { proxy: `${apiUrl}/**`},
  }
}
```

```bash
[nuxt] [request error] [unhandled] [500] fetch failed
  at Object.fetch (node:internal/deps/undici/undici:11372:11)  
  at process.processTicksAndRejections (node:internal/process/task_queues:95:5)  
  at async sendProxy (./node_modules/h3/dist/index.mjs:1084:20)  
  at async Object.handler (./node_modules/h3/dist/index.mjs:1680:19)  
  at async Server.toNodeHandle (./node_modules/h3/dist/index.mjs:1890:7)
```

**Fixes:**

- Ensures an empty body and "content-length" header for proxy DELETE requests.
- Adds tests to verify the fix.
- DELETE errors related to `async sendProxy (./node_modules/h3/dist/index.mjs)`

###  Checklist

- [ ] I have updated the documentation accordingly. (Check this box if you've also made documentation changes)

### DIFF
--- a/src/utils/proxy.ts
+++ b/src/utils/proxy.ts
@@ -57,6 +57,12 @@ export async function proxyRequest(
     opts.headers,
   );
 
+  // DELETE method must always have an empty object in the body, and a "content-length" header.
+  if (method === "DELETE" && body === undefined) {
+    body = Buffer.from("{}");
+    (fetchHeaders as { [key: string]: any })["content-length"] = 2;
+  }
+
   return sendProxy(event, target, {
     ...opts,
     fetchOptions: {
--- a/test/proxy.test.ts
+++ b/test/proxy.test.ts
@@ -240,6 +240,43 @@ describe("", () => {
 
       expect(resText).toEqual(message);
     });
+
+    it("can proxy DELETE request", async () => {
+      app.use(
+        "/delete",
+        eventHandler((event) => {
+          event.node.res.statusCode = 204;
+          return null;
+        }),
+      );
+
+      app.use(
+        "/",
+        eventHandler((event) => {
+          return proxyRequest(event, url + "/delete", { fetch });
+        }),
+      );
+
+      const result = await fetch(url + "/", {
+        method: "DELETE",
+      });
+      const { body, status, statusText } = result;
+      expect(status).toEqual(204);
+      expect(statusText).toEqual("No Content");
+      expect(body).toEqual(null);
+
+      const result2 = await fetch(url + "/", {
+        method: "DELETE",
+        body: Buffer.from("{}"),
+      });
+      expect(result2.status).toEqual(204);
+
+      const result3 = await fetch(url + "/", {
+        method: "DELETE",
+        body: null,
+      });
+      expect(result3.status).toEqual(204);
+    });
   });
 
   describe("multipleCookies", () => {


test/proxy.test.ts
```diff
@@ -240,6 +240,43 @@ describe("", () => {
 
       expect(resText).toEqual(message);
     });
+
+    it("can proxy DELETE request", async () => {
```
pi0, line 244: Tests seem to be pass even by commenting out the changes in utils/proxy we might add something that fails without it.

src/utils/proxy.ts
```diff
@@ -57,6 +57,12 @@ export async function proxyRequest(
     opts.headers,
   );
 
+  // DELETE method must always have an empty object in the body, and a "content-length" header.
+  if (method === "DELETE" && body === undefined) {
+    body = Buffer.from("{}");
```
pi0, line 62: By HTTP spec, it is not disallowed. However conventionally, HTTP servers fail when `DELETE` omits the body. I'm not sure using a JavaScript HTTP object would be best solution.

pi0:
Thanks for working on this PR dear @jasenmichael.

I'm not sure if you have already followed but this is relevant issue in undici that causes error (https://github.com/nodejs/undici/issues/2046)

With https://github.com/nodejs/undici/pull/2305, Node.js should now allow content-length of `null` instead of `0` to explicitly say there is no content which we could try to use instead.

guibulator:
So https://github.com/nodejs/undici/pull/2305  has been merged since https://github.com/nodejs/undici/releases/tag/v5.27.1.
Nitropack is already using undici 5.28.1. https://github.com/unjs/nitro/blob/b53e00191d0752c7ae7463a907f0858c8bb8182e/package.json#L59
I tried with only setting content-length to 0 or null with no body and it doesn't work. I can only make it work with the fix proposed in this MR.

pi0:
> Nitropack is already using undici 5.28.1

it doesn't depend on any specific version of undici. it is only a dev dependency of the repository itself for another reason (miniflare used in test)

pi0:
Thanks for your PR dear @jasenmichael. earlier in comments, new tests pass with/without the changes (https://github.com/unjs/h3/pull/605#discussion_r1442217945) also we cannot do it automatically to provide a probably unwanted body with data of `"{}"` to the DELETE requests only to bypass a bug in Node.js/UndiCi.

This is an important issue and I hope to find a fix but let's track in #375

pi0:
Thanks for kind words and more details. Well I have to say it is now even more interesting! I will try to make same reproduction as you have with Nuxt. If you have better one, please feel free to share it in #375 that would make help a lot in order to followup.

trc-mathieu:
Tested with nuxt 3.10 and node 20.11.0 and delete now works fine.

PR #606, opened 2024-01-05
fix(cookie): fix setCookie to respect serializeOptions

### LINKED ISSUE #598: Overwriting headers Set-Cookie
### Environment

node: 20.8.1
h3: 1.9.0

### Reproduction

await deleteCookie(event, 'test', { domain: example.test })
await deleteCookie(event, 'test', { domain: .example.test })

### Describe the bug

It is impossible to delete cookies using the "deleteCookie" method, cookies of the same type but with different parameters

https://github.com/unjs/h3/blob/main/src/utils/cookie.ts#L54-L56

### Additional context

_No response_

### Logs

_No response_

unite4:
### 🔗 Linked issue

#598 

### ❓ Type of change

### 📚 Description

Resolves #598 

### 📝 Checklist

### DIFF
--- a/src/utils/cookie.ts
+++ b/src/utils/cookie.ts
@@ -52,7 +52,15 @@ export function setCookie(
     setCookies = [setCookies as any];
   }
   setCookies = setCookies.filter((cookieValue: string) => {
-    return cookieValue && !cookieValue.startsWith(name + "=");
+    if (!cookieValue) {
+      return false;
+    }
+    const curCookie = parse(cookieValue);
+    const newCookie = parse(cookieStr);
+    return (
+      Object.entries(curCookie).toString() !==
+      Object.entries(newCookie).toString()
+    );
   });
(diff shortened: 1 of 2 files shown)


pi0 (approved):
Thanks!

tobiasdiez:
@unite4 @pi0 I'm a bit confused by the semantics of `setCookie` now. If I call it twice with the same name but different values, I get now two cookies instead of only one with the latest data. This feels counterintuitive. I encountered this in the context of the built-in session functionality. If you call `updateSession` you get two session cookies: an empty one and one with the updated data.

frenzzy:
Currently, it is not possible to set custom session data for the first request:
```js
const session = await useSession(...) // Adds set-cookie with empty data
session.update(...)                   // Adds set-cookie with custom data
```
The response returns two identical set-cookie headers, but it seems the browser uses only the first one.

Related: https://github.com/unjs/h3/issues/705

PR #607, opened 2024-01-06
fix(setResponseHeaders): fix types to allow partial header names

dargmuesli:
### 🔗 Linked issue

https://github.com/unjs/h3/pull/542
resolves https://github.com/unjs/h3/issues/614

### ❓ Type of change

### 📚 Description

Currently, the record requires all keys to be set.
See: https://github.com/maevsi/maevsi/actions/runs/7428213670/job/20215223408

Adding `Partial` allows `value` to be `undefined` though. I guess the idea for undefined values would be to remove the corresponding header, but that change can be reverted of course!

### 📝 Checklist

### DIFF
--- a/src/utils/response.ts
+++ b/src/utils/response.ts
@@ -107,10 +107,12 @@ export function getResponseHeader(
 
 export function setResponseHeaders(
   event: H3Event,
-  headers: Record<HTTPHeaderName, Parameters<OutgoingMessage["setHeader"]>[1]>,
+  headers: Partial<
+    Record<HTTPHeaderName, Parameters<OutgoingMessage["setHeader"]>[1]>
+  >,
 ): void {
   for (const [name, value] of Object.entries(headers)) {
-    event.node.res.setHeader(name, value);
+    event.node.res.setHeader(name, value!);
   }
 }
 


src/utils/response.ts
```diff
@@ -107,10 +107,16 @@ export function getResponseHeader(
...
 export function setResponseHeaders(
   event: H3Event,
-  headers: Record<HTTPHeaderName, Parameters<OutgoingMessage["setHeader"]>[1]>,
+  headers: Partial<
+    Record<HTTPHeaderName, Parameters<OutgoingMessage["setHeader"]>[1]>
+  >,
 ): void {
   for (const [name, value] of Object.entries(headers)) {
-    event.node.res.setHeader(name, value);
+    if (value === undefined) {
```
danielroe, line 115: I do not think we need to add logic to cover this. I think we can just assert `value!`.

dargmuesli: changed :raised_hands:

pi0: for reference Node.js will throw `ERR_HTTP_INVALID_HEADER_VALUE` for undefined https://github.com/nodejs/node/blob/main/lib/_http_outgoing.js#L677

pi0 (approved):

dargmuesli:
@pi0 can we have a patch release for this? :see_no_evil:

devunt:
This issue blocks upgrading `h3` to `1.10.0`.

PR #612, opened 2024-01-08
fix(getSession): use semaphore lock for unseal operation

### LINKED ISSUE #610: Concurrent calls to `getSession` leads to only the first one getting the session
### Environment

stackblitz
h3: 1.10.0
node: 18.18

### Reproduction

Barebone reproduction using `H3Event` and `useSession` only.
[Open in StackBlitz](https://stackblitz.com/edit/stackblitz-starters-jyfqat?file=index.js)

Solid-Start repro by the issue author here https://github.com/solidjs/solid-start/issues/1192
[Open in StackBlitz](https://stackblitz.com/edit/metruzanca-solidstart-getsession-bug-repro-xd4fmg?file=README.md)

Steps (for Solid-Start Repro):

1. Press Login button
2. Refresh page - You are logged out
3. Press Index link - You are logged in

### Describe the bug

When calling `getSession` multiple times concurrently on a single page request the first call will initialize the `context.sessions[sessionName]` to an empty session. All other calls will see that there is a session on the context and return it.

### Additional context

I guess `unsealSession` gives the other calls a chance to advance to the context check and return early.
If appropriate the session assignment to the context could be done after `unsealSession`. While testing it helped but this would lead to the session being unseald multiple times.
https://github.com/unjs/h3/blob/ae91fc8658315dca75da22b665eda4175eef7ea8/src/utils/session.ts#L108

### Logs

_No response_

JAD3N:
### 🔗 Linked issue

#610

### ❓ Type of change

### 📚 Description

I've added session locking that will asynchronously block consecutive `getSession` attempts when unsealing.

This should solve the problem where consecutive `getSession` calls break session data and prevent duplicate unseals (I tested the changes against the issue's [reproduction](https://stackblitz.com/edit/stackblitz-starters-jyfqat?file=index.js)).

Resolves #610

### 📝 Checklist


pi0 (approved):

pi0:
Thanks for the PR ❤️ I have added tests and few refactors to simplify using same `context.sessions` for locking.

As a future enhancement, we need to do concurrency/promise-reuse in function level which is currently not possible as `updateSession` has dependency on `getSession` promise but it will likely catch other edge case scenarios.

Issue #613, opened 2024-01-09
Support header filtering for `getRequestHeaders`


antoinerey:
### Environment

```
h3@1.10.0
node@18.18.2
```

### Reproduction

```ts
const headers = getRequestHeaders(event, ['host', 'cookie'])

const expected = { 
  host: '...', 
  cookie: '...', 
}

const actual = { 
  host: '...', 
  cookie: '...', 
  'accept-language': '...', 
  'user-agent': '...',
  // ... many more  
}
```

### Describe the bug

[When reading the README](https://github.com/unjs/h3/blob/main/README.md?plain=1#L225), the signature of `getRequestHeaders` looks like this: 

```ts
getRequestHeaders(event, headers)
```

I believe the second parameter was originally meant to retrieve only a subset of all request headers, similar to the `useRequestHeaders` of Nuxt. However, that documented signature does not match [the signature of the function implementation itself](https://github.com/unjs/h3/blob/ae91fc8658315dca75da22b665eda4175eef7ea8/src/utils/request.ts#L113).

I believe that the original intent was good, and `getRequestHeaders` should accept a second parameter to pick whatever headers we want to retrieve. In brief, the following code should work as expected:

```ts
const headers = getRequestHeaders(event, ['host', 'cookie'])
//    ^? { host: '...', cookie: '...' }
```

### Additional context

I'm open to creating a fix though.

pi0:
Please see reasonings in https://github.com/unjs/h3/pull/620#issuecomment-1962698915

You might want to apply filtering directly on top of this method.

antoinerey:
Got it. It makes sense. Thanks for the time! 🤗

PR #620, opened 2024-01-18
feat: `getRequestHeaders` filter headers

cpreston321:
### 🔗 Linked issue

Resolves : #613

### ❓ Type of change

- [X] ✨ New feature (a non-breaking change that adds functionality)

### 📚 Description

This solves for issues #613. This adds a second param as string[], that can filter what you want in the helper function.

### 📝 Checklist

- [X] I have linked an issue or discussion.
- [X] I have updated the documentation accordingly.


pi0:
Thanks for your work on this PR.

I have rebased it and thought little bit more. There are some reasons i think it might be to let it to the users to apply filtering:

- Normalization/lowercase overhead is not necessary but for a built-in function we might need it and it adds to perf
- Sometimes **excluding** is desired that could need another option
- Using `Set` could be faster while it also needs overhead **per-call**

PR #624, opened 2024-01-20
fix(sendStream): add support for file streaming

### LINKED ISSUE #623: Incompatible streams with raw values from unstorage when used in `sendStream`
### Environment

N/A

### Reproduction

https://github.com/GalacticHypernova/NuxtFailTest

Steps:
1. npm install
2. npm run dev
3. navigate to localhost:3000/images/Test2.jpg
4. profit

### Describe the bug

When sending streams of images using sendStream, it complains that the stream of the item is not compatible. That is beause, from what I understood, it takes the raw file content which is a buffer of bytes, and the sendStream function expects an object with `pipe` or `pipeTo` properties.

### Additional context

This doesn't occur when using `createReadStream` from `node:fs`

### Logs

_No response_

GalacticHypernova:
### 🔗 Linked issue

Closes #623 

### ❓ Type of change

### 📚 Description

This PR adds support for file streaming when used with unstorage by converting the buffer to a readable.

### 📝 Checklist

### DIFF
--- a/src/utils/response.ts
+++ b/src/utils/response.ts
@@ -227,7 +227,7 @@ export function sendStream(
   // Value is a buffer, like an image byte buffer
   if (Buffer.isBuffer(stream)) {
     // TODO
-    event.node.res.end(stream)
+    event.node.res.end(stream);
   }
   // Directly expose stream for worker environments (unjs/unenv)
   (event.node.res as unknown as { _data: BodyInit })._data = stream as BodyInit;
--- a/src/utils/response.ts
+++ b/src/utils/response.ts
@@ -76,7 +76,11 @@ export function getResponseStatusText(event: H3Event): string {
 }
 
 export function defaultContentType(event: H3Event, type?: string) {
-  if (type && !event.node.res.getHeader("content-type")) {
+  if (
+    type &&
+    event.node.res.statusCode !== 304 /* unjs/h3#603 */ &&
+    !event.node.res.getHeader("content-type")
+  ) {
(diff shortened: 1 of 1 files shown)


src/utils/response.ts
```diff
@@ -1,5 +1,5 @@
 import type { OutgoingMessage } from "node:http";
-import type { Readable } from "node:stream";
+import { Readable } from "node:stream";
```
pi0, line 2: We cannot depend on Node.js imports in h3.

src/utils/response.ts
```diff
@@ -220,7 +220,9 @@ export function sendStream(
   if (!stream || typeof stream !== "object") {
     throw new Error("[h3] Invalid stream provided.");
   }
-
+  if (Buffer.isBuffer(stream)) {
+    stream = Readable.from([stream]);
```
pi0, line 224: I am wondering since the data source is already in memory as Buffer, would there be any benefits of this over simply using `res.end(stream)` and let Node.js/OS TCP stack handle streaming?

GalacticHypernova: Is there a guarantee it would send it as a stream?

pi0: It is a TCP stream always. How to be chunked? it depends on many factors including stack and reverse proxy's behavior.

GalacticHypernova: @pi0 like this?

pi0:
Thanks for PR. I don't think we should wrap buffer into another stream but instead send it normally by `return` or `send()`

PR #625, opened 2024-01-21
feat: improve typed headers

GalacticHypernova:
### ❓ Type of change

### 📚 Description

This PR is something I thought about for a while. Since we have header intellisense, we might as well have value intellisense. This will ensure no wrong header values are set and would save time debugging header issues (a common example of this is the misconception that `image/jpg` is a valid content type, when in reality it doesn't have a universally acceptable type). Granted, this won't be a complete intellisense for all possible values as the value combinations can be endless, however it would give a good foundation.
EDIT: As of today, I have decided to expand this further and include status codes (maybe something else as well)

### 📝 Checklist


GalacticHypernova:
Marking this as draft as this is **very early** in development. I will gradually add everything I can, but it will take a good while.

GalacticHypernova:
As a side note, this should probably also be fixed in @types/node as some of the forced casts were made to comply with their generic typing

GalacticHypernova:
I believe this has covered everything in the scope of the PR. If I missed anything in any other function where it can be improved, please let me know and I'll gladly add it. There are a few things I would like to change, like the AnyType usage, but I do need a tiny bit of help in figuring out how to properly take care of multi-value parameter-included headers (like www-authenticate) but I think this is good as a baseline intellisense, I understand it can't be full-proof.

pi0 (approved):
Thanks!

(in summary made refactors to improve maintainability + relax types to only hint and still allow artibrary string/number as input)

GalacticHypernova:
Glad I can help!

PR #627, opened 2024-01-21
feat(headers): support array values for multi-field headers

GalacticHypernova:
### 🔗 Linked issue

Related to #625 
### ❓ Type of change

### 📚 Description

This PR adds a way to provide header values in arrays, which could be beneficial for headers that can accept multiple values (example: `Accept`). This PR will also make it much easier to implement #625 typing wise.

### 📝 Checklist

### DIFF
--- a/src/utils/response.ts
+++ b/src/utils/response.ts
@@ -112,7 +112,10 @@ export function setResponseHeaders(
   >,
 ): void {
   for (const [name, value] of Object.entries(headers)) {
-    event.node.res.setHeader(name, Array.isArray(value!) ? value!.join(", ") : value!);
+    event.node.res.setHeader(
+      name,
+      Array.isArray(value!) ? value!.join(", ") : value!,
+    );
   }
 }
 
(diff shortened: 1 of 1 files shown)


GalacticHypernova:
On second thought, this should probably be reworked to include semicolon separated headers, and perhaps create a helper function to determine what to separate by. I'm just worried about the performance side of it. Would love to hear your implementation suggestions!

GalacticHypernova:
Seeing as this is already supported by the underlying [http lib](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/http.d.ts#L589) I will keep this a draft until I find out it works fine. If it doesn't I will likely mark this as ready to review, is that fine?

pi0:
I agree that joining is a headache. Node.js already does process by joining with `; ` ([src](https://github.com/nodejs/node/blob/a072ec382f8b8a2e2e90e2520c1b94a668b4bc3b/lib/_http_outgoing.js#L624))

I think best is to let the users join multi values before calling methods if they need specific behavior.

PR #655, opened 2024-02-21
feat(response): add `sendIterable` util

passionate-bram:
### 🔗 Linked issue

https://github.com/unjs/h3/issues/581

### ❓ Type of change

### 📚 Description

Introduces a new utility function `sendIterable`, relevant types and an additional helper function are also exposed.

Newly exposed API:

```ts
export function sendIterable<Value = unknown, Return = unknown>(
  event: H3Event,
  iterable: IterationSource<Value, Return>,
  serializer: IteratorSerializer<Value | Return> = serializeIterableValue
) : Promise<void> {/* snip */}

// Supported iterable and iterator types of `sendIterable`
export type IterationSource<Val, Ret = Val> =
  | Iterable<Val>
  | AsyncIterable<Val>
  | Iterator<Val, Ret | undefined>
  | AsyncIterator<Val, Ret | undefined>
  | (() => (
    | Iterator<Val, Ret | undefined>
    | AsyncIterator<Val, Ret | undefined>
  ));

// The response stream (event.node.res) is in byte-mode, so you the types of data that can be written is restricted.
type SendableValue = string | Buffer | Uint8Array;
// The `undefined` return value is a short-cut to not writing anything to the response stream, for example if the data type cannot be serialized.
export type IteratorSerializer<Value> = (value: Value) => SendableValue | undefined;

// Default serializer that turns types that are not directly writable to the underlying network stream into ones that are.
export function serializeIterableValue(value: unknown): SendableValue | undefined {/* snip */}
```

The new feature allows users of h3 (and by extension Nitro and Nuxt) to write generator functions (`function*(){}`) that produce a response piece-by-piece. For more reason on justification see the linked issue.

Resolves https://github.com/unjs/h3/issues/581

### 📝 Checklist


pi0 (approved):
Thanks!

passionate-bram:
@pi0 A note worth mentioning in the docs is that the default buffering for server responses is per 16k bytes. So users may be surprised that the first `yield` writes out to the stream but subsequent ones do not, until they exceed the 16k buffer at which point all data up to that point is written out. 

This is a bit unfortunate for the primary use case to have live output for a long-running process, but by mentioning it users can either work around it (by somehow creating a different buffer for the node server) or accepting the limitation.

See also https://github.com/unjs/listhen/issues/161

PR #663, opened 2024-02-24
docs(examples/from-expressjs-to-h3): add node middleware usage

markthree:
### ❓ Type of change

### 📚 Description

I think we can add a guide on using middleware with h3 for users accustomed to Express, which would facilitate their migration. 

tip: This is also the confusion I have encountered in actual use; I believe they might face similar issues as well. 🤣

### 📝 Checklist


pi0 (approved):
Thanks

PR #742, opened 2024-04-30
feat(events): bodyValidator and queryValidator event arguments

DallasHoff:
### 🔗 Linked issue

unjs/nitro#2244

### ❓ Type of change

### 📚 Description

These changes introduce `bodyValidator` and `queryValidator` options to `defineEventHandler`. These can be passed validation functions, such as a Zod schema's `parse` method, to validate request bodies and query parameters and type them in the handler.

```ts
export default defineEventHandler({
  bodyValidator: z.object({
    id: z.string()
  }).parse,
  handler: async (event) => {
    const { id } = await readBody(event);
    // id is inferred as a string now
  }
});
```

This will also allow Nitro to type the `body` and `query` options of `$fetch` since the required types of the body and query can now be inferred from the event handler, something like this...

```ts
type Body = typeof import('../../routes/foo').default extends EventHandler<infer Request> 
  ? Request['body'] 
  : any
```

...opening the door to...

```ts
await $fetch('/foo', {
  method: 'POST',
  body: {}, // <-- body inferred as { id: string } from the Zod Schema, so this is a TS error
});
```

### 📝 Checklist

### DIFF
--- a/src/event/utils.ts
+++ b/src/event/utils.ts
@@ -1,4 +1,3 @@
-import { getQuery, readBody } from "src/utils";
 import type {
   EventHandler,
   LazyEventHandler,
@@ -9,8 +8,10 @@ import type {
   _ResponseMiddleware,
 } from "../types";
 import { hasProp } from "../utils/internal/object";
+import { validateData } from "../utils/internal/validate";
+import { readBody } from "../utils/body";
+import { getQuery } from "../utils/request";
 import type { H3Event } from "./event";
-import { validateData } from "src/utils/internal/validate";
 
 type _EventHandlerHooks = {
   onRequest?: _RequestMiddleware[];
--- a/src/event/utils.ts
+++ b/src/event/utils.ts
@@ -1,3 +1,4 @@
+import { getQuery, readBody } from "src/utils";
 import type {
   EventHandler,
   LazyEventHandler,
@@ -9,6 +10,7 @@ import type {
 } from "../types";
 import { hasProp } from "../utils/internal/object";
 import type { H3Event } from "./event";
+import { validateData } from "src/utils/internal/validate";
 
 type _EventHandlerHooks = {
   onRequest?: _RequestMiddleware[];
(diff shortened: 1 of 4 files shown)


pi0:
Thanks for PR. Mind you we had been working on a more generic proposal `validate` key on eventhandler (https://github.com/unjs/h3/pull/496). I'm more leaned towards supporting single validate rather than type specific separation (for body, query, params, etc)

DallasHoff:
I feel this solution is elegant for a few reasons:

- No need for mutating the event or having type assertions anywhere (`return event as EventWithValidatedBody`)
- Developers can easily type the body/query however they want whether its with a schema library or other function or if they have a type already defined somewhere else, they can just import it and pass it to `defineEventHandler` as a type argument.

```ts
type FooBody = {
  id: string;
}

export default defineEventHandler<FooBody>()
```

- Defining the type in either of these ways automatically makes `readBody` and `getQuery` return the correct type.
- On the reverse side, it's simple to infer the required body/query type from the handler if the developer needs to. h3 could also expose utility types to simplify it even further (e.g. `HanderBody<T extends EventHandler>`)

```ts
import fooHandler from 'api/foo';

type FooBody = fooHandler extends EventHandler<EventHandlerRequest<infer Body>> ? Body : never;
```

- No need to introduce a new function to h3
- No real performance impact for projects that do not want to use this feature
- In Nitro, it is now possible to make sure the route params, method, body, and query params in a fetch all match up. Entering one will narrow the others and give the corresponding auto-complete.
- Keep in mind that most handlers will not have both a required body _and_ query params, so having a generic validate function with extra steps to specify whether it validates the body or the query may not be as simple as having a dedicated option for each case.

pi0:
I appreciate the PR and sorry for the late update. We have removed object syntax for v2 in favor of new middleware format (check #1065, #1075) to intercept request and responses. Would be nice to reimplement this using middleware.

Issue #722, opened 2024-03-27
Non-graceful error when proxy is offline


manniL:
### Environment

- Operating System: `Darwin`
- Node Version:     `v21.6.2`
- Nuxt Version:     `-`
- CLI Version:      `3.11.1`
- Nitro Version:    `-`
- Package Manager:  `pnpm@8.5.0`
- Builder:          `-`
- User Config:      `-`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

1. Use the h3 playground example
2. Create a handler with the following code (StackBlitz won't work as it is about proxy):

```ts
eventHandler((event) => {
  let url = "http://some-non-existent.url.domain"
  return proxyRequest(event, url)
})
```

3. Visit the route

### Describe the bug

H3 responds with a 500 because the fetch request fails, either because the address can't be found or if the server is offline.

A more graceful way of handling would be nice 🤔

PR #746, opened 2024-05-05
fix(proxy): better error when upstream proxy fails

manniL:
### 🔗 Linked issue

#722

### ❓ Type of change

### 📚 Description

This PR should ensure that `proxyRequest`s fail gracefully instead of leading to an unhandled node error crashing the server. 

### 📝 Checklist


src/utils/proxy.ts
```diff
@@ -61,16 +61,22 @@ export async function proxyRequest(
...
+        method,
+        body,
+        duplex,
+        ...opts.fetchOptions,
+        headers: fetchHeaders,
+      },
+    });
+  } catch {
+    event.node.res.statusCode = 502;
+    event.node.res.statusMessage = "Bad Gateway";
```
pi0, line 77: I think only we can assume only proxy had been failed and perhaps we can double check if there is not already set statusCode and message on error (that currently is being ingored)

Additionally, this perhaps can be implemented directly in `sendProxy` util for consistent behavior

manniL: Moving back to draft until implemented! Good idea 👍🏻

manniL:
Updated to `sendProxy`. 

> I think only we can assume only proxy had been failed and perhaps we can double check if there is not already set statusCode and message on error (that currently is being ingored)

You mean an existing one in `event.node.statusCode` / `statusMessage` or in the error directly? If the latter, which values should I check?

pi0 (approved):
Thanks ❤️ (updated impl to still throw error to allow hooking into it but with `createError`)

PR #756, opened 2024-05-18
fix(node): make sure `onBeforeResponse` and `onAfterResponse` are called with error code

### LINKED ISSUE #754: Cannot track status codes
### Environment

- Node Version:     v20.12.2
- H3 Version:     1.11.1

### Reproduction

See https://stackblitz.com/edit/github-d9z5tl?file=app.ts

### Describe the bug

Forwarded from https://github.com/nuxt/nuxt/issues/27248

When I return an error in `eventHandler`'s using `createError`, global hooks `onBeforeResponse` and `onAfterResponse` behave unexpectedly:

- `onBeforeResponse` always returns `res.statusCode` equal to 200
- `onAfterResponse` does not work in case of an error in the handler

This does not track the actual status codes that users may encounter. For example, in my app I cannot collect statistics in Prometheus metrics.

```js
export const app = createApp({
  onBeforeResponse: (event) => {
    // will always be 200
    console.log('[beforeResponse]', event.node.res.statusCode);
  },
  onAfterResponse: (event) => {
    // not called
    console.log('[afterResponse]', event.node.res.statusCode);
  },
});

app.use(router);

router.get(
  '/',
  eventHandler((event) => {
    return createError({ statusCode: 404 });
  })
);
```

### Additional context

_No response_

### Logs

_No response_

imslepov:
### 🔗 Linked issue

Closes #754 

### ❓ Type of change

### 📝 Checklist


pi0 (approved):

pi0:
Thanks for PR and nice tests @imslepov ❤️ 

I have updated the implementation to minimize the required changes. Expected tests pass still but we call hooks only before sending error. It should be cleaner but requires more additional core refactors later.

PR #789, opened 2024-06-25
fix(node): hook calling order

imslepov:
Relates to #756

There is a bug where if the `onError` hook handles the event (by setting `event.handled = true`), the `onBeforeRequest` and `onAfterRequest` hooks are not called 😅

In this PR, I made the changes:
- Call `onBeforeResponse` hook before `onError` because `onError` may handle an error and the server may return a response.
- The `event.handled` flag does not affect the calling hooks `onBeforeResponse` and `onAfterRequest`.

### DIFF
--- a/src/adapters/node/utils.ts
+++ b/src/adapters/node/utils.ts
@@ -36,6 +36,10 @@ export function toNodeHandler(app: App): NodeHandler {
       event[_kRaw].responseCode = error.statusCode;
       event[_kRaw].responseMessage = error.statusMessage;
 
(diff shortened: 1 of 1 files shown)


pi0:
Thanks for PR dear @imslepov ❤️

#792 changes the whole internal structure to make sure hooks are reliable.

in the meantime, i guess the issue exists for v1. You can rebase changes to it (probably new PR would be easier)

imslepov:
@pi0 Thanks! Amazing work on version 2.

Created PR for v1 fix - https://github.com/unjs/h3/pull/793

PR #793, opened 2024-06-26
fix(node): v1 - hook calling order

imslepov:
Relates to https://github.com/unjs/h3/pull/756, https://github.com/unjs/h3/pull/789


pi0:
Hi again. 

v2 supports fine-grained control to handle errors (using global `onError` or new `onError()` middleware)

Your change seems reasonable, however, it makes a  behavior change which might cause unwanted regressions for v1 users (and we are going soon to make it maintenance-only), so better not to risk IMO.

Cheers ❤️

PR #795, opened 2024-06-27
feat(createError): inherit fields from `cause` as fallback

### LINKED ISSUE #691: Allow for parsing of error status codes from non-http response errors.
### Describe the feature

Hey there 👋 

I'm an [Apollo Client](https://www.apollographql.com/docs/react/) maintainer. I'd like to start a discussion in regards to error handling when using Apollo Client with Nuxt/h3 and see if we can collaborate on a solution.

We recently had [this issue](https://github.com/apollographql/apollo-client/issues/11634) opened in our repo noting that network errors thrown by Apollo Client aren't forwarding the status code to h3's underlying `H3Error` due to how the status code is parsed in h3's [error creation](https://github.com/unjs/h3/blob/main/src/error.ts) functions. It appears `createError` expects a [`statusCode` property](https://github.com/unjs/h3/blob/27f6fad97dd0292d82fe30d9a7f4fca7511763ea/src/error.ts#L102) in order to set this on the `H3Error` instance. Unfortunately this doesn't integrate well with Apollo Client because Apollo Client wraps all errors in its own `ApolloError` instance. The status code is lost since the status code is instead found in `error.networkError.statusCode`.

I'd love to contribute a way to allow for these libraries to work better together. Unfortunately we are stuck with this error structure for the foreseeable future and we'd like to avoid adding a `statusCode` property to `ApolloError` itself, especially since Apollo Client's core aims to be framework agnostic. 

One potential solution is that we could add a [`cause`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause) from `ApolloError` to the original error. With this perhaps `createError` could try to parse `status`/`statusCode` from `error.cause` if its available?

I'd love to get your thoughts on this! At the very least, perhaps a recommendation I could take back to the original issue would be useful. I'll be honest, I'm pretty unfamiliar with Nuxt/h3, so any help would be appreciated. Thanks!

### Additional information

jerelmiller:
Resolves #691

Updates the logic for detecting the `statusCode` from an error by checking if it can be found on `error.cause`. This helps better detect a status code from an error created by a library that creates their own wrapped errors (such as [Apollo Client](https://www.apollographql.com/docs/react/)'s [`ApolloError`](https://github.com/apollographql/apollo-client/blob/7111de266e4ae5b21cf8cb59531db6f8bb9b443f/src/errors/index.ts#L74-L114) class).

See https://github.com/apollographql/apollo-client/issues/11634 for more context on this change. We've opened https://github.com/apollographql/apollo-client/pull/11902 on our end to make this integration possible.


src/error.ts
```diff
@@ -121,6 +121,13 @@ export function createError<DataT = unknown>(
     err.statusCode = sanitizeStatusCode(input.statusCode, err.statusCode);
   } else if (input.status) {
     err.statusCode = sanitizeStatusCode(input.status, err.statusCode);
+  } else if (
+    typeof input.cause === "object" &&
```
jerelmiller, line 125: Since `cause` can have all sorts of shapes, I find it best to do some type checking here before we try and parse from it. If you prefer a different style on this kind of check (i.e. an external helper function), let me know and I'd be happy to refactor.

src/error.ts
```diff
@@ -121,6 +121,13 @@ export function createError<DataT = unknown>(
...
+    input.cause &&
+    "statusCode" in input.cause &&
+    typeof input.cause.statusCode === "number"
+  ) {
+    err.statusCode = sanitizeStatusCode(input.cause.statusCode, err.statusCode);
```
jerelmiller, line 130: Should I also try and detect `status` from `error.cause.status` since this is also checked on the top-level properties, or should I just check `statusCode` here?

src/error.ts
pi0: Inline check is fine only `cause` can be `null` too (also saw it in https://github.com/apollographql/apollo-client/pull/11902). `typeof null === "object"` and well next line fails so we can simplify it to truehy checking like `input.cause?.statusCode`.

pi0:
Thanks for PR dear @jerelmiller i will push some refactors to generalize it. 

(also note, current `main` branch is targeting upcoming h3 v2, if more ecosystem backward compatibility adoption is needed, we might want to backport it to v1 as well)

pi0 (approved):

pi0:
Thanks again for PR and sorry for late review. I have pushed few refactors, simplifying implementation and also we use `cause` to inherit `statusCode`, `statusMessage` and `fatal`/`unhandled` fields (all which could be masked with wrapper)

jerelmiller:
Wonderful. Thanks so much for the review and getting this in there!

PR #814, opened 2024-07-08
chore: add benchmark for web api

pi0:
Add baseline/h3 benchmarks for web API (`Request => Promise<Response>`) comparing h3's performance with DIY approaches and the goal of getting h3 perf close to native perf (while with a framework/router on top it is impossible, we can do best to get closer!)

(bundle test added in https://github.com/unjs/h3/commit/68735e56d32eb3f1d784dc90b20b74d994b7357d)

Results today: `pnpm run bench`

image

### DIFF
--- a/test/bench.test.ts
+++ b/test/bench.test.ts
@@ -1,0 +1,114 @@
+import { describe, it, expect } from "vitest";
+import { createRouter, createApp, readJSONBody, toWebHandler } from "../src";
+import { getQuery } from "../src/utils/request";
+import { setResponseHeader } from "../src/utils/response";
+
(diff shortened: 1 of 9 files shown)


test/bench.test.ts
```diff
@@ -0,0 +1,114 @@
+import { describe, it, expect } from "vitest";
+import { createRouter, createApp, readJSONBody, toWebHandler } from "../src";
+import { getQuery } from "../src/utils/request";
+import { setResponseHeader } from "../src/utils/response";
+
+// https://github.com/pi0/web-framework-benchmarks
```
pi0, line 6: private for now will open once ready!

PR #821, opened 2024-07-09
chore: remove unnecessary `await`

iiio2:
Here `val` is already awaited. In the next line in `_body` for conditional checking, we have given `await` again for val. Also the code editor is saying await has no effect here like that. That's why I thought may be we may remove it.


pi0:
changed after #822

iiio2:
Thanks @pi0

PR #824, opened 2024-07-11
docs(ServeStaticAssets): fix title format

iiio2:
Title is capitalized like other titles.


pi0:
Thanks for PR. we are moving to this new format `This is a title` instead of `This Is a Title`

PR #834, opened 2024-07-20
docs: add detailed error response example

lborgav:
Some properties are missing in the return json from docs.

The correct response for the first example of [Error Handling](https://h3.unjs.io/guide/event-handler#error-handling) is:

```json
{
  "statusCode": 400,
  "statusMessage": "Bad Request",
  "stack": [],
  "data": {
    "field": "email"
  }
}
```

Reference: Response in Hoppscotch:
image


pi0 (approved):
Thanks

PR #868, opened 2024-08-20
fix(serveStatic): set `allow` header when responding with a `405`

aaharu:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Allow

> This header must be sent if the server responds with a 405 Method Not Allowed status code to indicate which request methods can be used.

### DIFF
--- a/src/utils/static.ts
+++ b/src/utils/static.ts
@@ -14,6 +14,7 @@ export async function serveStatic(
   options: ServeStaticOptions,
 ): Promise<false | undefined | null | BodyInit> {
   if (event.request.method !== "GET" && event.request.method !== "HEAD") {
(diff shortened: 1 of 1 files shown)


src/utils/static.ts
```diff
@@ -14,6 +14,7 @@ export async function serveStatic(
   options: ServeStaticOptions,
 ): Promise<false | undefined | null | BodyInit> {
   if (event.request.method !== "GET" && event.request.method !== "HEAD") {
+    event.response.headers.set("allow", "GET, HEAD");
     if (!options.fallthrough) {
       throw createError({
```
pi0, line 19: I think we should move it inside the `if` condition as we only add header alongside the error.

PR #856, opened 2024-08-08
fix(cors): append `access-control-max-age` header to preflight response

aaharu:
resolves #854

PR #870, opened 2024-08-26
fix(cors)!: doesn't return `access-control-allow-origin` header when origin is disallowed

### LINKED ISSUE #355: CORS issue in creating origin headers
### Environment

Node: v18.15.0
npm: v9.5.0
Nuxt: v3.2.3 + nuxt-security v0.11.0

### Reproduction

https://github.com/maxdzin/nuxt-cors-issue

### Describe the bug

When working on API, I used [nuxt-security](https://github.com/Baroshem/nuxt-security) and I noticed that the `Access-Control-Allow-Origin` header is not set correctly.

So, there could be a bug related to the just-merged `@nozomuikuta/h3-cors` into H3.

There is a discussion about the CORS problems in this [nuxt-security issue](https://github.com/Baroshem/nuxt-security/issues/95) and it is related to `h3-cors` things.

In short, there's a response 'Access-Control-Allow-Origin' header is not set correctly, while the options contain specific origin values.

So, there are these problems:
1. If the `H3CorsOptions.origin` value is set as an array of strings or an array of RegExp, the `access-control-allow-origin` header is not set in this case. In order to set that response header correctly, I believe that [`createOriginHeaders`](https://github.com/unjs/h3/blob/main/src/utils/cors/utils.ts#L73) method must return with `access-control-allow-origin` value set in the last return block here:
```ts
export function createOriginHeaders(
  event: H3Event,
  options: H3CorsOptions
): H3AccessControlAllowOriginHeader {
  const { origin: originOption } = options;
  const origin = getRequestHeader(event, "origin");

  if (!origin || !originOption || originOption === "*") {
    return { "access-control-allow-origin": "*" };
  }

  if (typeof originOption === "string") {
    return { "access-control-allow-origin": originOption, vary: "origin" };
  }

  return isCorsOriginAllowed(origin, options)
    ? { "access-control-allow-origin": origin, vary: "origin" }
    : {};
}
```
So, if the `isCorsOriginAllowed` is false, it returns an empty object. But, it must be set by `originOption` value instead.

2. And there's confusion with [H3CorsOptions type](https://github.com/unjs/h3/blob/main/src/utils/cors/types.ts):
```ts
export interface H3CorsOptions {
  origin?: "*" | "null" | (string | RegExp)[] | ((origin: string) => boolean);
  ...
```
The string type is missing there.

### Additional context

_No response_

### Logs

_No response_

aaharu:
resolves #355

### DIFF
--- a/docs/2.utils/98.advanced.md
+++ b/docs/2.utils/98.advanced.md
@@ -191,7 +191,7 @@ router.use("/", async (event) => {
 
 ### `isCorsOriginAllowed(origin, options)`
 
-Check if the incoming request is a CORS request.
+Check if the origin is allowed.
 
 ### `isPreflightRequest(event)`
 
--- a/src/utils/cors.ts
+++ b/src/utils/cors.ts
@@ -13,12 +13,64 @@ import {
 export { isCorsOriginAllowed } from "./internal/cors.ts";
 
 export interface CorsOptions {
+  /**
+   * This determines the value of the "access-control-allow-origin" response header.
+   * If "*", it can be used to allow all origins.
+   * If an array of strings or regular expressions, it can be used with origin matching.
+   * If a custom function, it's used to validate the origin. It takes the origin as an argument and returns `true` if allowed.
+   *
+   * @see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin
+   * @default "*"
+   */
   origin?: "*" | "null" | (string | RegExp)[] | ((origin: string) => boolean);
+
+  /**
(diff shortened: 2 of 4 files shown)


pi0:
Can you please add a regression test?

aaharu:
It's going to be a conflict, so I'll work on it after #856.

aaharu:
IMO, `Access-Control-Allow-Origin: null` doesn't need to be supported.
Because it should not be used.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin#null

PR #872, opened 2024-08-31
docs: remove link to missing playground

samydoesit:
Removed the link of the missing playground.


pi0:
playground added back!

PR #883, opened 2024-09-21
chore: enable shell emulator

iiio2:


pi0:
i would suggest you add it to global npmrc for windows support.

iiio2:
@pi0 , ok. Thanks

PR #888, opened 2024-09-26
fix(readRawBody): handle `URLSearchParams`

### LINKED ISSUE #887: Using `readBody` with Nuxt internal routing fails to parse URLSearchParams
### Environment

- Operating System: `Linux`
- Node Version:     `v22.5.1`
- Nuxt Version:     `3.13.2`
- CLI Version:      `3.13.2`
- Nitro Version:    `2.9.7`
- Package Manager:  `npm@10.8.2`
- Builder:          `-`
- User Config:      `compatibilityDate`, `devtools`
- Runtime Modules:  `-`
- Build Modules:    `-`

### Reproduction

https://github.com/phoenix-ru/h3-urlsearchparams-reproduction

### Describe the bug

Users of `@sidebase/nuxt-auth` are experiencing the following problem with our module, described [here](https://github.com/sidebase/nuxt-auth/issues/883):
```
ERROR  [nuxt] [request error] [unhandled] [500] The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received an instance of URLSearchParams
```

We managed to pinpoint the problem to this piece of code:
https://github.com/sidebase/nuxt-auth/blob/a0704470583c5fea47c26c14b8673f8bb6a39a5f/src/runtime/composables/authjs/useAuth.ts#L120-L133

It works well when being called using HTTP, but fails when calling Nuxt endpoints from within a server. As you know,`$fetch` calls are translated to function calls in server code.

Upon some investigation, I was able to find the exact culprit - it is the `readRawBody` method and its `Buffer.from` return statement:
https://github.com/unjs/h3/blob/112fa338ef5bbde6207f3317d2ef51f0e52a59fa/src/utils/body.ts#L93

This is what happens:
- during a regular HTTP call it looks like `$fetch` serializes `URLSearchParams` and is therefore capable to deserialize them later;
- this doesn't happen during a function call - `readRawBody` receives `URLSearchParams` as-is;
- `URLSearchParams` successfully evades all the `if` checks and ends up in the last return statement.

I fixed it locally and verified the patch, I will submit a PR shortly.

### Additional context

_No response_

### Logs

_No response_

phoenix-ru:
resolves #887

### DIFF
--- a/src/utils/body.ts
+++ b/src/utils/body.ts
@@ -90,6 +90,8 @@ export function readRawBody<E extends Encoding = "utf8">(
       if (_resolved.constructor === Object) {
         return Buffer.from(JSON.stringify(_resolved));
       }
+      // TODO: Handle other BodyInit types
+      // https://developer.mozilla.org/en-US/docs/Web/API/Response/Response#body
       if (_resolved instanceof URLSearchParams) {
         return Buffer.from(_resolved.toString());
       }
--- a/src/utils/body.ts
+++ b/src/utils/body.ts
@@ -91,7 +91,7 @@ export function readRawBody<E extends Encoding = "utf8">(
         return Buffer.from(JSON.stringify(_resolved));
       }
       if (_resolved instanceof URLSearchParams) {
-        return Buffer.from(_resolved.toString())
+        return Buffer.from(_resolved.toString());
       }
       return Buffer.from(_resolved);
     });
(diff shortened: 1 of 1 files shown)


alimozdemir:
Could we have the review and approval for this? @pi0 @danielroe It's blocking a lot of people

julienguillot77:
Hello guys,
Sorry to rush you but can you act quickly to approve this PR and integrate it into a release? I am facing this error in my professional project and the deadline is January 2025. Thanks in advance.

pi0 (approved):

pi0:
Thanks. I added a TODO, we might handle other valid bodyinit types also in the future.

PR #890, opened 2024-09-30
feat(validate): provide validate error in `cause`

JianJroh:
To be more friendly to catch third-party errors with `cause`


pi0 (approved):
Thanks!

PR #900, opened 2024-10-15
docs(app-instance): capitalize text

iiio2:


pi0:
thanks for PR but case is fine.

PR #906, opened 2024-10-28
docs: include xForwardedFor param on getRequestIP

RihanArfan:
Adds a missing parameter on the docs for `getRequestIP()`. May be worth backporting to v1 docs too.
https://github.com/unjs/h3/blob/7324eeec854eecc37422074ef9f2aec8a5e4a816/src/utils/request.ts#L297-L307

### DIFF
--- a/docs/2.utils/1.request.md
+++ b/docs/2.utils/1.request.md
@@ -53,7 +53,7 @@ app.use("/", (event) => {
 });
 ```
 
-### `getRequestIP(event)`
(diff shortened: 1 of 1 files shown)


pi0:
Thanks for PR dear @RihanArfan. It is caused by an issue in [automd](https://github.com/unjs/automd) not preserving some params. 

Since it is an autogenerated content, your changes sadly will be lost after next lint `pnpm automd`. 

If you want to contribute and fix it, it is more than welcome I can give you some pointers 👍🏼

PR #910, opened 2024-11-08
docs: Add a util for handling proxy request.

yisibell:
Add a util for handling proxy request.

### DIFF
--- a/docs/2.utils/99.community.md
+++ b/docs/2.utils/99.community.md
@@ -27,3 +27,7 @@ You can use external h3 event utilities made by the community.
 ## `@intlify/h3`
 
 :read-more{to="https://github.com/intlify/h3"}
+
+## `h3-proxy`
+
+:read-more{to="https://github.com/yisibell/h3-proxy"}


pi0:
Would you like to contribute to improve `proxyRequest` features instead?

Since this is feature is deeply integrated (in nitro/nuxt route rules for example) it would be much better if we can improve core.

yisibell:
> Would you like to contribute to improve `proxyRequest` features instead?
> 
> Since this is feature is deeply integrated (in nitro/nuxt route rules for example) it would be much better if we can improve core.

It is more appropriate to use it as a community plugin, because the relationship between `h3-proxy' and 'proxyRequest' is equivalent to the relationship between 'http proxy middleware' and 'http proxy'.

> Would you like to contribute to improve `proxyRequest` features instead?
> 
> Since this is feature is deeply integrated (in nitro/nuxt route rules for example) it would be much better if we can improve core.

It is more appropriate to use it as a community plugin, because the relationship between `h3-proxy` and `proxyRequest` is equivalent to the relationship between [http-proxy-middleware](https://github.com/chimurai/http-proxy-middleware) and [http-proxy](https://github.com/http-party/node-http-proxy). 😄

PR #912, opened 2024-11-10
fix(withBase): update with `pathname` instead of `path`

### LINKED ISSUE #878: event.url.pathname is updated incorrectly in withBase method for NodeEvents
### Environment

Node.js v20.17.0
h3-nightly@2.0.0-1725475522.2d941d3

### Reproduction

Given this simple H3-nightly app running on node.js-
```ts
import { createH3, toNodeHandler, withBase } from "h3-nightly";
import { createServer } from "http";

const router = createH3()
    .get('/', (_event) => 'hello')
    .get('/abc', (_event) => 'abc')
    .use((event) => { console.log(event.path, event.pathname) })

const app = createH3()
    .use('/app/**', withBase('/app', router))

const port = 3000
createServer(toNodeHandler(app))
    .listen(port, () => console.log('started', `http://localhost:${port}`));
```

The url `'http://localhost:3000/app/abc/'` is resolved correctly.
But the same url with any searchParams like `'http://localhost:3000/app/abc/?'` is not resolved, sending 404.

### Describe the bug

The event path and pathname are updated incorrectly by the `set pathname`, when called from the `withBase` function.

### Additional context

_No response_

### Logs

_No response_

gulshan:
resolves #878

### DIFF
--- a/src/utils/base.ts
+++ b/src/utils/base.ts
@@ -20,7 +20,7 @@ export function withBase(base: string, input: EventHandler | H3): EventHandler {
 
   const _handler: EventHandler = async (event) => {
     const _pathBefore = event.url.pathname || "/";
-    event.url.pathname = withoutBase(event.path || "/", base);
+    event.url.pathname = withoutBase(event.pathname || "/", base);
     return Promise.resolve(_originalHandler(event)).finally(() => {
       event.url.pathname = _pathBefore;
     });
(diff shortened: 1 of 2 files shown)


pi0:
Can you please add a test as well? 🙏🏼

gulshan:
@pi0 The issue was only happening with `toNodeHandler` and most probably tied to how the `pathname` property is set in `NodeReqURLProxy`. But I put the fix in the common code. So, I am a bit confuse about how I should test it.

gulshan:
@pi0 Tried adding an integration test. Please check.

pi0 (approved):
Thanks ❤️

PR #922, opened 2024-11-30
docs: Update 1.index.md

danielrentz:
radix3 is now rou3


pi0:
thanks for PR, since it is on v1 branch it is fine to use `radix3` as name I guess (and link redirects), we soon swap it with main/v2 branch.

PR #930, opened 2024-12-08
perf(adapters/node): check _hasBody once

markthree:
It seems to avoid double-checking here, if there are body 🥰


pi0:
Thank you for PR. cross merged to unjs/srvx (https://github.com/unjs/srvx/commit/978a27d666476460091f360dda30d2e4fc47240d)

PR #849, opened 2024-07-24
docs: fix typo

jordypereira:

PR #941, opened 2025-01-02
docs: fix small typo in streaming docs

cjpearson:
#849 fixed one of these in the `v1` branch, but the live site seems to be showing the `main` branch content.


nozomuikuta (approved):
Thank you!

PR #951, opened 2025-01-14
docs(event-handler): provide clarity on package

iiio2:
actually `example-node-middleware` does not exist. I have given a comment on it so that it provides clarity like this for "web-handler" (https://h3.unjs.io/guide/event-handler#converting-from-web-handlers).


nozomuikuta (approved):
Thank you!

Issue #963, opened 2025-01-28
feat: Standard Schema validation


sandros94:
### Describe the feature

Considering that [`Standard Schema`](https://github.com/standard-schema/standard-schema) reached a stable release and it is gaining more and more traction, what are the plans for `getValidatedQuery`, `getValidatedRouterParams` and `readValidatedBody`?

It would be nice to support Standard Schema natively so that developers are free to use their own validation library without creating a dedicated adapter for each, nor a custom function returning a boolean as it currently is for `validateData`.

This can be implemented without introducing additional dependencies and, in theory, without breaking changes

### Additional information

- [x] Would you be willing to help implement this feature?

PR #964, opened 2025-01-28
feat(validation): implement Standard Schema

sandros94:
Resolves: #963 

- [x] implementation
- [x] tests (added missing `getValidatedRouterParams` tests)
- [x] docs

### DIFF
--- a/docs/2.utils/1.request.md
+++ b/docs/2.utils/1.request.md
@@ -131,9 +131,9 @@ app.use("/", (event) => {
 
 ### `getValidatedQuery(event, validate)`
 
-Get the query param from the request URL validated with validate function.
+Get the query param from the request URL validated with validate function or validation schema.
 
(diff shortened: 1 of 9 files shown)


sandros94:
I was also considering an optional argument for `readValidatedBody` to convert incoming `yaml` and `toml` to `json` using `confbox`.
Could this be implemented in this PR or a new one? 🤔

sandros94:
> Cool! Why not using package? @standard-schema/spec

For this use-case it is not strictly needed and as such I personally prefer to keep the deps counter as low as possible 😬

pi0:
Thanks for PR dear @sandros94 sorry it got stalled with merge conflits.

I have added a smaller subset for basic support in #1068 please feel free to ammend and improve if you have any feedbacks 🙏🏼

PR #971, opened 2025-02-06
docs: move links from radix3 to rou3

patak-cat:
The current radix3 link is redirecting to rou3, so the docs should probably have rou3 directly.

I was a bit confused about this issue/discussion still being open though. Should it be closed?
- https://github.com/unjs/rou3/issues/108

### DIFF
--- a/.github/workflows/autofix.yml
+++ b/.github/workflows/autofix.yml
@@ -13,10 +13,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - run: corepack enable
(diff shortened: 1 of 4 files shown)


nozomuikuta (approved):
Thank you for the PR!

Currently CI is not working due to `actions/setup-node@v4`, so I keep it unmerged just in case (even if I merged, CI will fail in `main` branch).

pi0:
Thanks for PR, just saw it. main branch / h3 v2 (and rou3) is not yet active. 

I will fix CI and update it as soon as could.

PR #973, opened 2025-02-18
fix(router): normalize added routes

huseeiin:
now you can add h3 routes with arabic letters.

### DIFF
--- a/src/h3.ts
+++ b/src/h3.ts
@@ -130,6 +130,7 @@ export const H3 = /* @__PURE__ */ (() => {
               ...(_handleMiddleware.middleware || []),
             ].filter(Boolean)
           : undefined;
+      route = new URL(route, "h://_").pathname;
       addRoute(this.#router, _method, route, {
         method: _method as HTTPMethod,
         route,
--- a/test/app.test.ts
+++ b/test/app.test.ts
@@ -241,6 +241,13 @@ describeMatrix("app", (t, { it, expect }) => {
     expect(await res.text()).toBe("valid");
   });
 
+  it("can add arabic routes", async () => {
+    t.app.get("/عربي", () => "valid");
+
+    const res = await t.fetch("/عربي");
+    expect(res.status).toBe(200);
+  });
+
   it.skipIf(t.target !== "node")(
     "wait for node middleware (req, res, next)",
     async () => {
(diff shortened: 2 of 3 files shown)


huseeiin:
@pi0 you might need to rerun prettier because i didn't have time to do that

src/h3.ts
```diff
@@ -277,7 +277,7 @@ class _H3 implements H3 {
     }
     const _method = (method || "").toUpperCase();
     const _handler = (handler as H3)?.handler || handler;
-    addRoute(this._router, _method, route, <H3Route>{
+    addRoute(this._router, _method, new URL(route, "http://localhost").pathname, <H3Route>{
```
pi0: We should probably do manual encoding (or do it behind a flag) since `URL` constructor adds runtime overhead

pi0:
(from chat)

```js
const preservedChars = [
  "%",
  "[",
  "]",
  "!",
  "*",
  "'",
  "(",
  ")",
  ";",
  ":",
  "@",
  "&",
  "=",
  "+",
  "$",
  ",",
  "/",
  "?",
  "#",
  "-",
];

/**
 * Transform the `pathname` segment of a URL like the browser does.
 */
export function sanitizePathname(url: string) {
  url = String(url);

  const segments = url.replace(/\\/g, "/").split("/");

  const normalizedSegments: string[] = [];

  for (const segment of segments) {
    if (segment === "") {
      continue;
    }

    if (segment === ".") {
      continue;
    }

    if (segment === "..") {
      if (normalizedSegments.length > 0) {
        normalizedSegments.pop();
      }
    } else {
      normalizedSegments.push(segment);
    }
  }

  const result = normalizedSegments
    .join("/")
    .split("")
    .map((char) =>
      preservedChars.includes(char) ? char : encodeURIComponent(char)
    )
    .join("");

  return result.startsWith("/") ? result : `/${result}`;
}
```

pi0 (approved):
Thanks lets give it a try with `URL` normalizer and optimize if needed

huseeiin:
if http://localhost was changed to h://_ for minification purposes, i want to add that this does not behave the same.

i know that it doesn't add a `/` at the beginning:
`new URL('', 'h://_').pathname` returns `''`
`new URL('', 'http://localhost').pathname` returns `'/'`

but this is fine because `h3` already does that. **no bugs were detected yet :)**.

pi0:
Intresting finding @huseeiin 

Yes it was to save few bytes but not worth for `3` less we can use `http://_`, do you mind to make a PR? 🙏🏼 

image

PR #974, opened 2025-02-19
fix(serveStatic): fall through to the next handler when `fallthrough` is set to true

### LINKED ISSUE #904: serveStatic's fallthrough option does not function as intended (possible fix included)
### Environment

- pnpm 9.10.0
- NodeJS 20.16.0

### Reproduction

```ts
import { createApp, defineEventHandler, serveStatic } from "h3";
import { readFile, stat } from "fs/promises";
import { join } from "path";

const publicDir = "public";

app.use(
  "/",
  defineEventHandler(async (event) => {
    return await serveStatic(event, {
      getContents: async (id) => await readFile(join(publicDir, id)),
      getMeta: async (id) => {
        const stats = await stat(join(publicDir, id)).catch(() => undefined);

        if (!stats || !stats.isFile()) {
          return;
        }

        return {
          size: stats.size,
          mtime: stats.mtime,
        };
      },
      fallthrough: true,
    });
  })
);

app.use(
  "/",
  defineEventHandler((event) => {
    return "Hello world";
  })
);

```

### Describe the bug

Adding `fallthrough: true` option to serveStatic does not pass the request to the next event handler/middleware and returns `false` which h3 interprets as json.

I think I identified the offending code in `src/utils/static.ts`, removing it fixes the problem but I don't know if it breaks anything else yet.
on line 24: `return false` should be `return` or `return undefined`
same thing on line 65

Refer to the [registering event handlers section](https://h3.unjs.io/guide/app#registering-event-handlers) in the app instance docs.

I can submit a PR with a test and the fix if needed.

### Additional context

_No response_

### Logs

_No response_

OmarMAttia7:
Resolves #904, includes a new test suite "serve static with fallthrough".
Minimal reproduction is included with the original issue.


pi0 (approved):
thnx ❤️

PR #978, opened 2025-02-20
feat: add status message based on status code common errors

IqroNegoro:
as the title say, i just add default statusMessage based on statusCode, i kinda not like that statusMessage is always empty and must explicitly add in any createError to show it up. especially in Nuxt server api response.

assume we have some error :
```js
} catch (error : any) {
    // return fields error and it message if validation error but not with statusMessage "Bad Request"
    if (error instanceof ValidationError) {
        throw createError({
           statusCode: 400,
           message: "Validation Error",
           data: {
             field: 'error message',
             // ...another error
          });
    }
    // return error 500 with Internal Server Error status message
   throw createError({
      statusCode: 500,
      statusMessage: "Internal Server Error",
      message: "Internal Server Error"
    });
}
```
with default statusMessage, we not have to explicitly add statusMessage on any createError, default statusMessage with handle it.

```js
throw createError({
  statusCode: 400,
  message: "Validation Error",
  data: {
    field: 'error message',
    // ...another error
});

// response HTTP
{
  statusCode: 400,
  statusMessage: "Bad Request"
  // ... another field
}
```
just reducing unnecessary repetition for good http responses...

with ```pnpm dev``` for testing any errors :

image

### DIFF
--- a/src/error.ts
+++ b/src/error.ts
@@ -136,6 +136,67 @@ export function createError<DataT = unknown>(
     err.statusMessage = sanitizeStatusMessage(statusMessage);
   }
 
+  if (!statusMessage && statusCode) {
+    switch (statusCode) {
+        case 400: {
+          err.statusMessage = "Bad Request";
+          break;
+        }
+        case 401: {
+          err.statusMessage = "Unauthorized";
+          break;
+        }
+        case 403: {
+          err.statusMessage = "Forbidden";
+          break;
+        }
+        case 404: {
+          err.statusMessage = "Not Found";
+          break;
+        }
+        case 405: {
+          err.statusMessage = "Method Not Allowed";
+          break;
+        }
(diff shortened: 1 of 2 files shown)


pi0:
Thank you so much for your pull request, while it is a nice addition, this will add extra bytes to h3 server code and HTTP clients (fetch, browser, curl, etc) already have a built-in list for default status code => message to show and having it partially is inconsistent / having fully, too much.

IqroNegoro:
> Thank you so much for your pull request, while it is a nice addition, this will add extra bytes to h3 server code and HTTP clients (fetch, browser, curl, etc) already have a built-in list for default status code => message to show and having it partially is inconsistent / having fully, too much.

built-in list for default status code => message? May I know the docs or something about this?

PR #980, opened 2025-02-21
fix: properly merge headers in fetchWithEvent

brc-dd:
fixes https://github.com/nuxt/nuxt/issues/31073

Currently, the merging behavior is inconsistent between useRequestFetch and client $fetch because of spreading headers without normalization. This PR fixes that by doing essentially same as what's done above in that file:

https://github.com/unjs/h3/blob/0a796ff1b8b570791386045881ca739942065460/src/utils/proxy.ts#L37-L42


pi0 (approved):
thanks! sorry i missed this

Issue #994, opened 2025-03-19
support for `JOSE` (Javascript Object Signing and Encryption)


sandros94:
This all started while doing some considerations for [`nuxt-auth-utils`](https://github.com/atinux/nuxt-auth-utils), but soon realized this could be easily implemented upstream, here.

Essentially, my idea is to introduce a new utility to manage `JOSE`s (Javascript Object Signing and Encryption), more commonly referred to JWT, although it is only one of them.

A simple definition about them is from a [this stackoverflow answer](https://stackoverflow.com/a/74257561):
> JOSE stands for JSON Object Signing and Encryption. It's a set of standards used to sign and encrypt data using JSON-based data structures. In other words, JWA, JWE, JWK, JWS, and JWT are all part of JOSE.
>
> TL;DR:
>
> - JWA: Defines a set of crypto algorithms and identifiers used in JWE, JWK and JWS.
> - JWK: Defines a representation of crypto keys used as input for the algorithms defined in JWA using JSON-based data structures.
> - JWE: Defines encrypted content using JSON-based data structures. The encryption/decryption is performed with the algorithms defined in JWA.
> - JWS: Defines signed content using JSON-based data structures. The signing/verification is performed with the algorithms defined in JWA
> - JWT: Defines subject claims using JSON-based data structures. The claims can be optionally protected via JWE or JWS.
>
> [...]

### Going more indepth

Quick disclaimer: I haven't finished reading all RFCs ([JWA](https://datatracker.ietf.org/doc/html/rfc7518), [JWK](https://datatracker.ietf.org/doc/html/rfc7517), [JWT](https://datatracker.ietf.org/doc/html/rfc7519), [JWS](https://datatracker.ietf.org/doc/html/rfc7797), [JWE](https://datatracker.ietf.org/doc/html/rfc7516)), as I started this topic only recently and it is quite a read 😰...

Each one does not exclude the other, in fact most of the use for `JWT` is based on a combination of `JWA`, `JWK`, `JWS`/`JWE`. What they have in common is the data structure and their encoding.

[...]

PR #997, opened 2025-03-22
feat: use JWE internally to `useSession`

sandros94:
Related to #994

This is a first implementation to use `JWE` ([RFC7516](https://datatracker.ietf.org/doc/html/rfc7516)) instead of `iron-crypto` as the internal `useSession` encryption.

The main goal is to provide better support for cross-platform/cross-language support in scenarios like edge networks and proxies.

### DIFF
--- a/src/types/utils/session.ts
+++ b/src/types/utils/session.ts
@@ -1,5 +1,5 @@
 import type { CookieSerializeOptions } from "cookie-es";
-import type { SealOptions } from "../../utils/internal/iron-crypto";
+import type { JWEOptions } from "../../utils/internal/jwe";
 import type { kGetSession } from "../../utils/internal/session";
 
 type SessionDataT = Record<string, any>;
@@ -24,7 +24,8 @@ export interface SessionConfig {
   cookie?: false | CookieSerializeOptions;
   /** Default is x-h3-session / x-{name}-session */
   sessionHeader?: false | string;
-  seal?: SealOptions;
+  /** JWE options for encryption/decryption */
+  jwe?: Partial<JWEOptions>;
   crypto?: Crypto;
   /** Default is Crypto.randomUUID */
   generateId?: () => string;
--- a/src/utils/internal/jwe.ts
+++ b/src/utils/internal/jwe.ts
@@ -213,6 +213,10 @@ export async function unseal(
   }
 
   // Check expiration
+  if (result.exp && typeof result.exp !== "number") {
+    throw new Error("Invalid expiration");
+  }
+
   if (result.exp && result.exp <= now - opts.timestampSkewSec * 1000) {
     throw new Error("Token expired");
   }
--- a/src/utils/internal/jwe.ts
+++ b/src/utils/internal/jwe.ts
@@ -1,0 +1,221 @@
+import crypto from "uncrypto";
+import {
+  base64Decode,
+  base64Encode,
+  textDecoder,
+  textEncoder,
+} from "./encoding";
+
+/**
+ * JWE (JSON Web Encryption) implementation for H3 sessions
+ */
+
+export interface JWEOptions {
+  /** Expiration time in milliseconds where 0 means forever. Defaults to 0. */
+  ttl: number;
+  /** Number of seconds of permitted clock skew for incoming expirations. Defaults to 60 seconds. */
+  timestampSkewSec: number;
+  /** Local clock time offset in milliseconds. Defaults to 0. */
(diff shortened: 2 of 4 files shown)


sandros94:
With https://github.com/unjs/h3/pull/997/commits/d0d0de25709b01fab2f7518ea9e4544b8b140b7f I decided to restart from scratch, as I was having issues when testing the JWEs against other tools. But while doing so I've noticed an issue with the current implementation of `base64Encoder` and `base64Decoder` in the encoder utils, in particular with tampering, thus I preferred to create two dedicated ones (in name of security 😬)

My next steps will be:
1. try to support other password-based algorithms and encoders out of the box
2. simplify the payload input (since it will be stringified anyway before encryption)
3. list and fully test with JWE implementations from other languages and ecosystems

casualmatt:
I wrote a simple test to check the compatibility of the generated JWE, and the results look good so far. I tested against two Go libraries and NGINX’s JWT module:

[Golang JWE test suite](https://github.com/JAMflow-Cloud/jwe-test-suite)
[NGINX JWT auth module](https://nginx.org/en/docs/http/ngx_http_auth_jwt_module.html)

pi0:
Thank you for initiating this dear @sandros94. 

While JWE migration looks promising, I prefer to keep backward compatibility with ironcrypto format so users can choose either and won't cause incompatibility with current usages.

It would be cool if at least in the first PR step, we minimize changes to the session.ts and postpone changes like the second expiration. 

Also to reduce bundle size we can share utils as much as possible. You mentioned there is an issue in the current base64 implementation (it is ported from Deno utils) can you elaborate more on this?

Thank you for helping on integration tests dear @casualmatt ❤️ your repo link gives 404 9is it private?)

sandros94:
> Thank you for initiating this dear @sandros94.
> 
> While JWE migration looks promising, I prefer to keep backward compatibility with ironcrypto format so users can choose either and won't cause incompatibility with current usages.
> 
> It would be cool if at least in the first PR step, we minimize changes to the session.ts and postpone changes like the second expiration.
> 
> Also to reduce bundle size we can share utils as much as possible. You mentioned there is an issue in the current base64 implementation (it is ported from Deno utils) can you elaborate more on this?

I haven't deeply looked at the current base64 src code yet. But essentially it was failing the [tampering test](https://github.com/JAMflow-Cloud/h3/blob/4f29e7ea611d5d594a38348d0f067c8b5ca6f71f/test/unit/jwe.test.ts#L84-L109). This might be due to the fact that the deno fork is able to take care about any potential tampering and reconstruct the original data, but in a JWS/JWE implementation (as far as I understand it) it should be treated as an _illegal_ operation and error out.

I'll look into other potential utils that  be shared, one of which I can think of are key generation and wrapping (there might be some shared crypto logic)

> Thank you for helping on integration tests dear @casualmatt ❤️ your repo link gives 404 9is it private?)

It is indeed private, sorry. I'm sure he'll publish it once he notices it

sandros94:
Also, on the topic of `useSession` API, JWE does support zip compression, but it requires full control of the payload (meaning that it should be compressed AFTER serialization, something improssilbe to do with the current implementation).

Please feel free to tell me if this might be too much of a stretch, but I also think this might be easily achievable with a custom `useSession` implementation that uses directly the `seal`/`unseal` utilities

casualmatt:
> 404 9is it private?
👉👈 ehh sorry 😁 it should be public now; I will probably add some bun and node test too, with a bash script to check a JTE token against all of them

sandros94:
Sorry @pi0 if I haven't worked on this, yet.

I cannot stop thinking about the data invalidation issue (that was also pointed out in this issue https://github.com/unjs/h3/issues/1004) while keeping a feature parity with iron-crypto implementation

sandros94:
I've been considering doing this for so long now, but ended up writing the underlying reason only now in a different PR ([here](https://github.com/h3js/h3/pull/1298#issuecomment-3862570269) and [the comment right after](https://github.com/h3js/h3/pull/1298#issuecomment-3862596953))

But essentially: the current Iron Crypto implementation, while not my favorite, is good enough for the general use that it offers and perfectly covers. While JWE, and those specs that build upon it, does require some changes that may disrupt too many in the community. Not to mention that at the current state the JWE implementation is lacking many aspects that I've learned in the months after.

For those that are interested in supporting JWT in their h3 projects, feel free to take a look at [`unjwt`](https://github.com/sandros94/unjwt) which also offers adapters for h3 like `useJWESession` and `useJWSSession`.

PR #1000, opened 2025-03-31
chore: include `node_modules` in `.prettierignore`

iiio2:


pi0:
thanks for PR but i think even with `prettier -c .` `node_modules` is already ignored and it is not needed (if I'm wrong and there is a case please lmk to reopen)

iiio2:
Oh, Ok @pi0 . Thank you.

PR #1010, opened 2025-04-02
fix: make session clearing resilient

### LINKED ISSUE #1004: Reduce session TTL immutability or stickiness
## Background
A common source of frustration with the `nuxt-auth-utils` library is the inability to refresh session cookies. The data within the cookie can be mutated and updated, but the expiration time can't be updated once the session is created.

### `nuxt-auth-utils` issues
https://github.com/atinux/nuxt-auth-utils/issues/356
https://github.com/atinux/nuxt-auth-utils/issues/294
https://github.com/atinux/nuxt-auth-utils/issues/314

The fundamental issue appears to be that the session `createdAt` date introduced in https://github.com/unjs/h3/pull/325 is set at session creation and there is no way to update this date again. Mutations on the session data can be made, but the original `createdAt` date is always used for both the session TTL and the `expires` date.

https://github.com/unjs/h3/blob/087d0639cc12a555a87d56fc0ef3cca1fcf282cf/src/utils/session.ts#L139-L146

Having a session with an immutable TTL isn't a huge issue in isolation because in theory you can simply create a new session with a more recent `createdAt`. That's where the **stickiness** issue comes into play.

While you can `clear` the session, an attempt to create a new session in the same `H3Event` call will result in getting back the same session (new ref, same data) because it always tries to restore a session's event before it creates a new session.

`updateSession` function calls `getSession` if no session exists:
https://github.com/unjs/h3/blob/087d0639cc12a555a87d56fc0ef3cca1fcf282cf/src/utils/session.ts#L123-L126

### 1) Prevent `updateSession` from restoring sessions
Independent of the need to refresh sessions, the fact that a call to `useSession().clear()` can be effectively reversed by a call to `useSession().update()` is counter-intuitive and probably a little dangerous, so at minimum the `updateSession` function should use an explicit (new) `createSession` function OR the `getSession` function be modified to support a `doNotRestoreSession` option.

https://github.com/unjs/h3/blob/087d0639cc12a555a87d56fc0ef3cca1fcf282cf/src/utils/session.ts#L123-L126

### 2) Add a `rotateSession` function
While not explicitly necessary if the `updateSession` bug is resolved, the ability to have an obvious and clear function for rebuilding the session using a new `createdAt` value would be helpful. I would expect it would little more than a wrapper around conjoined calls to the `clearSession` and `updateSession` that would accept an optional `SessionUpdate<T>` parameter.

### Additional information

nerdoza:
This PR addresses an issue in the session manager utility which results in an already cleared session returning when `useSession`, `session.update`, `getSession`, `updateSession`, or `sealSession` are used after calling `session.clear` or `clearSession`.

This ensures that after clearing the session, a new blank session is initialized on the event context so that subsequent session mutation calls cause updates on the newly initialized session, rather than restoring the original session from the event context.

There are still edge-cases where the session can be restored, but they require subverting the use of the provided methods and session manager. This fix largely relies on the understanding that the `event.context.sessions` object is the source of truth for expected session state, but in the absence of a valid session in the context, the original session will still be reconstructed.

The documentation for the session management has also been updated to add a section on session rotation, which should lend in resolving issues with the downstream `nuxt-auth-utils` issues.

Closes #1004

### DIFF
--- a/docs/4.examples/handle-session.md
+++ b/docs/4.examples/handle-session.md
@@ -37,7 +37,7 @@ This will initialize a session and return an header `Set-Cookie` with a cookie n
 If the request contains a cookie named `h3` or a header named `x-h3-session`, the session will be initialized with the content of the cookie or the header.
 
 > [!NOTE]
-> The header take precedence over the cookie.
+> The header takes precedence over the cookie.
 
 ## Get Data from a Session
 
@@ -86,6 +86,8 @@ We try to get a session from the request. If there is no session, a new one will
 
 Try to visit the page multiple times and you will see the number of times you visited the page.
 
+If a `maxAge` is configured, the expiration date of the session will not be updated with a call to `update` on the session object, nor with a call using the `updateSession` utility. For details on how to update teh expiration date, see the [rotate session section](#rotate-a-session).
+
 > [!NOTE]
 > If you use a CLI tool like `curl` to test this example, you will not see the number of times you visited the page because the CLI tool does not save cookies. You must get the cookie from the response and send it back to the server.
 
@@ -108,6 +110,30 @@ app.use("/clear", async (event) => {
 ```
 
 h3 will send a header `Set-Cookie` with an empty cookie named `h3` to clear the session.
+
+## Rotate a Session
+
+The session identifier and expiration date are immutable. If a `maxAge` is configured for the session, the expiration date is set when the session is created and there is no way to extend the session beyond the initially determined expiration date.
+
+If the session context needs to be extended, the session must be rotated. This means the current session must first be cleared and a new session must be created. A new session identifier will be created and the expiration date will be set to the current time plus the `maxAge` value.
+
+```js
+import { useSession } from "h3";
+
+app.use("/rotate", async (event) => {
+  const session = await useSession(event, {
+    password: "80d42cfb-1cd2-462c-8f17-e3237d9027e9",
+    maxAge: 60 * 60 * 24 * 7, // 7 days
+  });
+
+  const data = session.date;  // Retrieve the current session data
+  await session.clear();      // Clear the current session
+  await session.update(data); // Create a new session with the original data
+
+  return "Session rotated";
+});
+
+```
 
 ## Options
 
--- a/docs/4.examples/handle-session.md
+++ b/docs/4.examples/handle-session.md
@@ -86,7 +86,7 @@ We try to get a session from the request. If there is no session, a new one will
 
 Try to visit the page multiple times and you will see the number of times you visited the page.
 
-If a `maxAge` is configured, the expiration date of the session will not be updated with a call to `update` on the session object, nor with a call using the `updateSession` utility. For details on how to extend the session's expiration date, see the [Rotating a Session section](#rotating-a-session).
+If a `maxAge` is configured, the expiration date of the session will not be updated with a call to `update` on the session object, nor with a call using the `updateSession` utility.
 
 > [!NOTE]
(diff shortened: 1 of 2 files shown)


sandros94 (commented):
Just out of curiosity: I imagine [this](https://github.com/unjs/h3/pull/1010/files/16bb190aae2d6397316edaf40724d5a1b240d4e5#diff-18d5f672d3dc05dd3257bf22180a2b97abbc517bee318e14ad0f1b6b7acacd97R233-R238) will have a cost on memory 🤔

```ts
  // Initialize a new empty session object
  event.context.sessions![sessionName] = {
    id: generateId(config),
    createdAt: Date.now(),
    data: new EmptyObject(),
  } satisfies Session;
```

PS: I messed up the comment 🙄

nerdoza:
@pi0, this PR is ready for review and possibly merge.

docs/4.examples/handle-session.md
```diff
@@ -109,6 +111,29 @@ app.use("/clear", async (event) => {
...
+```js
+import { useSession } from "h3";
+
+app.use("/rotate", async (event) => {
+  const session = await useSession(event, {
+    password: "80d42cfb-1cd2-462c-8f17-e3237d9027e9",
+    maxAge: 60 * 60 * 24 * 7, // 7 days
+  });
+
+  const data = session.date; // Retrieve the current session data
```
pi0, line 129:
```suggestion
  const data = session.data; // Retrieve the current session data
```

pi0:
Thanks for well well-written explanations.

For the purpose of session rotation, I believe we might need a new util or at least flag on clear, normal `clear()` means logout.  Also, by havingan  explicit way of calling rotate, we can make user usage easier to not make them cache previous value and restore with 3 lines of code.

The refactors of this PR look nice addition btw. Can you please convert this PR to a refactor only and then a subsequent to draft idea for explicit rotate?

nerdoza:
@pi0, updates to remove documentation about session rotation have been applied. I agree with your point about `clear()` and for clarity this implementation does remove the data and does setup a new session stub to prevent session restoration by the next session call, but because it bypasses the cookie setting phase it won't actually propagate a new session until `updateSession()` is called.

src/utils/session.ts
```diff
@@ -198,12 +218,21 @@ export async function unsealSession(
...
-  const sessionName = config.name || DEFAULT_SESSION_NAME;
-  if (event.context.sessions?.[sessionName]) {
-    delete event.context.sessions![sessionName];
+  const sessionName = getSessionName(config);
+  if (!event.context.sessions) {
+    event.context.sessions = new EmptyObject();
   }
+
+  // Clobber the original session with a new session
+  event.context.sessions![sessionName] = {
```
pi0, line 229: This should be reverted to allow this PR move forward as a refactor step.

nerdoza: I'm a bit confused @pi0. This line is the main point of this PR. If you want to isolate changes I'd be more apt to revert the refactor changes on this PR and apply them on a separate PR intended for refactor changes. Thoughts?

pi0: We can do this as part of #1022 for session rotate. clear purpose is to destroy session fully not to create a new one.

nerdoza: @pi0, I'm in complete agreement on that point, but the implementation at the code level is not 1:1 with the user experience. When the user calls `clear()` they want the session cleared, and then if they call `update()` they would expect that this creates a new session (or at least that's what I would assume). Because the `update()` call is eagerly restoring the session from the `request`, calling `update()` returns a mutation of the same session that was previously cleared.

**TLDR;** If a new session is not set on the `context.sessions[sessionName]` when `clear()` is called, then the session isn't actually called, it's just temporarily reset (and will be restored on next access call).

PR #1012, opened 2025-04-06
fix: Support for nested BigInt serialization through JSON replacer

victorpagan:
Closed https://github.com/unjs/h3/pull/1007 because of mistakes. I redid this and updated it to reflect the latest changes.

Original issue here https://github.com/unjs/h3/issues/474 and mine https://github.com/unjs/h3/issues/995

I created a test for it as well. Appears to satisfy the use case (and hopefully others). I left the replacer open ended. I noticed that this would potentially solve a small oversight with the ```isJSONSerializable``` function since it does not check for serializable primitives nested inside of Arrays/Objects.

### DIFF
--- a/src/response.ts
+++ b/src/response.ts
@@ -2,7 +2,7 @@ import type { H3Config, H3Event } from "./types";
 import type { H3Error } from "./types/h3";
 import { Response as SrvxResponse } from "srvx";
 import { createError } from "./error";
-import { isJSONSerializable } from "./utils/internal/object";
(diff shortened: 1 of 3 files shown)


pi0:
Appreciate the PR.

When returning an object value from h3 handlers, it directly passes to `JSON.stringify`, the same behavior of `Response.json({})`. Sadly, standards don't allow bigint, and we cannot diverge behavior from standards, at least for the default behavior in h3.

PR #1016, opened 2025-04-08
fix playground

### LINKED ISSUE #1015: Playground does not work anymore
### Describe the change

Hi 👋 . I work at StackBlitz and I just noticed that our h3 starter on our homepage does not work anymore. The reason is that the `app.ts` file got removed in the `playground` directory, only containing a `package.json` now https://github.com/unjs/h3/commit/632952b4ea8c4f093825116b7ce336bccc1b0388#diff-09c6c4e7a13cd6019c71f6b3e7e9c87c0f5dc9b890febb0b94db371e70e8392c.

Would it be possible to bring this back :)?

### URLs

_No response_

### Additional information

- [x] Would you be willing to help?

SamVerschueren:
resolves #1015

This PR adds the `app.ts` back to the playground so it works again on StackBlitz.

### DIFF
--- a/playground/app.ts
+++ b/playground/app.ts
@@ -1,0 +1,13 @@
+import { createApp, createRouter, eventHandler } from "h3";
+
+export const app = createApp();
+
+const router = createRouter();
+app.use(router);
+
+router.get(
+  "/",
+  eventHandler((event) => {
+    return { path: event.path, message: "Hello World!" };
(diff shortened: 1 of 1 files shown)


pi0:
Oh how i misseed this 🙈 d51850ecc06f16fae1ba8d13908b200de21134f3

PR #1023, opened 2025-04-16
docs: update for v2 usage

pi0:


docs/1.guide/1.index.md
```diff
@@ -8,91 +8,97 @@ icon: ph:book-open-duotone
...
-
-## Installing h3
-
-You can import h3 either as an npm package or from CDN.
+::tip
+Router is powered by [🌳 rou3](https://github.com/unjs/rou3), an ultra fast and tiny route matcher engine.
+::
 
-### Install as npm package
+What is beautiful in h3 is that all you have to do to make a response, is to simply return it! Responses can be simple **string**, **JSON object**, **binary data**, **stream** or standard [Web Response](https://developer.mozilla.org/en-US/docs/Web/API/Response/Response).
```
nozomuikuta, line 83: just a nitpick: adding a comma before the last option.

```suggestion
What is beautiful in h3 is that all you have to do to make a response, is to simply return it! Responses can be simple **string**, **JSON object**, **binary data**, **stream**, or standard [Web Response](https://developer.mozilla.org/en-US/docs/Web/API/Response/Response).
```

docs/1.guide/2.event-handler.md
```diff
@@ -8,7 +8,7 @@ icon: fluent:target-24-regular
 
 After creating an [app instance](/guide/app), you can start defining your application logic using event handlers.
 
-An event handler is a function that receive an `Event` instance and returns a response. You can compare it to controllers in other frameworks.
+An event handler is a function that receive an `H3Event` context and returns a response.
```
nozomuikuta, line 11: nitpick: adding `s` to verb

```suggestion
An event handler is a function that receives an `H3Event` context and returns a response.
```

docs/1.guide/5.routing.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Registering Routes
+
+H3 app instance has a built in router system to register routes matching patterns with parameter or specific HTTP method.
+
+::tip
+Router is powered by [🌳 rou3](https://github.com/unjs/rou3), an ultra fast and tiny route matcher engine.
+::
+
+In the example below, we register a route requests to the `/hello` endpoint with HTTP GET method:
```
nozomuikuta, line 13: The part "we register a route requests to ..." would need some more words for collocation (between "route" and "requests"), I guess.

docs/1.guide/5.routing.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Middleware
+
+You can register [event handlers](/guide/event-handler) to app instance using the `app.use`:
+
+```js
+app.use("/hello", () => "Hello world!");
+```
+
+This will register the event handler to the app instance and will be called for every request starting with the prefix `/hello`. This means that the event handler will be called for `/hello`, `/hello/world`, `/hello/123`, etc.
```
nozomuikuta, line 80: If I'm correct, what will be called is the event handler, won't it?
Grammatically the subject is "This" (calling `app.use`) in this sentence.

```suggestion
This will register the event handler to the app instance, and it will be called for every request starting with the prefix `/hello` (i.e. `/hello`, `/hello/world`, `/hello/123`, etc.).
```

docs/1.guide/5.routing.md
```diff
@@ -0,0 +1,108 @@
...
+
+You can define multiple event handlers for the same route. h3 will try to to call them one by one in order of registration until one of them returns a response. This is called `stack runner`.
```
nozomuikuta, line 82: nipick: typo

```suggestion
You can define multiple event handlers for the same route. h3 will try to call them one by one in order of registration until one of them returns a response. This is called `stack runner`.
```

nozomuikuta, line 82: question: 

by the way, what does `stack runner` actually mean?
Is it an array of event handlers that is stacked, or is it the h3 instance (as what calls event handlers) where multiple event handlers are registered to a route?

docs/1.guide/5.routing.md
```diff
@@ -0,0 +1,108 @@
...
+  // No response returned
+});
+app.use("/", () => "Second");
+```
+
+If all handlers get called and no response is returned, h3 will end the request with 404 status response.
+
+> [!NOTE]
+> Using an empty `return` or `return undefined` make a [`404 Not Found`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404) status response.
+> Also using `return null` will make a [`204 No Content`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204) status response.
```
nozomuikuta, line 105: nitpick: fix typo and suggestion

```suggestion
> [!NOTE]
> Using an empty `return` or `return undefined` makes a [`404 Not Found`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404) status response.
> Also using `return null` makes a [`204 No Content`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204) status response.
```

docs/1.guide/5.routing.md
```diff
@@ -0,0 +1,108 @@
...
+
+> [!TIP]
+> Use `return {}` or `return ""` or `return true` to make an explicit response.
```
nozomuikuta, line 108: nitpick: suggestion

```suggestion
> [!TIP]
> Use `return {}`, `return ""`, or `return true` to make an explicit response.
```

nozomuikuta (commented):
The next version API looks so clean!!

I just read through the docs changes and here are some comments.

docs/1.guide/5.routing.md
pi0: stack is basically the array of global middleware. h3 v1 was only based on stack + addon router, in v2 we only use it for global middleware.

pi0: removed the stack part

PR #1026, opened 2025-04-22
fix: handle `Headers` when merging proxy headers

### LINKED ISSUE #865: mergeHeaders unexpected behaviour / type issue
### Environment

`Nuxt 3.12.2 with Nitro 2.9.6`
development mode

### Reproduction

```ts
const headers = new Headers()
if (session.token)
    headers.set("Authorization", `Bearer ${session.token}`)
else
    console.warn(`[PROXY] No token set`)
if (session.domain)
    headers.set("X-Domain", session.domain)
else
    console.warn(`[PROXY] No domain set`)

// note: no 'Authorization' header in the event headers
const mergedHeaders = mergeHeaders(event.headers, headers)

console.log(mergedHeaders.get("Authorization"))
// log: undefined
```

### Describe the bug

The `mergeHeaders` utility won't merge multiple instances of Headers. This is a result of the underlying logic using Object.entries to get a list of key-value pairs which returns an empty array on a Headers instance.

Since `Headers` satisfies the `HeadersInit` type, this should either work, or the types should be changed appropriately.
Once it's clear how we should continue, I will open a PR.

### Additional context

I found this issue when trying to use the proxyRequest / sendProxy utility, basically all of the other utilities depending on mergeHeaders suffer from the type / logic issue and it's not that trivial to find, because you think something is wrong in your code.

### Logs

_No response_

kricsleo:
resolves #865

### DIFF
--- a/src/utils/internal/proxy.ts
+++ b/src/utils/internal/proxy.ts
@@ -56,7 +56,9 @@ export function mergeHeaders(
   }
   const merged = new Headers(defaults);
   for (const input of _inputs) {
-    for (const [key, value] of Object.entries(input!)) {
+    const entries =
+      input instanceof Headers ? input.entries() : Object.entries(input);
+    for (const [key, value] of entries) {
       if (value !== undefined) {
         merged.set(key, value);
       }
--- a/test/proxy.test.ts
+++ b/test/proxy.test.ts
@@ -38,7 +38,7 @@ describeMatrix("proxy", (t, { it, expect, describe }) => {
             fetch,
             headers: { "x-custom1": "overridden" },
             fetchOptions: {
-              headers: { "x-custom2": "overridden" },
+              headers: new Headers({ "x-custom2": "overridden" }),
             },
           });
         });


src/utils/internal/proxy.ts
```diff
@@ -56,7 +56,9 @@ export function mergeHeaders(
   }
   const merged = new Headers(defaults);
   for (const input of _inputs) {
-    for (const [key, value] of Object.entries(input!)) {
+    const entries =
+      input instanceof Headers ? input.entries() : Object.entries(input);
```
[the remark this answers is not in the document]
kricsleo: Good idea 👍, done.

src/utils/internal/proxy.ts
```diff
@@ -56,9 +56,13 @@ export function mergeHeaders(
...
   for (const input of _inputs) {
-    for (const [key, value] of Object.entries(input!)) {
+    const entries =
+      typeof input.entries === "function"
+        ? input.entries()
+        : Object.entries(input);
+    for (const [key, value] of entries) {
       if (value !== undefined) {
-        merged.set(key, value);
+        merged.set(key.toString(), value);
```
pi0, line 65: is `.toString()`  for types only?

kricsleo: It's because `Array.entries()` => `[number, string][]`, whereas `Headers.set(name: string, value: string)`

pi0: might be typescript issue? how keys can be number?

pi0 (approved):

PR #1027, opened 2025-04-22
fix: handle `Headers` when merging proxy headers 

### LINKED ISSUE #865: mergeHeaders unexpected behaviour / type issue

kricsleo:
resolves #865

This backports the fix to v1. (see #1026 for v2)
If it's not needed, feel free to close it. ❤️

### DIFF
--- a/src/utils/proxy.ts
+++ b/src/utils/proxy.ts
@@ -259,8 +259,9 @@ function mergeHeaders(
   }
   const merged = new Headers(defaults);
   for (const input of _inputs) {
-    const entries =
-      typeof input.entries === "function"
+    const entries = Array.isArray(input)
+      ? input
+      : typeof input.entries === "function"
         ? input.entries()
         : Object.entries(input);
     for (const [key, value] of entries) {
--- a/src/utils/proxy.ts
+++ b/src/utils/proxy.ts
@@ -260,10 +260,12 @@ function mergeHeaders(
   const merged = new Headers(defaults);
   for (const input of _inputs) {
     const entries =
-      input instanceof Headers ? input.entries() : Object.entries(input);
+      typeof input.entries === "function"
+        ? input.entries()
+        : Object.entries(input);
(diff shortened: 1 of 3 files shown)


src/utils/proxy.ts
```diff
@@ -259,9 +259,14 @@ function mergeHeaders(
...
-    for (const [key, value] of Object.entries(input!)) {
+    const entries = Array.isArray(input)
+      ? input
+      : typeof input.entries === "function"
+        ? input.entries()
+        : Object.entries(input);
+    for (const [key, value] of entries) {
       if (value !== undefined) {
-        merged.set(key, value);
+        merged.set(key.toString(), value);
```
pi0, line 269: Can you please update same as #1026

kricsleo: Done

PR #1028, opened 2025-04-22
fix: ensure `callNodeListener` resolves after `res.end()` is called

### LINKED ISSUE #779: Bun adapter cannot load webpage unless cache is disabled
### environment

"h3": "^1.11.1",
"vite": "^5.2.11",
"typescript": "^5.0.0"
bun v1.1.13

### reproduction

combine bun, vite (dev server middleware mode), and h3.
attempt to serve the vite dev server through h3 with the bun adapter
go to localhost on your browser, now try disabling cache in the network tab

### describe the bug

the webpage will stall on "waiting for localhost" indefinitely.
turning off cache makes the webpage load near instantly.
my domain knowledge in http is not very good, but i presume it has something to do with the cache headers not being passed to bun.serve properly?

### additional context

this only happens with the bun adapter, using the node adapter on the bun runtime works completely fine

### logs

kricsleo:
resolves #779

This is a minimal quick fix for `h3@v1`. (I'm not sure if it still needs fixing, though 🤔)

The `res.once` binding was happening too late. Sometimes, `res.end()` would be called and the events would have already fired, preventing the callbacks from firing. This would cause the response to hang indefinitely.

The WIP `h3@2` doesn't need this fix because it binds earlier: https://github.com/h3js/h3/blob/ac609aae8fa93947b564de55e2d1a6bec5ad18ad/src/adapters.ts#L75-L80

### DIFF
--- a/src/adapters/node.ts
+++ b/src/adapters/node.ts
@@ -110,12 +110,13 @@ export function callNodeListener(
       }
       return err ? reject(createError(err)) : resolve(undefined);
     };
+    if (isMiddleware) {
+      res.once("close", next);
+      res.once("error", next);
+    }
     try {
       const returned = handler(req, res, next);
-      if (isMiddleware && returned === undefined) {
-        res.once("close", next);
-        res.once("error", next);
-      } else {
+      if (!isMiddleware || returned !== undefined) {
         resolve(returned);
       }
     } catch (error) {


pi0:
We have much better native support for Bun in v2. Bun should gurantee identical behavior to Node.js so it could have been an old bug.  it might be risky at this point for v1 to change.

PR #1029, opened 2025-04-22
refactor: improve types

pi0:
- Enable `isolatedDeclartions` to make sure all export types are explicit and not infered (WIP)
- Use `.ts` extension for imports to make sure `src/` is ESM-valid code can natively run in Node and Deno 
- Colocate util types in same file of utils
- Add `type: "module"` to pkg
- Remove legacy type exports
- Removed types headers (@danielroe working on a unified lib for types)


productdevbook:
Why isolatedDeclartions are needed? I know that roldown now supports without it

pi0:
> Why isolatedDeclartions are needed? 

Type safety first. We are explicit about every function. Also using isolatedDeclartions, builds are insanely faster since no `typescript` package is needed to infer.

PR #1033, opened 2025-04-23
fix(cache): correct comparison in cache headers

### LINKED ISSUE #562: handleCacheHeaders - modifiedTime isn't working properly when set as a string.
https://github.com/unjs/h3/blob/b7aca9614e7e8e8921fd54804463ce78fdb034f1/src/utils/cache.ts#L26-L33

In the documentation, the property modifiedTime is typed as follow : `modifiedTime?: string | Date;`
The issue here is that even though on line 27 the property is turned from 'string | Date' into a Date, on line 30, the comparison uses `opts.modifiedTime: string | Date` rather than `modifiedTime: Date`.

Therefore the `cacheMatched` value will always return false when the modifiedTime options is set with a string.

kricsleo:
resolves #562


pi0 (approved):
thnx!

PR #1035, opened 2025-04-23
chore: remove duplicate `type` fields

kricsleo:
This PR removes the duplicate `type` fields.


pi0:
Thanks for PR ❤️ fixed in main

PR #1040, opened 2025-04-24
fix(node): pass through piped stream

### LINKED ISSUE #1039: h3@v2 - Response `.pipe()` not working as expected (breaks Vite public asset serving)
### Environment

Node v23.7.0
h3-nightly@2.0.0-20250424-160249-bea54cf

### Reproduction

After upgrading to the latest nightly version of H3 and integrating Vite with it, I noticed that the public directory assets are no longer being served (which should be handled by Vite). When requesting a file from the public directory (e.g., /my-file), the request hangs indefinitely and returns nothing.

Vite serves static assets from the public directory using the sirv package, which internally uses `fs.createReadStream(pathToFile).pipe(res)` to stream the file response.
src: https://github.com/lukeed/sirv/blob/d061616827dd32d53b61ec9530c9445c8f592620/packages/sirv/index.mjs#L92C5-L92C44

However, in h3@v2, this streaming approach (.pipe(res)) doesn't appear to work as expected — the request never completes.

**Reproduction**
Note: Please open the `/my-file` route in each demo to see the behavior difference.

Broken (H3 v2):
https://stackblitz.com/edit/github-n9w1dpxp-m15lbyau?file=app.js
Request to `/my-file` hangs.

Working (H3 v1):
https://stackblitz.com/edit/github-n9w1dpxp?file=app.ts
Request to `/my-file` successfully returns the file content.

### Describe the bug

The file should be streamed and served to the client when requested, just like in H3 v1, but piping to the response from a node middleware does not return anything, and the request stays open and returns nothing.

### Additional context

_No response_

### Logs

```sh

```

Narixius:
Fixes #1039 

By resolving the stream piped to the response in the `callNodeHandler` fn, the issue described in the mentioned issue will be fixed.

### DIFF
--- a/src/adapters.ts
+++ b/src/adapters.ts
@@ -86,7 +86,7 @@ function callNodeHandler(
   return new Promise((resolve, reject) => {
     res.once("close", () => resolve(kHandled));
     res.once("finish", () => resolve(kHandled));
-    res.once("pipe", () => resolve(kHandled));
+    res.once("pipe", (stream) => resolve(stream));
     res.once("error", (error) => reject(error));
     try {
       if (isMiddleware) {


pi0 (approved):
Thanks!

PR #1041, opened 2025-04-25
fix(setCookie): unique by `name`, `domain` and `path` only

### LINKED ISSUE #1037: Cookie merging inconsistency
### Environment

Node v22.14.0, h3@1.15.1 

### Reproduction

```ts
// Import h3 as npm dependency
import { createApp, createRouter, defineEventHandler, getResponseHeaders, setCookie } from "h3";

// Create an app instance

// Create a new router and register it in app
app.use(router);

// Add a new route that matches GET requests to / path
router.get(
  "/",
  defineEventHandler((event) => {
    setCookie(event, 'my-cookie', 'hello', {
        secure: true,
    })
    
    setCookie(event, 'my-cookie', 'goodbye', {
        secure: false,
    })

    console.log(getResponseHeaders(event)['set-cookie'])
    return { message: "⚡️ Tadaa!" };
  }),
);
```

### Describe the bug

When calling `setCookie` multiple times h3 will de-duplicate the `Set-Cookie` headers if they are for the same cookie. However this de-duplication doesn't exactly match how cookies behave in browsers. If two `setCookie` calls have different HttpOnly, Secure or SameSite attributes h3 will not de-duplicate and will send two Set-Cookie headers. However, the browser will only store one cookie.

For example the demo sets two cookies with the same name, path, domain and different secure flags. 

`getResponseHeaders(event)` looks like

```
{
  'set-cookie': [ 'my-cookie=hello; Path=/; Secure', 'my-cookie=goodbye; Path=/' ]
}
```

And the result in the browser cookie storage

Image

Merging less than necessary is better than merging too aggressively, but it would be nice if the behavior matched. 

### Additional context

https://httpwg.org/specs/rfc6265.html#sane-set-cookie-semantics

> If the user agent receives a new cookie with the same cookie-name, domain-value, and path-value as a cookie that it has already stored, the existing cookie is evicted and replaced with the new cookie.

### Logs

```sh

```

kricsleo:
resolves #1037

This PR uniquely identifies cookies by only three segments: "cookie-name, domain-value, and path-value" according to this spec: https://httpwg.org/specs/rfc6265.html#rfc.section.4.1.2

> If the user agent receives a new cookie with the same cookie-name, domain-value, and path-value as a cookie that it has already stored, the existing cookie is evicted and replaced with the new cookie. Notice that servers can delete cookies by sending the user agent a new cookie with an Expires attribute with a value in the past.

### DIFF
--- a/src/utils/cookie.ts
+++ b/src/utils/cookie.ts
@@ -95,6 +95,7 @@ export function deleteCookie(
     maxAge: 0,
   });
 }
+
 /**
  * Cookies are unique by "cookie-name, domain-value, and path-value".
  *
--- a/src/utils/cookie.ts
+++ b/src/utils/cookie.ts
@@ -95,14 +95,11 @@ export function deleteCookie(
     maxAge: 0,
   });
 }
-
+/**
+ * Cookies are unique by "cookie-name, domain-value, and path-value".
+ *
+ * @see https://httpwg.org/specs/rfc6265.html#rfc.section.4.1.2
+ */
 function _getDistinctCookieKey(name: string, options: Partial<SetCookie>) {
-  return [
-    name,
-    options.domain || "",
-    options.path || "/",
-    Boolean(options.secure),
-    Boolean(options.httpOnly),
-    Boolean(options.sameSite),
-  ].join(";");
+  return [name, options.domain || "", options.path || "/"].join(";");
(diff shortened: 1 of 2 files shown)


pi0 (approved):
Thanks!

Issue #504, opened 2023-08-10
Support for trusted proxy headers


harlan-zw:
### Describe the feature

As a module author, for maximum compatibility, it makes life easier to just use the proxy headers and fallback to node specifics. The issue is that these won't be safe in all environments and can be spoofed (i.e `x-forwarded-for`)

To solve this we need a simple implementation that:
1. allows users to provide a list of trusted proxies
2. allows module authors to generate the headers using this list of trusted proxies

I think combining 1 and 2 may be difficult but maybe you have some ideas on that. As a simple first implementation a `getRequestTrustedHeaders(event, trustedProxies)` may be enough.

Inspo: https://laravel.com/docs/master/requests#configuring-trusted-proxies

### Additional information

- [X] Would you be willing to help implement this feature?

pi0:
We have `getRequestHost({ xForwardedHost: true })`. Main benefit of using utils for each semi-standard header likt this is that we can support other vendor header as same name. How do you feel about introducing smaller utils like `getRequestIP` instead?

harlan-zw:
Not sure if this explanation is answering the issue about trusting easily spoofed headers. 

H3 should provide a small util function to only get the trusted values, similar to https://github.com/symfony/http-foundation/blob/6.3/Request.php#L783

It would then be on frameworks to allow configuring the trusted proxies and on modules to use that config.

pi0:
I am up to supporting a special event context property such as `event.context.isTrustedProxy` to automatically enable headers detection however it should be user responsibility solely to write logic. Based on server setup it is different how to determine trustibility.

harlan-zw:
> I am up to supporting a special event context property such as `event.context.isTrustedProxy` to automatically enable headers detection however it should be user responsibility solely to write logic. Based on server setup it is different how to determine trustibility.

Yes, I'm not concerned about users' usage, but about any integration code relying on these headers to work.

Consider this as an example: Nitro implements an in-built rate limiter, to effectively rate limit we need to know the trusted IP of the request. Nitro has no current way to know if it can trust the connecting IP, generically. This is because the x-forwarded-for header can easily be spoofed on some servers without proxies that use it.

Now blow this example up to Nuxt modules and all the other ways we need to know trusted IPs then you should have a better understanding of the problem. The solution requires multiple layers to solve, but as a starting point, we could introduce a way to filter for trusted headers similar to https://github.com/symfony/http-foundation/blob/6.3/Request.php#L2004.

[...]

PR #1046, opened 2025-05-03
feat(proxy): support `forwardHeaders` and `filterHeaders`

### LINKED ISSUE #709: proxyRequest does not proxy the Accept header
### Environment

Node 18.16.1

### Reproduction

N/A

### Describe the bug

Following https://github.com/unjs/h3/pull/646, proxyRequest ignores any `Accept` headers.

This is breaking my application (following a nuxt upgrade from 3.9 -> 3.11) as we proxy api requests via the server to a backend API, which requires us to send through an `Accept` header.

As a fix for the timebeing I can explicitly pass through the `Accept` header as follows

```
  return proxyRequest(event, targetPath, {
    headers: {
      Accept: event.headers.get('Accept') || ''
    }
  })
```

but this feels a bit unnecessary to me, I would have expected h3 to send through this header by default.

### Additional context

_No response_

### Logs

_No response_

### LINKED ISSUE #825: Host header is ignored when using proxyRequest()
### Environment

- Node version: v20.15.1
- H3: v1.12.0

### Reproduction

- Initialize a new Nuxt 3 app
- Add the following to `nuxt.config.ts`:

```typescript
runtimeConfig: {
    public: {
      backend: {
        url: "http://localhost:8000",
        host_header: "dev.localhost:8000",
      },
    },
  },
```

- Create an API route: `server/api/[...].ts`
- Use the following snippet:

```typescript
import { joinURL } from "ufo";

  const {
    public: { backend },
  } = useRuntimeConfig(event);

  const path: string = event.path.replace(/^\/api\//, "");
  const target: string = joinURL(backend.url, path);

  console.info(
    `Proxying request: (frontend) ${event.method} ${event.path} -> (backend) ${target}`,
  );

  return await proxyRequest(event, target, {
    headers: { "Host": backend.host_header },
  });
});
```

- Use the newly created API route, like `GET localhost:3000/api/users/`, and observe that in the receiving end (backend) the `Host` header is not replaced by the correct value.

### Describe the bug

When using the `proxyRequest` function in the latest Nuxt JS version, the `Host` header is ignored. From looking into the Github issues, a similar problem was found, as seen here: https://github.com/unjs/h3/issues/709 (in this case, the `Accept` header is ignored), introduced by this PR: https://github.com/unjs/h3/pull/646.

In the case of my project, we use the `Host` header to change the tenant domain on the backend side (one subdomain per tenant).

A culript piece of code was found here: https://github.com/unjs/h3/blob/17fc905d707f982c5b3e4c99b011a9fafc9b2834/src/utils/proxy.ts#L25-L33
where the `Host` header and others are explicitly being ignored.

If a fix is necessary, I can provide a PR with the code fix.

### Additional context

_No response_

### Logs

_No response_

kricsleo:
resolves #825
resolves #709

related: #504
related: https://github.com/nitrojs/nitro/issues/3160

TODO:
- [ ] Maybe [`opts.host` of `getProxyRequestHeaders`](https://github.com/h3js/h3/blob/f1115f3f577a1fef2ed2cf8f11db64d9760b5db6/src/utils/proxy.ts#L152) could be deprecated with this new option 🤔

### DIFF
--- a/src/utils/proxy.ts
+++ b/src/utils/proxy.ts
@@ -13,7 +13,7 @@ import { EmptyObject } from "./internal/obj.ts";
 
 export interface ProxyOptions {
   headers?: HeadersInit;
-  trustedHeaders?: string[];
+  forwardHeaders?: string[];
   fetchOptions?: RequestInit & { duplex?: "half" | "full" } & {
     ignoreResponseError?: boolean;
   };
@@ -52,7 +52,7 @@ export async function proxyRequest(
   const fetchHeaders = mergeHeaders(
     getProxyRequestHeaders(event, {
       host: target.startsWith("/"),
-      trustedHeaders: opts.trustedHeaders,
+      forwardHeaders: opts.forwardHeaders,
     }),
     opts.fetchOptions?.headers,
     opts.headers,
@@ -153,12 +153,12 @@ export async function proxy(
  */
 export function getProxyRequestHeaders(
   event: H3Event,
-  opts?: { host?: boolean; trustedHeaders?: string[] },
+  opts?: { host?: boolean; forwardHeaders?: string[] },
 ): Record<string, string> {
   const headers = new EmptyObject();
   for (const [name, value] of event.req.headers.entries()) {
     if (
-      opts?.trustedHeaders?.includes(name) ||
+      opts?.forwardHeaders?.includes(name) ||
       !ignoredHeaders.has(name) ||
       (name === "host" && opts?.host)
     ) {
--- a/src/utils/proxy.ts
+++ b/src/utils/proxy.ts
@@ -14,6 +14,7 @@ import { EmptyObject } from "./internal/obj.ts";
 export interface ProxyOptions {
   headers?: HeadersInit;
   forwardHeaders?: string[];
+  filterHeaders?: string[];
   fetchOptions?: RequestInit & { duplex?: "half" | "full" } & {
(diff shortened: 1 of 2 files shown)


src/utils/proxy.ts
```diff
@@ -49,7 +50,10 @@ export async function proxyRequest(
 
   // Headers
   const fetchHeaders = mergeHeaders(
-    getProxyRequestHeaders(event, { host: target.startsWith("/") }),
+    getProxyRequestHeaders(event, {
+      host: target.startsWith("/"),
+      trustedHeaders: opts.trustedHeaders,
```
pi0, line 55: I think we can call it `proxyHeaders` or `includeHeaders` (can be coupled with another `excludeHeader`). They are not trusted only being passed.

kricsleo: I once named it `forwardHeaders`; perhaps `includeHeaders` would be bettet, as you said

pi0: I like `forwardHeaders`!

kricsleo: Done 🙌

pi0:
While on it, do you mind to also add an option filterHeaders as counterpart (both can override built-in behavior)

kricsleo:
`filterHeaders` added.

pi0 (approved):
thnx ❤️

PR #1047, opened 2025-05-03
fix(sendStream): handle aborted (endless) streams correctly

### LINKED ISSUE #1045: Dangling SSE (Server Send Event) Streams
### Environment

"h3": "^1.14.0"

### Reproduction

1. Have endless SSE source
2. Proxy it with H3
3. Open proxied stream in browser (2nd route)
4. Close browser window
5. Proxied route still recieving data -> not terminated

```javascript
import { createApp, createRouter, eventHandler, sendStream } from "h3";
import { $fetch } from "ofetch";

app.use(router);

router.get(
  "/sse/stream",
  eventHandler(async (event) => {
    // replace with any (endless) sse source, I used pythons FastAPI + sse_starlette
    const response = await $fetch(
      "http://localhost:8000/rooms/9b3209ef-d675-4619-b510-57d5ed559915/queue/sse",
      {
        responseType: "stream",
        method: "GET",
        headers: {
          cookie: "participant_session=93bf571c-d1f5-470a-8bd7-f66d8e6ea9ac;",
        },
      },
    );
    event.node.res.setHeader(
      "Content-Type",
      "text/event-stream; charset=utf-8",
    );
    event.node.res.setHeader("Cache-Control", "no-cache");
    event.node.res.setHeader("Connection", "keep-alive");
    event.node.res.setHeader("X-Accel-Buffering", "no");
    return sendStream(event, response);
  }),
);

router.get(
  "/",
  eventHandler(async () => {
    return new Response(
      `<script>
      const es = new EventSource("/sse/stream");
      es.onmessage = (e) => {
        console.log(e.data);
      };
    </script>`,
      {
        headers: {
          "Content-Type": "text/html",
        },
      },
    );
  }),
);
```

### Describe the bug

I initially attempted to proxy my Python backend through a Nuxt.js application, which uses Nitro (and thereby h3). However, I encountered an issue where Server-Sent Events (SSE) connections were not being properly closed when the client disconnected.

After investigating, I suspect the root cause lies in the `sendStream` implementation. It appears that `sendStream` waits for the stream to end naturally but doesn't account for cases where the stream is endless and the client disconnects. As a result, data continues to be sent even though the client is no longer connected - essentially sending it into the void and keeping dangling connections.

### Additional context

_No response_

### Logs

```sh

```

JHolcman-T:
resolves #1045

Implements the fix proposed in #1045. Enables handling of aborted (endless) streams e.g. proxied SSE streams.

**Note:**
This is my first contribution, feel free to suggest improvements of any kind 😊

### DIFF
--- a/src/utils/response.ts
+++ b/src/utils/response.ts
@@ -414,6 +414,19 @@ export function sendStream(
     hasProp(stream, "pipeTo") &&
     typeof (stream as ReadableStream).pipeTo === "function"
   ) {
+    // instantiate a new AbortController to handle request aborts
+    const abortController = new AbortController();
+
+    // shared function to handle aborts
+    const doAbort = () => {
+      abortController.abort("Client closed or lost connection.");
+    }
+
+    // Handle request aborts
+    event.node.res.on("close", doAbort);
+    event.node.res.on("finish", doAbort);
+    event.node.res.on("error", doAbort);
+
     return (stream as ReadableStream)
       .pipeTo(
         new WritableStream({
@@ -421,6 +434,9 @@ export function sendStream(
             event.node.res.write(chunk);
           },
         }),
+        {
+          signal: abortController.signal
+        },
       )
       .then(() => {
         event.node.res.end();


pi0:
Thank you for PR, looks like a nice fix.

Would you be able to add one unit test?

JHolcman-T:
> Thank you for PR, looks like a nice fix.
> 
> Would you be able to add one unit test?

Sure, where would you expect this unittest to be located - I mean in which *.test.ts file?

JHolcman-T:
Hi @pi0

I came up with this test case. The only thing is I am not sure where to put it. It seems to me that no test file really fits :/ maybe `utils.test.ts` ?

```javascript
    it("can abort endless stream request", async () => {
      let connectionClosed = false;
      let streamCancelled = false;

      app.use(
        "/",
        eventHandler(async (event) => {
          event.node.res.setHeader(
            "Content-Type",
            "text/event-stream; charset=utf-8",
          );
          event.node.res.setHeader("Cache-Control", "no-cache");
          event.node.res.setHeader("Connection", "keep-alive");
          event.node.res.setHeader("X-Accel-Buffering", "no");

          // Detect when client disconnects
          event.node.res.on("close", () => {
            connectionClosed = true;
          });

          // clean up the stream
          let intervalId;
          const encoder = new TextEncoder();

          const stream = new ReadableStream({
            start(controller) {
              intervalId = setInterval(() => {
                controller.enqueue(encoder.encode("data: ping...\n\n"));
              }, 10);
            },
            // this is called when the stream has been cancelled trouhgh the AbortController signal in the sendStream function
            // commenting out the part of abort signal in the sendStream function => this will not be called after the request is cancelled
            cancel() {
              streamCancelled = true;
              // clean up the stream
              clearInterval(intervalId);
            },
          });

          return sendStream(event, stream);
        }),
      );

      const res = await fetch(url + "/", {
        method: "GET",
        headers: {
          Accept: "text/event-stream",
        },
      });

      if (!res.body) {
        throw new Error("No response body");
      } else {
        let messagesCounter = 0;
        const abort = new AbortController();
        res.body
          .pipeTo(
            new WritableStream({
              write(chunk) {
                messagesCounter += 1;
                if (messagesCounter > 5) {
                  // abort reading of the stream after few chunks
                  // simulate client disconnect
                  abort.abort();
                }
              },
            }),
            { signal: abort.signal },
          )
          // hanlde the error that is thrown when the stream is aborted
          .catch((err) => {});
      }
      await new Promise((resolve) => {
        setTimeout(() => {
          resolve(true);
        }, 500);
      });
      expect(res.status).toEqual(200);
      expect(connectionClosed !== false).toBe(true);
      expect(streamCancelled !== false).toBe(true);
    });
```

PR #1048, opened 2025-05-05
Update 1.request.md

ismailsabet:
Maybe we should consider removing getHeader and getHeaders entirely - no point in having two functions (`getHeader` and `getRequestHeader`) that do the same thing.

Makes it harder for people going through the documentation.

Perhaps we can do the same for all duplicated functions?
- getHeader and getRequestHeader
- setHeader and setResponseHeader
- defineEventHandler and eventHandler
- defineLazyEventHandler and lazyEventHandler ...

### DIFF
--- a/docs/2.utils/1.request.md
+++ b/docs/2.utils/1.request.md
@@ -33,7 +33,7 @@ Get a request header by name.
 
 ```ts
 export default defineEventHandler((event) => {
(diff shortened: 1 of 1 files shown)


pi0:
Thanks for PR. in v1, both are aliases. in v2, we deprecate both in favor of `event.req.headers.get`

PR #1065, opened 2025-05-30
feat: chainable middleware

pi0:
This PR replaces traditional middleware, which could be run one by one (or a new drafted version of routable middleware), with a much simpler and usable alternative. (Inspired by [srvx middleware](https://srvx.h3.dev/guide/middleware) and hono [`next()`](https://hono.dev/docs/api/context#res) function ❤️ )

The `next()` function allows access to the finally resolved (raw) response value.

Each middleware can either intercept by directly returning another value (before or after the final response) or returning same value of `next()` (which can be either raw response or a promise)

With changes of this PR, all middleware have access to matched route params (a long lasting issue)

```js
app.use((event, next) => {
  console.log(`Request: ${event}`)
  return next(); // return next() is optional
});

app.use(async (event, next) => {
  const value = await next(); // access to final "OK!" value 
  return value;
});

// Middleware can be filtered to only run on certain methods or route patterns
app.use(() => {}, { method: "GET", route: "/test/**" });

app.get("/**", (event) => {
  return "OK!"
});
```

(note, resolver tests removed since planning to remove also function in next PR. it was mainly legacy need for websockets)


pi0:
Top level hooks are more higher order but yes in many cases middleware can be used instead in a extendable way.

`onError` also handles `Response` generation and sending errors

`onBeforeResponse` accepts resolved `Response`

`onRequest` is kinda unnecessary now but can be for backward compat and little faster option. (might remove or deprecate later)

gulshan:
How does next behaves in case of errors?

pi0:
> How does next behave in case of errors?

It can throw sync or async errors, therefore can be used in async `try/catch` to gracefully handle of other middleware. If not, error (sync or async) bubble ups.

If you have time to try error handling behavior, would be nice to document and test it better 🙏🏼

PR #1068, opened 2025-05-30
feat(validate): standard schema support

### LINKED ISSUE #963: feat: Standard Schema validation

pi0:
Smaller rework of #964 (thanks @sandros94 ❤️ ) resolves #963 to support [standard schema](https://github.com/standard-schema/standard-schema) validators (zod, valibot, etc)

### DIFF
--- a/src/utils/internal/standard-schema.ts
+++ b/src/utils/internal/standard-schema.ts
@@ -1,0 +1,70 @@
+// https://github.com/standard-schema/standard-schema
+
+/** The Standard Schema interface. */
+export interface StandardSchemaV1<Input = unknown, Output = Input> {
+  /** The Standard Schema properties. */
+  readonly "~standard": Props<Input, Output>;
+}
+
+/** The Standard Schema properties interface. */
+export interface Props<Input = unknown, Output = Input> {
+  /** The version number of the standard. */
+  readonly version: 1;
+  /** The vendor name of the schema library. */
+  readonly vendor: string;
+  /** Validates unknown input values. */
+  readonly validate: (
+    value: unknown,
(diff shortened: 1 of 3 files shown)


sandros94:
@pi0 thank you so much for taking the time to restart this. I did try to fix the merge conflicts before but failed.

Unfortunately, while this will work with most generic schemas it will create typing issues with transformations during the validation (to fix this we have to use `InferOutput<SchemaType>` as a return for this).

Taking in example a simple query which contains a `page` and `size` for paginating content in the db, `/api/producs?page=3&size=20` we know that `3` and `20` will be considered strings. As such we can create a schema that would take this into account so that we both validate it and return the correct type:

```ts
const stringToNumber = z
  .string()
  .regex(/^-?(?:\d+(?:\.\d+)?|\.\d+)$/, 'Must be a number string')
  .transform(Number);

const querySchema = z.object({
  page: stringToNumber.optional(),
  size: stringToNumber.optional(),
});
```

Fastest implementation to fix this is to use function overloads, ~~will draft a PR if it is fine for you~~ https://github.com/h3js/h3/pull/1074

Issue #1060, opened 2025-05-21
`onError` hook is not called for nested H3 instances


gulshan:
### Environment

h3-nightly 2.0.0-20250428-102154-f1115f3
Node version: v22.15.1

### Reproduction

- Create a H3 instance with a `onError` hook.
- Nest it within another H3 instance.
- The `onError` hook is ignored.

### Describe the bug

`onError` hook is not called for nested H3 instances

### Logs

```sh

```

pi0:
please provide reproduction (code) 🙏🏼

gulshan:
Please find this reproduction-
```typescript
import { H3, type H3Event, toNodeHandler, withBase } from "h3-nightly";
import { createServer } from "http";

const nestedApp = new H3({
    onError(error: Error, _event: H3Event) {
        console.error('nestedApp error >>>', error);
    }
});

nestedApp
    .get('/', (_event) => 'hello')
    .get('/err', () => { throw new Error('nestedApp error') });

const app = new H3()
    .use('/**', withBase('/app', nestedApp));

const port = 3000
createServer(toNodeHandler(app))
    //                     ^^^
    // use `nestedApp` instead of `app` here to get the onError hook running
    .listen(port, () => console.log('started', `http://localhost:${port}`));
```

In this above code, the `onError` hook does not execute.
But if the `nestedApp` is run directly using the handler, the `onError` hook will execute.

@pi0

pi0:
Top-level hooks like `onRequest` and `onError` are called by the main app that controls the lifecycle.

Based on new chainable middlware, have added new `onError` util (#1083) that can be used to easily add middleware hooks that work regardless of using in app, sub-app or per event handler.

```js
import { H3, serve, onError, withBase } from "h3";

const nestedApp = new H3().use(
  onError((event, error) => {
    console.error(
      `[Nested App Error] [${event.req.method}] ${event.url.pathname} !! ${error.message}`,
    );
  }),
);

nestedApp
  .get("/", (_event) => "hello")
  .get("/err", () => {
    throw new Error("nestedApp error");
  });

const app = new H3().use("/**", withBase("/app", nestedApp));

serve(app);
```

gulshan:
Hi @pi0 ,
Is it possible to support changing the response in the `onError` middleware, like the `onError` hook?

gulshan:
[...]

PR #1069, opened 2025-05-31
fix(hooks): ensure nested `onError` hooks are called when necessary

kricsleo:
resolves https://github.com/h3js/h3/issues/1060

### DIFF
--- a/src/h3.ts
+++ b/src/h3.ts
@@ -14,6 +14,7 @@ import type {
   Middleware,
   MiddlewareOptions,
 } from "./types/handler.ts";
+import { createError } from "./error.ts";
 
 /**
  * Serve the h3 app, automatically handles current runtime behavior.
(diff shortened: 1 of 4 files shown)


pi0:
Thank you for the PR. I think it should be solved simpler without this changes to only inherit `onError` without modification to `h3.handler.

kricsleo:
Could you share more details with `only inherit onError` 🙏, I've tried several times but haven't been able to get it working in all of these situations:

- `app.get("/nested/**", nestedApp);`
- `app.get("/nested/**", withBase("/nested", nestedApp));`
- `app.use(nestedApp)`
- `app.use(withBase("/nested", nestedApp))`

Especially for `withBase("/nested", nestedApp)`, we can only access `nestedApp.config.onError` within the `h3.handler`(`h3.handler = h3.handler.bind(this)`）.

pi0:
(putting to draft to explain better later today, likely we can advocate use of new middleware of fine gained error handling at different levels)

pi0:
Hi. Please see my comment in https://github.com/h3js/h3/issues/1060#issuecomment-2935481752

(we probably need to better document it) 

Inheriting hooks might require additional complexity for each, while middleware always works.

kricsleo:
Yeah, with the new middleware design, we can handle this better.

What do you think about the tests for the nested `onError` handler in this PR? Are they needed? ❤️

PR #1074, opened 2025-06-02
fix(validation): standard-schema return type

sandros94:
Reported in https://github.com/h3js/h3/pull/1068#issuecomment-2931648390

This solves the issue when a standard-schema compatible library both validates and transform the data. For example as demonstrated in the use-case of the above comment:

> Taking in example a simple query which contains a `page` and `size` for paginating content in the db, `/api/producs?page=3&size=20` we know that `3` and `20` will be considered strings. As such we can create a schema that would take this into account so that we both validate it and return the correct type:
>
> ```ts
> const stringToNumber = z
>   .string()
>   .regex(/^-?(?:\d+(?:\.\d+)?|\.\d+)$/, 'Must be a number string')
>   .transform(Number);
>
> const querySchema = z.object({
>   page: stringToNumber.optional(),
>   size: stringToNumber.optional(),
> });
> ```

### DIFF
--- a/src/utils/body.ts
+++ b/src/utils/body.ts
@@ -1,7 +1,11 @@
 import type { H3Event } from "../types/event.ts";
 import type { InferEventInput } from "../types/handler.ts";
 import { createError } from "../error.ts";
-import { validateData, type ValidateFunction } from "./internal/validate.ts";
+import { validateData, type ValidateResult } from "./internal/validate.ts";
+import type {
+  StandardSchemaV1,
+  InferOutput,
+} from "./internal/standard-schema.ts";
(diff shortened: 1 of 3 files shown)


sandros94:
Not the prettiest implementation tbh 😅

pi0 (approved):
Thank you ❤️

pi0:
@sandros94 if you had a chance can you please help updating https://h3.dev/examples/validate-data section? 🙏🏼

sandros94:
> @sandros94 if you had a chance can you please help updating https://h3.dev/examples/validate-data section? 🙏🏼

Yes sure, will draft something by tomorrow

PR #924, opened 2024-12-02
docs(request): `getValidated*` utils are async

danielrentz:
The functions `getValidatedQuery`and `getValidatedRouterParams` are async. This PR fixes the JSDoc examples.

PR #867, opened 2024-08-18
docs: add missing `parse` suffix for validation utils

romanhrynevych:
This PR fix #866

Issue #866, opened 2024-08-18
Missing `parse` suffix for validation utils


romanhrynevych:
### Describe the change

I was trying to use [example](https://h3.unjs.io/utils/request#getvalidatedqueryevent-validate) of `getValidatedQuery` to get queries but receive type error in vs code editor:

<img width="1041" alt="image" src="https://github.com/user-attachments/assets/640c51bc-f18e-44d6-963d-476ef1b89361">

I find out that `parse` suffix was missing there, but it was in this part of docs: 
<img width="999" alt="image" src="https://github.com/user-attachments/assets/4a0edded-85e0-4061-aafb-c58f25358ad8">

### URLs

https://h3.unjs.io/examples/validate-data#utilities-for-validation

### Additional information

- [X] Would you be willing to help?

PR #1080, opened 2025-06-03
docs(validation): rewrite data validation page

sandros94:
Followup for https://github.com/h3js/h3/pull/1074#issuecomment-2932520771

A full rewrite of https://h3.dev/examples/validate-data while also going a bit more in details.

### DIFF
--- a/docs/4.examples/validate-data.md
+++ b/docs/4.examples/validate-data.md
@@ -6,10 +6,10 @@ icon: ph:arrow-right
 
 > Ensure that your data are valid and safe before processing them.
 
-When you receive data from user on your server, you must validate them. By validate, we mean that the shape of the received data must match the expected shape. It's important because you can't trust user input.
+When you receive data on your server, you must validate them. By validate, we mean that the shape of the received data must match the expected shape. It's important because you can't trust the data coming from unknown sources, like a user or an external API.
 
 > [!WARNING]
-> Do not use a generic as a validation. Providing an interface to a utility like `readJSONBody` is not a validation. You must validate the data before using them.
+> Do not use type generics as a validation. Providing an interface to a utility like `readBody` is not a validation. You must validate the data before using it.
 
 ## Utilities for Validation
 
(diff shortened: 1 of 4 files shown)


simon-kolkmann:
@sandros94 This PR introduced a regression of https://github.com/h3js/h3/issues/866 by removing the `.parse` suffix when showcasing a zod validator. The documented way of using e.g. `getValidatedQuery` with zod now produces a typescript error.

sandros94:
> @sandros94 This PR introduced a regression of https://github.com/h3js/h3/issues/866 by removing the `.parse` suffix when showcasing a zod validator. The documented way of using e.g. `getValidatedQuery` with zod now produces a typescript error.

I will double check this shortly, could you share the type error you are seeing in the meantime?

Keep in mind that all h3 v2 beta do support the [standard-schema spec](https://github.com/standard-schema/standard-schema)

sandros94:
@simon-kolkmann I'm not able to reproduce, using any standard-schema compatible zod relase (`^3.24.0`) and `h3@2.0.0-beta.3`

Feel free to reply here or even better tag me while opening a new issue with a small reproduction 🙏

simon-kolkmann:
@sandros94 Thanks for the quick response! I guess this is only a problem of the v1 docs then. My nuxt app uses `h3@1.15.3` and `zod@3.24.3`. [The second example](https://v1.h3.dev/utils/request#getvalidatedqueryevent-validate) fails with

```
Argument of type 'ZodObject<{ key: ZodString; }, "strip", ZodTypeAny, { key: string; }, { key: string; }>' is not assignable to parameter of type 'ValidateFunction<QueryObject | undefined>'.
  Type 'ZodObject<{ key: ZodString; }, "strip", ZodTypeAny, { key: string; }, { key: string; }>' provides no match for the signature '(data: unknown): ValidateResult<QueryObject | undefined> | Promise<ValidateResult<QueryObject | undefined>>'.
```

I just now recognized that the example is also missing `async...await` which was added in https://github.com/h3js/h3/pull/924. 

Are the v1 docs still maintained or is the focus solely on v2? If reasonable, a backport of https://github.com/h3js/h3/pull/867 and https://github.com/h3js/h3/pull/924 to the v1 branch would probably help people like me who are still using v1.

pi0:
Feel free to open PR for improving v1 docs

sandros94:
> Feel free to open PR for improving v1 docs

@pi0 are the jsdocs shared between v1 and v2? Indeed the v1 shows the wrong example for the automd generated ones, while the correct one in [the related example page](https://v1.h3.dev/examples/validate-data) 🤔

pi0:
no jsdocs are specific to each major/branch.

PR #1086, opened 2025-06-04
feat(defineEventHandler): support `.fetch` interface

pi0:
Reuse existing composable internals to enable event handlers defined with `defineEventHandler` directly be a usable `.fetch`-able interface!

Thanks to initial inspiration from @productdevbook since he was asking how to just use the (value => Response) part of h3 ❤️ 

```js
import { defineEventHandler } from "h3"; // 5kB (2kB gzipped)

const log = (event) => {
  console.log(`[${event.req.method}] ${event.req.url}`);
};

// ✔ Inferred Types: (event: H3Event<EventHandlerRequest>) => { hello: string; }
const handler = defineEventHandler({
  // ✔ Request middleware
  middleware: [log],
  // ✔ Directly return a value or throw an error
  handler: () => ({ hello: "world" }),
});

// ✔ compatible with .fetch API
// Use for teesting or integrate into any compatible framework and runtime!
// [GET] http://localhost/test
// { hello: "world" }
console.log(await handler.fetch("/test").then((res) => res.json()));
```

### DIFF
--- a/examples/handler-fetch.mjs
+++ b/examples/handler-fetch.mjs
@@ -1,0 +1,19 @@
+import { defineEventHandler } from "h3"; // 5kB (2kB gzipped)
+
+const log = (event) => {
+  console.log(`[${event.req.method}] ${event.req.url}`);
+};
+
+// ✔ Inferred Types: (event: H3Event<EventHandlerRequest>) => { hello: string; }
+const handler = defineEventHandler({
+  // ✔ Request middleware
+  middleware: [log],
+  // ✔ Directly return a value or throw an error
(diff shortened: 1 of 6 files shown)


productdevbook:
if we put the middleware name as `modules` and consider every middleware as a `module`, there may be a different perspective.

productdevbook:
```ts
// ✔ Inferred Types: (event: H3Event<EventHandlerRequest>) => { hello: string; }
const handler = defineEventHandler({
  // ✔ Request middleware
  module: [log],
  // ✔ Directly return a value or throw an error
  handler: () => ({ hello: "world" }),
});
```

PR #1090, opened 2025-06-05
feat: `HTTPError`

### LINKED ISSUE #767: Deprecate error  `statusMessage` / `statusCode` (node) in favor of `status`/`statusText` (web)
[`status`](https://developer.mozilla.org/en-US/docs/Web/API/Response/status) and [`statusText`](https://developer.mozilla.org/en-US/docs/Web/API/Response/statusText) are Web Response props.

### LINKED ISSUE #768: Allow arbitrary additional props to be passed to `createError`
### Describe the feature

I want to send an additional prop in my error responses. For example `code: "invitation_not_found"` and I don't want it to be under data, but in the root error object.

### Additional information

pi0:
This PR replaces `H3Error` with the new `HTTPError` class.

Changes:
- `H3Error` renamed to `HTTPError` (`H3Error` export exists)
- New `HTTPError.isError` util for safe instance check
- Internal implementation improvement for proper inheritance
- New signutures `new HTTPError(details)`, `HTTPError.code(code, details)` and `new HTTPError(message, details)` for more ergonomic usage.
- `statusCode` and `statusMessage` deprecated in favor of `status` and `statusText` (resolves #767)
- `{ statusCode, statusMessage }` changed to `{ status, statusText }` in default error handler responses.
- `.fatal` flag removed as barely used (can still be manually attached)
- `.unhandled` is `false` by default and only explicitly enabled for errors user did not handle or did not throw using `new HttpError`
- Add new `{ body }` key to allow overriding top-level JSON keys (resolves #768)
- For consistency between dev and prod, in unhandled errors always message, data and body are hidden from JSON responses.

Examples:

```js
// HttpError.status
HTTPError.status(404)
HTTPError.status(418, "I'm a teapot")
HTTPError.status(403, "Forbidden", { message: "Not authenticated" })

// new HttpError("message", details?)
new HTTPError("This is an error", { status: 400, cause: error })

// new HttpError(details)
new HTTPError({ message: "This is an error", status: 500, statusText: "Not Found", data: {} })
```

### DIFF
--- a/test/hooks.test.ts
+++ b/test/hooks.test.ts
@@ -72,8 +72,6 @@ describeMatrix("hooks", (t, { it, expect }) => {
     expect(t.hooks.onRequest).toHaveBeenCalledTimes(1);
     expect(t.hooks.onRequest.mock.calls[0][0].path).toBe("/foo");
 
-    console.log(t.hooks.onError.mock.calls[0]![0]!);
-
     expect(t.hooks.onError).toHaveBeenCalledTimes(1);
     expect(t.hooks.onError.mock.calls[0]![0]!.status).toBe(500);
     expect(t.hooks.onError.mock.calls[0]![0]!.cause).toBeInstanceOf(TypeError);


OskarLebuda:
I'm a bit concerned about the ergonomics and potential confusion caused by the multiple `HttpError` constructor signatures. Currently, it seems like we can do:

```ts
new HttpError("This is an error", { status: 400 });
new HttpError(400, { message: "This is an error" });
new HttpError({ status: 404, statusText: "Not Found" });
```

While this provides flexibility, it also opens the door to ambiguity and inconsistent usage across codebases. For example:

```ts
// Which one is correct?
new HttpError("Something went wrong", { status: 500 });
new HttpError(500, { message: "Something went wrong" });
```

Both achieve the same thing, but introduce cognitive load. Developers might wonder: *Should I pass status first, or message?* This can especially be confusing in large teams or public APIs.

#### Possible alternatives:

* Stick to a single, consistent constructor pattern.
* Or introduce named static constructors for clarity, e.g.:

```ts
HttpError.fromStatus(404, "Not Found");
HttpError.fromMessage("Unauthorized", { status: 401 });
```

This could improve readability and avoid overload hell, especially in TypeScript projects where autocomplete may offer unclear suggestions.

What do you think?

pi0:
Thanks for feedback @OskarLebuda. I like the static idea better than constructor overloads and increases readability!

I have added `HttpError.status(code)` support.

We need to support `new HttpError(string)` signature only to be compatible with `Error` constructor.

OskarLebuda:
> Thanks for feedback @OskarLebuda. I like the static idea better than constructor overloads and increases readability!
> 
> I have added `HttpError.status(code)` support.
> 
> We need to support `new HttpError(string)` signature only to be compatible with `Error` constructor.

Just looked at the changes - and IMO they look significantly better! 😍  Thanks 🙏🏻

atinux:
Great feedback @OskarLebuda, I really like the new changes!

src/error.ts
```diff
@@ -1,168 +1,242 @@
...
-  }
+    const status = sanitizeStatusCode(
+      (details as ErrorBody)?.status ||
+        (details?.cause as ErrorBody)?.status ||
+        (details as ErrorBody)?.status ||
+        (details as ErrorInput)?.statusCode,
+      500,
+    );
+
+    const stautText = sanitizeStatusMessage(
```
septatrix, line 180: Typo?

septatrix (commented):
Will the new `HTTPError` automatically choose the correct statusText based on the status code or do they still fall back to "Server Error" (which is just wrong) or should I file a new bug for that?

src/error.ts
pi0: yep, it is internal however please feel free to make PR to fix!

pi0: > Will the new HTTPError automatically choose the correct statusText 

HTTP clients (browser, or curl for example) do automatically show default status text.

Also note that in modern protocol http2 does not has concept of status text. (we need to document this!)

septatrix: > HTTP clients (browser, or curl for example) do automatically show default status text.

No, they show the one provided by the server which for Nuxt always seems to be "Server Error" unless explicitly overridden:

```sh
curl --show-headers -X POST http://localhost:3000/api/entities
HTTP/1.1 401 Server Error
...
{
  "error": true,
  "url": "http://localhost:3000/api/entities",
  "statusCode": 401,
  "statusMessage": "Server Error",
  "message": "Unauthorized",
  "stack": [
    "Unauthorized",
    "at createError (/home/septatrix/Documents/Simply5/programming/participations/node_modules/.pnpm/h3@1.15.3/node_modules/h3/dist/index.mjs:71:15)",
    "at requireUserSession (/home/septatrix/Documents/Simply5/programming/participations/node_modules/.pnpm/nuxt-auth-utils@0.5.20_magicast@0.3.5/node_modules/nuxt-auth-utils/dist/runtime/server/utils/session.js:34:13)",
    "at async requireRole (/home/septatrix/Documents/Simply5/programming/participations/server/utils/auth.ts:8:1)",
    "at async Object.handler (/home/septatrix/Documents/Simply5/programming/participations/server/api/entities/index.post.ts:6:1)",
    "at async file:///home/septatrix/Documents/Simply5/programming/participations/node_modules/.pnpm/h3@1.15.3/node_modules/h3/dist/index.mjs:2003:19)",
    "at async Object.callAsync (/home/septatrix/Documents/Simply5/programming/participations/node_modules/.pnpm/unctx@2.4.1/node_modules/unctx/dist/index.mjs:72:16)",
    "at async Server.toNodeHandle (/home/septatrix/Documents/Simply5/programming/participations/node_modules/.pnpm/h3@1.15.3/node_modules/h3/dist/index.mjs:2295:7)"
  ]
}
```

pi0: H3 error by default does not add status text, and probably layers on top (Nitro=>Nuxt) should also avoid overriding it too, to allow default status being inferred (also, I emphasize statusText in modern http2 connections is completely removed, so it is not reliable)

image


```js
const {serve,H3, HTTPError}=require('h3');const app=new H3().get("/", () => { throw HTTPError.status(403) });serve(app)
```

septatrix: > H3 error by default does not add status text, and probably layers on top (Nitro=>Nuxt) should also avoid overriding it too

Okay, I will investigate and report in the other projects if I find anything

PR #1092, opened 2025-06-07
refactor(serveStatic): check mtime before etag

markthree:
I think we should check mtime first and then ETag. 

Because when generating ETag, we always need to read the file content and perform heavy CPU computations. 
In contrast, checking mtime only requires accessing the file's modification time metadata.

There are no breaking changes - just a logical order adjustment. 
It's a small change, but it dramatically improves service performance.

### DIFF
--- a/src/utils/static.ts
+++ b/src/utils/static.ts
@@ -117,18 +117,6 @@ export async function serveStatic(
     throw new HTTPError({ statusCode: 404 });
   }
 
-  if (meta.etag && !event.res.headers.has("etag")) {
-    event.res.headers.set("etag", meta.etag);
-  }
-
-  const ifNotMatch =
-    meta.etag && event.req.headers.get("if-none-match") === meta.etag;
-  if (ifNotMatch) {
-    event.res.status = 304;
-    event.res.statusText = "Not Modified";
-    return "";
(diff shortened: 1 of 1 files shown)


pi0:
This is nice but we **always** resolve meta first in L105. How this can improve perf?

pi0:
Only reason it can be faster would be implementation of `getMeta` uses a lazy-getter on etag and in case etag generation is on demand. Nitro for exmple pre-calculated meta and inline them in server bundler.

pi0 (approved):
Thnx, merging as refactor since it _can_ possibly be used as perf improvement by implementation.

markthree:
> This is nice but we **always** resolve meta first in L105. How this can improve perf?

You're right. I missed the logic. Maybe this place could be on-demand

markthree:
> Only reason it can be faster would be implementation of `getMeta` uses a lazy-getter on etag and in case etag generation is on demand. Nitro for exmple pre-calculated meta and inline them in server bundler.

Yes, the nitrojs precomputation is very smart. But I've observed that if users only use h3, they might not think of it or be lazy and want less code (yes, I'm one of them 🐶)

PR #1096, opened 2025-06-08
refactor: only accept `request` for `H3.fetch`

pi0:
[Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) signuture is `(req, init?)`; however, no other runtimes agree on this, yet called it `.fetch`...

- Bun: `fetch(request, server)`
- Deno: `fetch(request, info)`
- Cloudflare: `fetch(request, env, ctx)`

It might be safer to have `h3.fetch` strictly only accept a single `(request)`. This also helps avoid wrongly added overhead if one does `export default app.fetch` (which causes recreating `Request` with a second irrelevant arg considered RequestInit)

Downside, it takes away the easy DX of doing `app.fetch("url", { headers: {} })` :(

### DIFF
--- a/src/h3.ts
+++ b/src/h3.ts
@@ -36,12 +36,9 @@ export const H3 = /* @__PURE__ */ (() => {
       this.handler = this.handler.bind(this);
     }
 
-    fetch(
-      request: ServerRequest | URL | string,
-      options?: RequestInit,
-    ): Promise<Response> {
+    fetch(request: ServerRequest | URL | string): Promise<Response> {
       try {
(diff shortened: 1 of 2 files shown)


pi0:
Canceling the idea for now. Not worth killing DX only to prevent shooting in the foot implied by runtimes.

pi0:
Reopening as it can be also a fatal issue for integrations, e.g if second arg uses an overlapping key like `cache` it can break fetch implementations.

PR #1103, opened 2025-06-11
feat: optional route meta

productdevbook:
we have to send all the incoming options. the user may want to add something special and use it later.


pi0:
We shall expose new .meta optional prop

This is not a fix but a new feature. Please consider first opening an issue all the times 🙏

pi0:
~> #1118

PR #1112, opened 2025-06-19
optional peer dependency on fetchdts

sirenkovladd:
image

move the 101kb dependency to devDependency that is used only for the type

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,6 @@
   },
   "dependencies": {
     "cookie-es": "^2.0.0",
-    "fetchdts": "^0.1.4",
     "rou3": "^0.6.3",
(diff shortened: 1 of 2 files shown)


pi0:
I appreciate the PR, but type support is a _feature_ of h3 (we just moved some types from internal to an external package)

It could be an optional peer dependency but considering the negative DX (of having to install an additional package only for types), internal complexitiy (we need conditional type checking to see if dependency exists or not), and additional few kbs is probably not worth it. total h3 install size is still relatively small (almost half of [hono](https://packagephobia.com/result?p=hono) for one example..)

Once fetchdts was more stable, we could rebundle types as an alternative.

Issue #1117, opened 2025-06-20
readMultipartFormData in v2


huseeiin:
### Describe the feature

some projects that are still on v1 use `readMultipartFormData`
for backwards compatibility, this can be implemented in h3 v2

### Additional information

- [ ] Would you be willing to help implement this feature?

PR #1120, opened 2025-06-21
feat: add `readMultipartFormData` for backward compatibility

kricsleo:
resolves https://github.com/h3js/h3/issues/1117

### DIFF
--- a/src/_deprecated.ts
+++ b/src/_deprecated.ts
@@ -99,13 +99,13 @@ export const readFormData: (event: H3Event) => Promise<FormData> =
 /** @deprecated Please use `event.req.formData()` */
 export async function readMultipartFormData(event: H3Event): Promise<
   Array<{
-    data: Buffer;
+    data: Uint8Array;
     name?: string;
     filename?: string;
     type?: string;
   }>
 > {
-  const formData = await readFormDataBody(event);
+  const formData = await event.req.formData();
 
   return Promise.all(
     [...formData.entries()].map(async ([key, value]) => {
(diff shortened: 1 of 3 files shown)


pi0:
Can you add a test (in matrix) for multi part?

kricsleo:
Tests ported from H3 v1 😁

src/_deprecated.ts
```diff
@@ -96,6 +96,32 @@ export async function readFormDataBody(event: H3Event): Promise<FormData> {
...
+/** @deprecated Please use `event.req.formData()` */
+export async function readMultipartFormData(event: H3Event): Promise<
+  Array<{
+    data: Buffer;
+    name?: string;
+    filename?: string;
+    type?: string;
+  }>
+> {
+  const formData = await readFormDataBody(event);
```
pi0: We could directly use `event.req.formData()` here.

src/_deprecated.ts
```diff
@@ -96,6 +96,32 @@ export async function readFormDataBody(event: H3Event): Promise<FormData> {
...
+
+  return Promise.all(
+    [...formData.entries()].map(async ([key, value]) => {
+      return value instanceof Blob
+        ? {
+            name: key,
+            type: value.type,
+            filename: value.name,
+            // Convert Blob to Buffer for backward compatibility
+            data: Buffer.from(await value.arrayBuffer()),
```
pi0: We could convert the value to a `Uint8Array` instead and avoid adding Node.js runtime dependency (Buffer is a superclass)

src/_deprecated.ts
```diff
@@ -96,6 +96,32 @@ export async function readFormDataBody(event: H3Event): Promise<FormData> {
...
+          }
+        : { name: key, data: Buffer.from(value) };
```
pi0: Do you know what is value type in other case? We should ideally avoid depending on `Buffer` and Node.js APIs even in backward compatible utils.

src/_deprecated.ts
kricsleo: Done.

src/_deprecated.ts
kricsleo: Done.

pi0 (commented):
Thank you ❤️

src/_deprecated.ts
kricsleo: Previously, this was just `Buffer`, but now, as you pointed out, we can all use `Unit8Array`.

PR #1130, opened 2025-06-27
feat: support typed `req.headers`

danielroe:
this adds type support for `event.req.headers`

CleanShot 2025-06-27 at 10 41 22@2x

### DIFF
--- a/src/adapters.ts
+++ b/src/adapters.ts
@@ -9,7 +9,11 @@ import type {
 } from "srvx";
 import type { H3 } from "./h3.ts";
 import type { H3EventContext } from "./types/context.ts";
-import type { EventHandler, EventHandlerResponse } from "./types/handler.ts";
+import type {
+  EventHandler,
+  EventHandlerResponse,
+  TypedServerRequest,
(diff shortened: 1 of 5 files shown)


pi0:
Auto-complete on request headers is really awesome, but it seems it will break type-compatibility with `Request`/`Headers`, requiring manual type casting from users.

Currently, both `ServerRequest` and `TypedServerRequest` are type-compatible with `Request`. This allows natively using any web-standard typed util with h3 event.

```js
import { type EventHandler, defineHandler } from "h3";

function webUtil(req: Request) {}
function webUtil2(headers: Headers) {}

defineHandler((event) => {
  // Both will fail 
  webUtil(event.req);
  webUtil2(event.req.headers);
});


const event: H3Event
webUtil(event.req);
webUtil2(event.req.headers);
```

---

It looks incompatibility comes from header values, can we limit it to keys to allow both auto-complete for keys and also type compat with web types?

danielroe:
we can indeed - nice idea!

pi0 (approved):
💅🏼

PR #1137, opened 2025-07-01
feat(serveStatic): support default and custom mimes

### LINKED ISSUE #1136: add default mimes for `serveStatic`
Discussed in https://github.com/orgs/h3js/discussions/1135

H3 built-in `serveStatic` utility can support mimes for common file extensions to avoid sending them as `application/octet-stream.`

We cannot depend on external packages to keep package size minimal but a list of common extensions can be aded.

We probably also need to support a new option to allow custom mime resolution with more extensions.

outslept:
Taken the types from https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/MIME_types/Common_types (excluding not so used like ms office related)

Closes #1136


src/utils/static.ts
```diff
@@ -2,6 +2,39 @@ import type { H3Event } from "../event.ts";
 import { HTTPError } from "../error.ts";
 import { withLeadingSlash, withoutTrailingSlash } from "./internal/path.ts";
 
+const COMMON_MIME_TYPES: Record<string, string> = {
```
pi0: Can you please refactor it to `internal/mimes.ts`?

src/utils/static.ts
```diff
@@ -2,6 +2,39 @@ import type { H3Event } from "../event.ts";
...
+  ".webm": "video/webm",
+
+  ".zip": "application/zip",
+
+  ".pdf": "application/pdf",
+};
+
+function getMimeType(path: string): string {
+  const ext = path.slice(Math.max(0, path.lastIndexOf("."))).toLowerCase();
+  return COMMON_MIME_TYPES[ext] || "application/octet-stream";
```
pi0: Not having a default MIME value fallback is probably better.

src/utils/static.ts
```diff
@@ -2,6 +2,39 @@ import type { H3Event } from "../event.ts";
...
+  ".mp4": "video/mp4",
+  ".webm": "video/webm",
+
+  ".zip": "application/zip",
+
+  ".pdf": "application/pdf",
+};
+
+function getMimeType(path: string): string {
+  const ext = path.slice(Math.max(0, path.lastIndexOf("."))).toLowerCase();
```
pi0: It would probably be safer if we explicitly handle `indexOf === -1` situation and early return `undefined`

outslept:
Everything should be addressed now 👀

src/utils/static.ts
```diff
@@ -1,6 +1,16 @@
 import type { H3Event } from "../event.ts";
 import { HTTPError } from "../error.ts";
 import { withLeadingSlash, withoutTrailingSlash } from "./internal/path.ts";
+import { COMMON_MIME_TYPES } from "./internal/mimes.ts";
+
+function getMimeType(path: string): string | undefined {
```
pi0: We could move this function also to internal/mime.ts and use export to add unit tests

outslept: Sure, give me a minute

src/utils/static.ts
```diff
@@ -1,6 +1,16 @@
...
+  const dotIndex = path.lastIndexOf(".");
```
pi0: Here i think we might have an issue where we might check dot before last slash like /foo/bar.txt/baz/ 


(solving it is important only for next comment to pass ext to user callback)

src/utils/static.ts
```diff
@@ -51,6 +61,11 @@ export interface ServeStaticOptions {
    * When set to true, the function will not throw 404 error when the asset meta is not found or meta validation failed
    */
   fallthrough?: boolean;
+
+  /**
+   * Custom MIME type resolver function
+   */
+  getMimeType?: (path: string) => string | undefined;
```
pi0: I was thinking maybe it is easier if we give (ext, url) to callback? (then users don’t have to write ext parser themselves only add custom mime mapping.

src/utils/static.ts
outslept: Good catch, missed this

PR #1143, opened 2025-07-02
feat: experimental `defineRoute`

### LINKED ISSUE #1088: defineRoute
### Describe the feature

In CLI and codebase scanning architectures, type inference and code merging often create inconsistencies with methods like .get, .post, etc., leading to a different set of challenges in every project. To prevent this and achieve a strictly consistent system, introducing a new structured approach unlocks many bottlenecks. In the modern ecosystem, support for StandardSchema significantly simplifies module development—even in specialized modules like OpenAPI and others.

```ts
import { H3, defineEventHandler, defineRoute, readBody, readValidatedBody } from "h3";
import { z } from "zod";

export const book = defineRoute({
  method: "GET",
  route: "/api/books",
  routerParams: z.object({
    id: z.string(),
  }),
  queryParams: z.object({
    page: z.string().optional(),
    limit: z.string().optional(),
  }),
  input: z.object({ 
    title: z.string(),
  }),
  output: z.object({
    message: z.string(),
    name: z.string(),
  }),
  handler: async (event)=> {
    const data = await readBody(event);
    const data2 = await readValidatedBody(event)

    return {
      message: "Book list",
      name: ''
    }
  }
});


const app = new H3();

app.addRoute(book);
```

### Additional information

productdevbook:
cc #1088


src/utils/route.ts
```diff
@@ -0,0 +1,70 @@
...
+ * }))
+ * ```
+ */
+export function defineRoute(
+  config: Omit<RouteDefinition, "method" | "route"> & {
+    method?: HTTPMethod;
+    route?: string;
+  },
+): EventHandler {
+  const handler = defineHandler({
```
pi0: We could reuse `defineValidatedHandler` here

src/utils/route.ts
```diff
@@ -0,0 +1,70 @@
...
+ *   middleware: [ ... ]
+ * }))
+ * ```
+ */
+export function defineRoute(
+  config: Omit<RouteDefinition, "method" | "route"> & {
+    method?: HTTPMethod;
+    route?: string;
+  },
+): EventHandler {
```
pi0: Returned value should be **a plugin** that adds a route; otherwise, this utility is no different from the `defineValidatedhandler` utility.

https://github.com/h3js/h3/issues/1088#issuecomment-2957019391

docs/4.examples/define-route.md
```diff
@@ -0,0 +1,331 @@
...
+## Why defineRoute?
+
+In modern applications, you often need:
+
+- **Type safety** for route parameters and request bodies
+- **Consistent validation** across all endpoints
+- **Route metadata** for authentication, caching, rate limiting
+- **Middleware integration** for cross-cutting concerns
+- **Better developer experience** with IntelliSense and error checking
+- **Seamless integration** with schema libraries like Zod, Valibot, etc.
```
pi0: All of above are achieved with normal `defineValidated` utility. `defineRoute` was proposed to allow modular/encapsulated registration.

docs/4.examples/define-route.md
```diff
@@ -0,0 +1,331 @@
...
+const route = defineRoute({
+  method: "GET",
+  route: "/api/users",
+  handler: () => ({ users: [] }),
+});
+
+app.on(route.method, route.route, route.handler, route.options);
+```
+
+### Method 2: createService (alias)
```
pi0: We probably shouldnt introduce more aliases to keep exports simple.

src/utils/route.ts
```diff
@@ -0,0 +1,127 @@
...
+    // Determine which schemas to use for validation based on config
+    const bodySchema =
+      config.input && validationConfig.input ? config.input : undefined;
+    const querySchema =
+      config.queryParams && validationConfig.queryParams
+        ? config.queryParams
+        : undefined;
+
+    // Create handler with conditional validation using defineValidatedHandler
+    const handler: EventHandler =
```
OskarLebuda: Instead of repeating `defineValidatedHandler` twice, we create `baseHandler` once and conditionally spread body/query schemas if needed.
Also dropped the as unknown as EventHandler - one cast is enough.
Cleaner and easier to follow :) 

```typescript
// Determine if validation for body or query is needed
const needsValidation = bodySchema || querySchema;

// Create the common base handler with optional routerParams validation
const baseHandler = createBaseHandler(async (event: H3Event) => {
  if (config.routerParams && validationConfig.routerParams) {
    await getValidatedRouterParams(event, config.routerParams);
  }

  return await config.handler(event);
});

// Construct the validated handler dynamically based on the presence of schemas
const validatedHandler = defineValidatedHandler({
  middleware: config.middleware,
  ...(needsValidation && { body: bodySchema }),
  ...(needsValidation && { query: querySchema }),
  handler: baseHandler,
}) as EventHandler;

// Final handler export
const handler: EventHandler = validatedHandler;
```

src/utils/route.ts
```diff
@@ -0,0 +1,127 @@
...
+      routerParams: true,
+      queryParams: true,
+      input: true,
+      output: true,
+      ...config.validation,
+    };
+
+    // Create base handler function with conditional output validation
+    const createBaseHandler = (baseHandler: (event: H3Event) => any) => {
+      if (config.output && validationConfig.output) {
```
OskarLebuda: I think a better approach here would be to use an early exit, like:

```ts
if (!config.output || !validationConfig.config) {
  return baseHandler
}

// rest of the code
```

OskarLebuda (changes requested):

pi0:
Thank you for working on this @productdevbook and thanks for the suggestions @OskarLebuda ❤️ 

I have pushed some refactors (mostly temporary code removals) to make the PR small enough to merge and iterate.

A summary of changes:
- No duplicate `method` and `route` added to custom route meta (h3 already keeps this info in `H3Route.{method,route}`
- Removed extra response and params validation from defineRoute (both really good ideas!!), instead of duplicating a logic that can diverge from `defineValidatedHandler` capabilities, we should add them to `defineValidatedHandler`
- Used flatten options (`validate: { query, ... }`) to be consistent with current `defineValidatedHandler`
- Removed extensive example, the biggest benefit of `defineRoute` is it's **modularity** that encapsulates all route definition. Validation, Type safety and Middleware while also possible with this util, they are generic h3 features via `define*` utils.


Next steps: (if you like to contribute please split PRs, it makes it much faster to iterate 🙏🏼) 
- Improve `defineValidatedHandler` utility: (all will be also supported to `defineRoute`)
  - Auto assign `.meta` from definition (fix: #1147)
  - Use `{ validate: { query, ... } } ` syntax for more clarity (#1148)
  - Support response and route param validation 
  - Probably also a good time to document it in the data validation section (/cc @sandros94)
- Improve `defineRoute` (internal) type safety by augmenting `H3Event` generics from validate functions
- It would be nice if you could make a stackblitz demo for "Books App" using H3 and `defineRoute` we could link it as an interactive example to the docs once tested enough

PR #1144, opened 2025-07-03
feat(proxy): add `secure` option to disable ssl verification

Waleed-KH:
Added a new `secure` option to the proxy. This allows toggling between secure and unsecure connections.

This option will be useful when you need https in dev environment which require to disable SSL verification, otherwise you'll keep getting 502 error.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,8 @@
     "cookie-es": "^2.0.0",
     "fetchdts": "^0.1.5",
     "rou3": "^0.7.3",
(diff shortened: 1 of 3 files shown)


pi0:
I appreciate the PR and however fully agree we need basic features like disabling tls verify, http2, proxy and unix sockets that are only possible via undici, it is a Node.js dependency which we cannot introduce for h3.

I would suggest trying global undici interceptors or passing agent as fetchOptiona to proxt util.

Issue #982, opened 2025-02-24
Custom error status for readValidatedBody, readValidatedQuery, etc


cosbgn:
### Describe the feature

I love `readValidatedBody` but 400 as status error is too generic, I use it for all sort of errors and I would like to have the ability to set a better error code. 

Ideally it would simply be: `readValidatedBody(event, schema.parse, 422)`

it can be 400 as default so it wouldn't be a breaking change, even tough I would personally prefer 422

### Additional information

- [ ] Would you be willing to help implement this feature?

pi0:
PR welcome. Perhaps we could keep it more generic like `readValidatedBody(event, schema, { failStatus: 422 })`. /cc @sandros94 what do you think?

sandros94:
> PR welcome. Perhaps we could keep it more generic like `readValidatedBody(event, schema, { failStatus: 422 })`. /cc [@sandros94](https://github.com/sandros94) what do you think?

Yes, an api that I usually add into my own projects (focused on valibot tho) is something like:

```ts
function createValidationError<DataT = unknown>(
  error: any,
  overrides?: Partial<H3Error<DataT>> & {
    status?: number
    statusText?: string
  },
) {
  return createError({
    status: 400,
    statusMessage: 'Validation failed',
    ...overrides,
    data: error,
    message: v.isValiError(error)
      ? v.summarize(error.issues)
      : error.message
        ? error.message
        : undefined,
  })
}
```

Used indeed as `readValidatedBody(event, schema, { statusMessage: 'My Message' })`

But rethinking about this it would be nice to access the `ReadonlyArray<Issue>` that comes from Standard-Schema's issues via a callback or similar, WDYT?

sandros94:
I did experiment with this last weekend, I do plan to work on what I think is a nice and flexible implementation in the upcoming days/weeks

PR #1146, opened 2025-07-05
feat: customizable validation errors

sandros94:
resolves https://github.com/h3js/h3/issues/982

I've added an additional argument to each `readValidatedBody`, `getValidatedQuery` and `getValidatedRouteParams` which can be either an error object or a function that returns one. If a standard-schema is being used, then the `issues` array is available as an argument for the function variant.

This allows for the following custom error, which will ouput a markdown list of all the errors recorded:
```ts
import * as v from "valibot";

app.post("/", async (event) => {
  const body = await readValidatedBody(
    event,
    v.object({
      name: v.pipe(v.string(), v.minLength(3), v.maxLength(20)),
      age: v.pipe(v.number(), v.integer(), v.minValue(1)),
    }),
    (issues) => ({
      statusText: "Custom validation error",
      message: v.summarize(issues),
    }),
  );
});
```

I've also updated the `defineValidatedHandler`, which brings three more properties: `bodyErrors`, `headersErrors` and `queryErrors`. All allowing for either error object or function that returns one

Added two types, but I'm not super satisfied with `ValidateIssues` naming (I'm open for suggestions):
```ts
export type ValidateIssues = ReadonlyArray<Issue>;
export type ValidateError =
  | ErrorDetails
  | ((issues: ValidateIssues) => ErrorDetails);
```

I've also made sure that if a `new HTTPError` is thrown during validation, its content will be directly passed as is, quite useful for non-standard-schema validations.

### DIFF
--- a/docs/2.utils/1.request.md
+++ b/docs/2.utils/1.request.md
@@ -22,7 +22,7 @@ app.get("/", async (event) => {
 });
 ```
 
-### `readValidatedBody(event, validate)`
+### `readValidatedBody(event, validate, error?)`
 
 Tries to read the request body via `readBody`, then uses the provided validation schema or function and either throws a validation error or returns the result.
 
@@ -31,7 +31,7 @@ You can use a simple function to validate the body or use a Standard-Schema comp
 **Example:**
 
 ```ts
-app.get("/", async (event) => {
+app.post("/", async (event) => {
   const body = await readValidatedBody(event, (body) => {
     return typeof body === "object" && body !== null;
   });
@@ -42,12 +42,31 @@ app.get("/", async (event) => {
 
 ```ts
 import { z } from "zod";
-app.get("/", async (event) => {
+app.post("/", async (event) => {
   const objectSchema = z.object({
     name: z.string().min(3).max(20),
     age: z.number({ coerce: true }).positive().int(),
   });
   const body = await readValidatedBody(event, objectSchema);
+});
+```
+
+**Example:**
+
+```ts
+import * as v from "valibot";
+app.post("/", async (event) => {
+  const body = await readValidatedBody(
+    event,
+    v.object({
+      name: v.pipe(v.string(), v.minLength(3), v.maxLength(20)),
+      age: v.pipe(v.number(), v.integer(), v.minValue(1)),
+    }),
+    (issues) => ({
+      statusText: "Custom validation error",
+      message: v.summarize(issues),
+    }),
+  );
 });
 ```
 
@@ -188,7 +207,7 @@ app.get("/", (event) => {
 });
 ```
 
-### `getValidatedQuery(event, validate)`
+### `getValidatedQuery(event, validate, error?)`
 
 Get the query param from the request URL validated with validate function.
 
@@ -218,7 +237,25 @@ app.get("/", async (event) => {
 });
 ```
 
-### `getValidatedRouterParams(event, validate, opts: { decode? })`
+**Example:**
+
+```ts
+import * as v from "valibot";
+app.get("/", async (event) => {
+  const params = await getValidatedQuery(
+    event,
+    v.object({
+      key: v.string(),
+    }),
+    (issues) => ({
+      statusText: "Custom validation error",
+      message: v.summarize(issues),
+    }),
+  );
+});
+```
+
+### `getValidatedRouterParams(event, validate, opts: { decode? }, error?)`
 
 Get matched route params and validate with validate function.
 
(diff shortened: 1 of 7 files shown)


sandros94:
Oh, is it normal that we have a `decode` option for route params but not for queries? 🤔

src/handler.ts
```diff
@@ -61,8 +65,11 @@ export function defineValidatedHandler<
 >(def: {
   middleware?: Middleware[];
   body?: RequestBody;
+  bodyErrors?: ValidateError;
   headers?: RequestHeaders;
+  headersErrors?: ValidateError;
   query?: RequestQuery;
+  queryErrors?: ValidateError;
```
pi0: Wondering if we could instead support a simple object for custom validation error, something like:

```js
validateError: (error: { issues }, source: 'body' | 'headers  | 'query' 
```

This way we don't need to add a 2x new option for each validator, validator logic can be shared and also object format can be removed too IMO, as function can cover it.

src/utils/body.ts
```diff
@@ -52,7 +52,7 @@ export async function readBody<
 export async function readValidatedBody<
   Event extends H3Event,
   S extends StandardSchemaV1,
->(event: Event, validate: S): Promise<InferOutput<S>>;
+>(event: Event, validate: S, error?: ValidateError): Promise<InferOutput<S>>;
```
pi0: For future API extensions (if we need to support more options), it would be better if 3rd arg is an Options object with `error` or `formatError` as key.

src/handler.ts
sandros94: I totally agree with dropping the object format, makes more sense in a _on error hook_ perspective.

For `defineValidatedHandler` I like this api (https://github.com/h3js/h3/pull/1146/commits/ec3aab8f075d0bcc5c58af2c4930715e9c118401), WDYT?
```ts
  validationErrors?: {
    body?: (issues: ValidateIssues) => ErrorDetails;
    headers?: (issues: ValidateIssues) => ErrorDetails;
    query?: (issues: ValidateIssues) => ErrorDetails;
  };
```

pi0: I like it, only probably simpler `validateErrors` instead of `validationErrors`? (i plan to use `validate: {}` also for defineValidatedHandler it might be better for consistency)

test/handler.test.ts
```diff
@@ -105,27 +105,29 @@ describe("handler.ts", () => {
...
+          status: 500,
+          statusText: "Custom Zod body validation error",
+          message: summarize(issues),
+        }),
+        headers: (issues) => ({
+          status: 500,
+          statusText: "Custom Zod headers validation error",
+          message: summarize(issues),
+        }),
+        query: (issues) => ({
```
pi0: My feeling is that most of usages, this will be very similar for formatting different validation sources also considering if there is a new feature (for example response validation), users need to explicitly add another handler.


We could keep it simple validationError: (error, source) => {}

docs/2.utils/1.request.md
```diff
@@ -51,6 +51,27 @@ app.get("/", async (event) => {
...
+import * as v from "valibot";
+app.post("/", async (event) => {
+  const body = await readValidatedBody(
+    event,
+    v.object({
+      name: v.pipe(v.string(), v.minLength(3), v.maxLength(20)),
+      age: v.pipe(v.number(), v.integer(), v.minValue(1)),
+    }),
+    {
+      onError: (issues) => ({
```
pi0: Can input be in { issues } format (or we pass whole result to format error?

Both for forward compatibility and also support custom errors

sandros94:
@pi0 any update on this?

pi0:
Hi dear @sandros94. It is in my todo list, in meantime do you have time to help on fixing merge conflicts and adressing https://github.com/h3js/h3/pull/1146#discussion_r2188235982 🙏🏼

sandros94:
> Hi dear @sandros94. It is in my todo list, in meantime do you have time to help on fixing merge conflicts and adressing https://github.com/h3js/h3/pull/1146#discussion_r2188235982 🙏🏼 

Yes absolutely! And sorry, I completely forgot about that review 😅

sandros94:
I'm terribly sorry, I didn't realize that in 1e4d72079488c1d63bcb3f82491b8559883bbb71 I did not update the docs 🙈

For now I would leave it as only `{ issues }`, to better follow the [standard-schema spec `FailureResult`](https://github.com/standard-schema/standard-schema/blob/266a01bad23b706fc3e23471f2e9a1ffbbca0824/packages/spec/README.md#the-interface) and leave room for future expansion.

Let me know if there is anything else I should take care of

pi0:
@sandros94 updated implementation to be more compact API. Love your opinion on latest changes.

sandros94:
> @sandros94 updated implementation to be more compact API. Love your opinion on latest changes.

Thank you, and yes it required a bit of cleanup

pi0 (approved):

Issue #1138, opened 2025-07-01
StandardSchema-Compatible WebSocket Route Definition


productdevbook:
### Describe the feature

```ts
import { defineWebSocketRoute } from "h3";
import { z } from "zod";

export const webhook = defineWebSocketRoute({
  method: "GET",
  path: "/api/webhook",
  websocket: {
    upgrade: (request) => {
      console.log("Opening WebSocket connection:", request.url);
    },
    open: (peer) => {
      console.log("WebSocket connected:", peer.id);
      peer.send("Welcome!");
    },
    message: (peer, message) => {
      console.log("Incoming message:", message);
      peer.send(`Received: ${message}`);
    },
    close: (peer) => {
      console.log("WebSocket closed:", peer.id);
    },
  }
});
``` 

### Additional information

- [x] Would you be willing to help implement this feature?

pi0:
I love the idea together with #1088, lets track it from single place.

PR #1149, opened 2025-07-07
feat: experimental defineWebSocketRoute

productdevbook:
## Summary
cc: https://github.com/h3js/h3/issues/1088
cc: https://github.com/h3js/h3/issues/1138

- Added `defineWebSocketRoute` utility function to simplify WebSocket route definitions
- Added `WebSocketRouteDefinition` interface for type-safe WebSocket route configuration
- Exported new utilities from main index.ts
- WebSocket routes always use GET method (as per WebSocket upgrade spec)

## Description

This PR introduces a new `defineWebSocketRoute` function that provides a consistent API pattern with `defineRoute` for defining WebSocket endpoints. This makes it easier to create WebSocket routes as plugins that can be registered with `app.register()`.

### Interface Definition

```typescript
export interface WebSocketRouteDefinition {
  /** Route pattern, e.g. '/api/ws' */
  route: string;
  
  /** WebSocket hooks */
  websocket: Partial<WSHooks>;
  
  // TODO: Support middleware when implemented in defineWebSocketHandler
  // TODO: Support metadata when implemented in WebSocket routing
}
```

**Key design decisions:**
- WebSocket routes always use `GET` method (standard for WebSocket upgrades)
- No `method` parameter - WebSocket protocol requires GET
- `middleware` and `meta` support deferred until implemented in underlying `defineWebSocketHandler`

### Example usage:
```typescript
import { defineWebSocketRoute } from "h3";

export const chatRoute = defineWebSocketRoute({
  route: "/api/chat",
  websocket: {
    open: (peer) => {
      console.log("WebSocket connected:", peer.id);
      peer.send("Welcome!");
    },
    message: (peer, message) => {
      console.log("Incoming message:", message);
      peer.send(`Echo: ${message}`);
    },
    close: (peer) => {
      console.log("WebSocket closed:", peer.id);
    },
  }
});

// Register with app
app.register(chatRoute);
```

### Compatibility with existing APIs

This is fully compatible with the existing `defineWebSocketHandler`:

```typescript
// New way (as plugin)
app.register(defineWebSocketRoute({ route: "/ws", websocket: hooks }));

// Traditional way (still works)
app.on("GET", "/ws", defineWebSocketHandler(hooks));
```

## Test plan
- [x] Test WebSocket route registration with `app.register()`
- [x] Verify WebSocket upgrade works correctly
- [x] Test with different route patterns (including params like `/ws/:id`)
- [x] Ensure backwards compatibility with existing WebSocket handler methods
- [x] Verify GET method is always used
- [x] All tests passing after merge with main

## Review feedback addressed
- Removed `method` parameter - WebSocket routes always use GET (@pi0)
- Removed `middleware` and `meta` properties - not yet supported (@pi0)
- Added TODO comments for future feature support (@productdevbook)
- Types kept co-located in same file (@pi0)
- Property name stays `websocket` not `handler` (@pi0)


src/utils/route.ts
```diff
@@ -69,3 +71,65 @@ export function defineRoute(def: RouteDefinition): H3Plugin {
     h3.on(def.method, def.route, handler);
   };
 }
+
+/**
+ * WebSocket route definition options
+ */
+export interface WebSocketRouteDefinition {
```
OskarLebuda, line 78: Can we do a bit of refactoring and move the types from here into a separate file? Same goes for `RouteDefinition`.

OskarLebuda (changes requested):

src/utils/route.ts
```diff
@@ -69,3 +71,65 @@ export function defineRoute(def: RouteDefinition): H3Plugin {
...
+
+  /**
+   * Route pattern, e.g. '/api/ws'
+   */
+  route: string;
+
+  /**
+   * WebSocket hooks
+   */
+  websocket: Partial<WSHooks>;
```
productdevbook, line 87: websocket name -> handler ? maybe.

src/utils/route.ts
```diff
@@ -69,3 +71,65 @@ export function defineRoute(def: RouteDefinition): H3Plugin {
...
+
+  /**
+   * Optional middleware to run before WebSocket upgrade
+   */
+  middleware?: Middleware[];
+
+  /**
+   * Additional route metadata
+   */
+  meta?: Record<string, unknown>;
```
pi0: Middleware and meta are not supported. Types should be removed (at least untill they are supported by `defineWebSocket`

src/utils/route.ts
pi0: Let's keep using `websocket`. `handler` is not same defenition.

src/utils/route.ts
```diff
@@ -69,3 +71,65 @@ export function defineRoute(def: RouteDefinition): H3Plugin {
...
 }
+
+/**
+ * WebSocket route definition options
+ */
+export interface WebSocketRouteDefinition {
+  /**
+   * HTTP method for the route (typically 'GET' for WebSocket upgrades)
+   */
+  method?: HTTPMethod;
```
pi0: WebSocket handlers should not accept a method.

src/utils/route.ts
pi0: No, same as `RouteDefinition`, i prefer types of utils to be co-located in same file.

src/utils/route.ts
productdevbook: We should take the comment line and leave a TODO ?

productdevbook:
Hi team! I've addressed all the review feedback:

**Changes made:**

1. ✅ **Removed `method` property** (@pi0's feedback)
   - WebSocket routes now always use GET method
   - Removed the optional `method` parameter from `WebSocketRouteDefinition`

2. ✅ **Removed `middleware` and `meta` properties** (@pi0's feedback)
   - These are not yet supported by `defineWebSocketHandler`
   - Added TODO comments for future implementation (@productdevbook's suggestion)

3. ✅ **Kept types in the same file** (@pi0's preference)
   - `WebSocketRouteDefinition` stays co-located in `src/utils/route.ts`

4. ✅ **Kept `websocket` property name** (@pi0's decision)
   - Not changing to `handler` as they have different semantics

5. ✅ **Merged latest from main**
   - Updated tests to use `~routes` internal API
   - Updated to use `app.request()` instead of `app.fetch()`
   - All tests passing ✅

**Commits:**
- `4369b94` - Address PR review feedback
- `81eec94` - Merge main branch
- `2ca3e4f` - Update tests for API changes

The branch is ready for re-review! 🚀

PR #1150, opened 2025-07-07
feat: request body limit

### LINKED ISSUE #859: Limit upload size
### Describe the feature

I use nuxt for my api.
I am only reading a password from a post request body. But if I upload a large file (>2GB) it loads everything into ram.

Is there a way to close the request after e.g. 1kb?

### Additional information

productdevbook:
## Summary
This PR introduces a new plugin system for limiting request body sizes in H3 applications, addressing the issue raised in #859.

## Features
- 🔌 Plugin-based approach for flexible body size limiting
- 🎯 Route-specific limits with include/exclude patterns
- ⚡ Performance-optimized with Content-Length header checking
- 🧪 Comprehensive test coverage

## Usage
```js
import { defineBodySizeLimitPlugin } from "h3";

const bodySizeLimit = defineBodySizeLimitPlugin({
  maxSize: 1024 * 1024, // 1MB
  routes: ["/api/upload", /^\/api\/files/],
  exclude: ["/api/large-upload"]
});

app.register(bodySizeLimit);
```

## Why Plugin Approach?
Instead of adding `maxBodySize` directly to route definitions, this plugin approach provides:
- Better performance (no overhead when not needed)
- More flexibility (apply to multiple routes at once)
- Cleaner separation of concerns
- Easy to enable/disable globally

## Test Coverage
- ✅ Global body size limiting
- ✅ Route-specific limiting with patterns
- ✅ Route exclusion functionality
- ✅ Integration with `defineRoute`

Closes #859

### DIFF
--- a/test/unit/package.test.ts
+++ b/test/unit/package.test.ts
@@ -27,6 +27,7 @@ describe("h3 package", () => {
         "createEventStream",
         "createRouter",
         "defaultContentType",
+        "defineBodySizeLimitPlugin",
         "defineEventHandler",
         "defineHandler",
         "defineLazyEventHandler",


src/utils/body-limit.ts
```diff
@@ -0,0 +1,86 @@
...
+ * ```
+ */
+export function defineBodySizeLimitPlugin(
+  options: BodySizeLimitOptions,
+): H3Plugin {
+  return (h3) => {
+    h3.use(
+      defineMiddleware(async (event: H3Event) => {
+        const url = event.req.url;
+        const path = url ? new URL(url, "http://localhost").pathname : "/";
```
OskarLebuda, line 45: Creating a new `URL` object for every request is computationally expensive - it involves parsing the entire URL structure, validating components, and creating multiple internal objects.
This operation can be 5–10x slower than simple string operations when you're only trying to extract the pathname.

You can optimize it by using the utility from `srvx`:
[https://github.com/h3js/srvx/blob/main/src/\_url.ts](https://github.com/h3js/srvx/blob/main/src/_url.ts)
or just go with something like:

```ts
event.req.url?.split('?')[0] || '/'
```

src/utils/body-limit.ts
```diff
@@ -0,0 +1,86 @@
...
+
+        // Check if route should be excluded
+        if (options.exclude) {
+          for (const pattern of options.exclude) {
+            if (typeof pattern === "string" && path === pattern) return;
+            if (pattern instanceof RegExp && pattern.test(path)) return;
+          }
+        }
```
OskarLebuda, line 53: The code currently performs a linear search through all exclude patterns on every single request, which gets slower as the number of patterns increases.

It’s also repeatedly doing type checks like `(typeof pattern === "string")` for the same patterns on every request.
You can optimize this by pre-processing the patterns during plugin initialization - split them into two collections:

* A `Set` for string patterns (for O(1) lookup)
* An array for `RegExp` patterns

This way, you eliminate runtime type checking and significantly improve string matching performance.

src/utils/body-limit.ts
```diff
@@ -0,0 +1,86 @@
...
+          const size = Number.parseInt(contentLength, 10);
+          if (!Number.isNaN(size) && size > options.maxSize) {
+            throw new HTTPError({
+              status: 413,
+              statusText: "Payload Too Large",
+              message: `Request body size ${size} exceeds the limit of ${options.maxSize} bytes`,
+            });
+          }
+        }
+      }),
```
OskarLebuda, line 83: The current implementation only validates the `Content-Length` header, which can be missing in chunked transfer encoding, deliberately omitted, or even falsified by malicious clients to bypass the limit. This creates a significant security vulnerability where large payloads can still get through.

OskarLebuda (changes requested):

src/utils/body-limit.ts
```diff
@@ -0,0 +1,86 @@
...
+ * const bodySizeLimit = defineBodySizeLimitPlugin({
+ *   maxSize: 1024 * 1024, // 1MB
+ *   routes: ["/api/upload", /^\/api\/files/],
+ *   exclude: ["/api/large-upload"]
+ * });
+ *
+ * app.register(bodySizeLimit);
+ * ```
+ */
+export function defineBodySizeLimitPlugin(
```
pi0, line 38: It should be a `middleware` not a plugin to apply in places needed.

With middleware, pattern and route matching can be defined also on definition therefore we do not need to (re)implement logic inside this utility.

src/utils/body-limit.ts
pi0: https://github.com/h3js/h3/pull/1150/files#r2189726225, middleware itself should not add extra logic for route matching

src/utils/body-limit.ts
pi0: https://github.com/h3js/h3/pull/1150/files#r2189726225

src/utils/body-limit.ts
```diff
@@ -0,0 +1,86 @@
...
+            if (pattern instanceof RegExp && pattern.test(path)) {
+              matches = true;
+              break;
+            }
+          }
+          if (!matches) return;
+        }
+
+        // Check Content-Length header first
+        const contentLength = event.req.headers.get("content-length");
```
pi0, line 72: This is not a safe implementation as @OskarLebuda pointed out. `content-length` can be faked. 

We need to both check header and transform body stream into another controller that force-stops reading body as soon as max length is reached.

src/utils/body-limit.ts
pi0: https://github.com/h3js/h3/pull/1150/files#r2189537524

pi0:
Thanks for PR again my dear friend. Similar implementation added in #1222

PR #1152, opened 2025-07-08
feat: log unhandled errors

### LINKED ISSUE #1151: errors thrown in handlers cannot be debugged
### Environment

2.0.0-20250702-142007-a8a12f8 + nodejs 23.11.0

### Reproduction

```ts
import { serve, H3 } from "h3";

const app = new H3();

app.get("/", (e) => {
	throw new Error("hello");
});

serve(app);
```

### Describe the bug

response:
```json
{ "status": 500, "unhandled": true, "message": "HTTPError" }
```
console: *nothing*

### Additional context

when `debug` is set to `true`, the response will include the error stack which is good.
i think the error should be logged to the console regardless of the `debug` value.

### Logs

```sh

```

huseeiin:
fix #1151 by logging unhandled errors to the console (in despite of `debug`)


huseeiin:
btw i don't know why there are 2 other commits :neutral_face: i just created this branch, apologies if its my fault

src/h3.ts
```diff
@@ -74,6 +74,7 @@ export const H3Core = /* @__PURE__ */ (() => {
           handlerRes = this.handler(event);
         }
       } catch (error: any) {
+        console.error(error);
```
pi0: I think we could move it to [here](https://github.com/h3js/h3/blob/main/src/response.ts#L59) and log any unhandled error.

We also need to add a new `silent: boolean` config to allow opting-out from this behavior.

Issue #875, opened 2024-09-02
Common pattern with `readBody` may lead to CSRF


OhB00:
### Environment

N/A

### Reproduction

The following pattern for creating endpoints is common, and may lead to a CSRF attack, provided it is a `GET`/`POST` handler, and only requires strings, or arrays of strings in the body.

The following is an example of what a vulnerable handler might look like:
```js
export default defineEventHandler(async e => {
  // Authenticated endpoint
  await VerifyAuthenticationHere(e)

  // Assumption here that content-type is validated.
  const body = await readBody(e)

  // Optional validation here ...

  // Some sensitive action that relies upon body
  if (body.action === "update") {
    await UpdateUser(body.new_username)
  }
})
```

### Describe the bug

The `readBody` utility will accept multiple different content type's, including `application/json` `application/x-www-form-urlencoded` and `text/plain`. 

There is an assumption that the `readBody` handler will only emit an object for the `application/json` content type, however, this also occurs for `application/x-www-form-urlencoded`. 

This has implications for security as cross-site requests can be sent when using the ` application/x-www-form-urlencoded`, but they cannot be sent when using `application/json`.

If `readBody` is called without appropriate CSRF protections, or without verifying the `content-type` header, the application may be at risk of a [CSRF attack](https://owasp.org/www-community/attacks/csrf).

### Additional context

The following handler is an example of how an attacker could exploit one of these issues:
```js
await fetch('https://whatever.io/api/edit', {
    method: 'POST',
    headers:{
      'Content-Type': 'application/x-www-form-urlencoded'
    },    
    body: new URLSearchParams({
        'username': 'hacked!',
    }),
    mode: 'no-cors'
}
```

pi0:
Closing as note added and for h3 v2 we are advocating `event.req.json()` which is strict JSON body parsing using web standard Requeest. Thanks again 🙏🏼

PR #1164, opened 2025-07-18
feat(readBody): support formdata type

huseeiin:
## Summary by CodeRabbit

* **New Features**
  * Added an optional `options` parameter to request body parsing, including `type: "formData"` for `multipart/form-data`.
  * Exposed a `ReadBodyOptions` type for the new API.

* **Bug Fixes**
  * Default parsing remains JSON, with `400 Bad Request` for invalid JSON.
  * URL-encoded bodies are parsed from `application/x-www-form-urlencoded` (or when explicitly requested).
  * Empty bodies still return `undefined`; repeated fields are preserved as arrays.

* **Documentation**
  * Updated `readBody` documentation and examples to reflect the default JSON behavior and explicit multipart opt-in.

* **Tests**
  * Added multipart/form-data tests (valid parsing, missing opt-in behavior, and malformed payload errors).

### DIFF
--- a/src/utils/body.ts
+++ b/src/utils/body.ts
@@ -26,11 +26,15 @@ export async function readBody<
   _T = InferEventInput<"body", _Event, T>,
 >(event: _Event): Promise<undefined | _T> {
   const text = await event.req.text();
+  const contentType = event.req.headers.get("content-type") || "";
+
+  if (contentType.startsWith("multipart/form-data"))
+    return Object.fromEntries(await event.req.formData()) as _T;
+
   if (!text) {
     return undefined;
   }
 
-  const contentType = event.req.headers.get("content-type") || "";
   if (contentType.startsWith("application/x-www-form-urlencoded")) {
     return parseURLEncodedBody(text) as _T;
   }
--- a/src/utils/body.ts
+++ b/src/utils/body.ts
@@ -7,6 +7,18 @@ import type { InferEventInput } from "../types/handler.ts";
 import type { ValidateResult } from "./internal/validate.ts";
 import type { StandardSchemaV1, FailureResult, InferOutput } from "./internal/standard-schema.ts";
 
+export interface ReadBodyOptions {
+  /**
+   * Force a parser instead of inferring it from the request `Content-Type`.
+   *
+   * - `"json"` (default): parse as JSON.
+   * - `"text"`: return the raw string body.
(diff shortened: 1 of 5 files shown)


src/utils/body.ts
```diff
@@ -29,11 +29,15 @@ export async function readBody<
   _T = InferEventInput<"body", _Event, T>,
 >(event: _Event): Promise<undefined | _T> {
   const text = await event.req.text();
+  const contentType = event.req.headers.get("content-type") || "";
+
+  if (contentType.startsWith("multipart/form-data"))
```
pi0: We should make it opt-in by supporting `event, opts?: { type: "json" | formData" }` and throw an HTTPError if it is not.

huseeiin: > We should make it opt-in by supporting `event, opts?: { type: "json" | formData" }` and throw an HTTPError if it is not.

i support this. but my changes remain valid, right? since `opts` is optional. when you specify `type` in `opts`, you're specifying a certain body type, when you're not, you're parsing any body as an object or string!

pi0: As security measure we should make types opt-in other than JSON (so if no options only json)

pi0: Context: https://github.com/h3js/h3/issues/875

pi0x:
Pushed e2e146c reworking this toward the opt-in design discussed above (#875).

**What changed**
- `readBody(event, { type })` — `formData` is now **strictly opt-in** and never auto-detected from `Content-Type`, so an untrusted request can no longer push `readBody` into multipart parsing on its own. Default JSON and existing `x-www-form-urlencoded` auto-parse are unchanged (no breaking change).
- Malformed multipart now throws a clean `HTTPError(400)`, matching the JSON path (addresses CodeRabbit's error-handling comment).
- Repeated form fields are preserved as arrays instead of collapsing to the last value — shared with url-encoded parsing (addresses CodeRabbit's duplicate-field comment).
- Added `readBody`-based multipart tests (opt-in parse, no auto-parse without opt-in, malformed → 400), running in both web and node.

**Open question:** the thread said "if no options only json" — I kept the existing `x-www-form-urlencoded` auto-parse by default to avoid a breaking change beyond this PR. If you want strict JSON-only default, it's a one-line change plus updating the one urlencoded test to pass `{ type: "urlencoded" }`.

> 🤖 This change was drafted with Claude Code (Opus 4.8) and reviewed by me before pushing.

pi0 (approved):

Issue #1166, opened 2025-07-22
handleCors() returns falsy sentinel (""), conflicting with example usage.


jakobrosenberg:
### Environment

h3: 2.0.0-beta.1

### Reproduction

Simply check the declared type:

```javascript
declare function handleCors(event: H3Event, options: CorsOptions): false | "";
```

### Describe the bug

**Description:**
`handleCors()` returns an empty string (`""`) when handling preflight requests. However, the official example suggests using a truthy check `(if (corsRes))`, which fails since `""` is falsy. This makes the example incorrect and the API unintuitive.

**Expected:**
Return a truthy sentinel (e.g. `true` or `"handled"`) to indicate preflight handled, matching the example pattern.

**Actual:**
Returns falsy sentinel (`""`), making:

```javascript
if (corsRes) return corsRes
```
ineffective.

### Logs

```sh

```

PR #1167, opened 2025-07-23
docs: fix description of`handleCors`'s return value

kricsleo:
resolves https://github.com/h3js/h3/issues/1166

(Also wondering if we can revert the `v1`'s `boolean` return value to minimize breaking changes 🤔)

### DIFF
--- a/docs/2.utils/4.security.md
+++ b/docs/2.utils/4.security.md
@@ -97,7 +97,7 @@ Handle CORS for the incoming request.
 
 If the incoming request is a CORS preflight request, it will append the CORS preflight headers and send a 204 response.
 
-If return value is `true`, the request is handled and no further action is needed.
+If return value is not `false`, the request is handled and no further action is needed.
 
 **Example:**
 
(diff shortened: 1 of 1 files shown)


pi0:
Yes i think might make sense for v1 (also v2) to avoid returning falsy like `""` value.

Do you know why we did that? (i have to check history) but seems your suggestion for avoid that is more sensable than asking for `=== false` check.

kricsleo:
I think this change might be unintentional. Firstly, it was [`return true`](https://github.com/h3js/h3/pull/783/files#diff-23f136e88f626c4862f90495c4a3abffbc9990eb4fff430d18b485dc7397dbdfR82) when handled, then it was changed to [`return noContent(...)`](https://github.com/h3js/h3/pull/792/files#diff-23f136e88f626c4862f90495c4a3abffbc9990eb4fff430d18b485dc7397dbdfR84). From this point on, the falsy check no longer works as expected since `noContent` would return an empty string.

randName:
I wonder if `noContent` should just construct a proper `Response` instead of returning an empty string, then the original check should work

kricsleo:
>  `noContent` should just construct a proper `Response`

Yeah, I agree with this; it would be more usable. 👍
This PR can be safely closed if https://github.com/h3js/h3/pull/1177 is merged.

PR #1169, opened 2025-07-23
feat(EventStream): support comment event

### LINKED ISSUE #1168: SSE support for comment message
### Environment

h3 v1.15.3

### Reproduction

Create an EventStream and try to push a comment:

```ts
import { createApp, createRouter, eventHandler, createEventStream } from "h3";

app.use(router);

router.get(
  "/",
  eventHandler((event) => {
    const eventStream = createEventStream(event);

    // Send a heartbeat to prevent connections from timing out
    const interval = setInterval(async () => {
      await eventStream.push({ comment: 'keep-alive' }); // ts error
    }, 1000);

    // cleanup the interval when the connection is terminated or the writer is closed
    eventStream.onClosed(() => {
      clearInterval(interval);
    });

    return eventStream.send();
  }),
);
```

### Describe the bug

The EventStream does support comment message so it is impossible to implement a heartbeat, which may be crucial if your server is behind an HAProxy killing idle connections.

See [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#event_stream_format)
> Note: The comment line can be used to prevent connections from timing out; a server can send a comment periodically to keep the connection alive.

### Additional context

_No response_

### Logs

```sh

```

HugoMuller:
resolves #1168  

Add support for comment, so doing this is now possible (see #1168 for a full reproduction scenario):

```ts
const eventStream = createEventStream(event);
await eventStream.push({ comment: 'keep-alive' });
```


pi0:
Nice idea. Howevr combining API to send normal messages with comments might be tricky since someone might want  to send an actual object with `comment` key in it.

We might introduce `ee.pushComment(...)` instead.

Can you please also rebase to `main` branch? (we could backport if necessary after)

HugoMuller:
I rebased and the PR targets `base` now.

I kept `ee.push(...)` untouched and added a new method `ee.pushComment()`.
I also introduced a new `formatEventStreamComment()` but I wanted to keep only 1 `_sendEvent` to handle both comment and message.
Is this ok by you?

pi0:
Thanks for update. It would be probably better to keep these two fully seperated to avoid internal bugs.

PR #1171, opened 2025-07-24
docs: fix example for noContent util

### LINKED ISSUE #1170: wrong example for noContent utility
### Describe the change

The `noContent` utility expects `event` to be passed but the given example had left it out

```js
app.get("/", () => noContent());
```

### URLs

https://h3.dev/utils/response#nocontentevent-code

### Additional information

- [x] Would you be willing to help?

randName:
resolves #1170

I chose to explicitly pass in the event rather than do it implicitly like this

```js
app.get("/", noContent);
```

in case there are [future signature changes](https://jakearchibald.com/2021/function-callback-risks/)

### DIFF
--- a/docs/2.utils/2.response.md
+++ b/docs/2.utils/2.response.md
@@ -117,7 +117,7 @@ Respond with an empty payload.<br>
 **Example:**
 
 ```ts
-app.get("/", () => noContent());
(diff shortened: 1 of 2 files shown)


pi0 (approved):
Thanks!

Issue #1173, opened 2025-07-25
backport X-Forwarded-Host fix to v1


aturingmachine:
### Environment

The npm version tagged `1.15.3` does not include the changes seen in PR #993. The release for `1.15.3` was cut on April 28, 2025 and #993 was merged to main on April 16, 2025.

The PR merged commit is cb9fc4736644348b7130af559f0909fb40fa95c4, which is said to appear in the tag `1.15.3` - which it does in the GitHub tag, however it does not appear in the code on NPM for that same tag.

> Apologies that I cannot permalink to the actual lines in question via NPM

- [NPM 1.15.3 Code](https://www.npmjs.com/package/h3/v/1.15.3?activeTab=code)
  - `/h3/dist/index.mjs:309`
- [NPM 2.0.0-beta.1 Code](https://www.npmjs.com/package/h3/v/2.0.0-beta.1?activeTab=code)
  - `/h3/dist/h3.mjs:1016`
- [GitHub tag 1.15.3 `getRequestHost` function](https://github.com/h3js/h3/blob/5bd27a52d752197fe8c9bc395cfc245977dc9b42/src/utils/request.ts#L212)
  - This is the function I would expect to see in the packaged code on npm

Notably the `1.15.3` code _on NPM_ is missing the `.split` that is called on the `x-forwarded-host` header. 

### Reproduction

There is no reproduction as this issue is related to the NPM package and not a runtime issue.

### Describe the bug

I was expecting to get the fix from PR #993 in the `1.15.3` version of h3, however that code is not in the packaged version on NPM.

### Logs

```sh

```

pi0:
`main` is h3 v2 (github UI somehow wrongly showing it), feel free to open a PR to `v1` to backport changes

aturingmachine:
> `main` is h3 v2 (github UI somehow wrongly showing it), feel free to open a PR to `v1` to backport changes

Is this to say that even when selecting the tag `1.15.3` GitHub was showing me the v2 code? (I don't doubt this as a possibility, just want to clarify).

I have downloaded the `1.15.3` assets from the release tag, and the fix is in that code. The [tag diff](https://github.com/h3js/h3/compare/v1.15.2...v1.15.3#diff-ec5a1223686ea9c57fb3593ca303c83c06211b8ca65d8080a1ef77610c898ddbR217) also shows the code as being there, so this would be a huge issue with GitHub.

pi0:
[...]

PR #1175, opened 2025-07-27
fix(getRequestHost): return first host from `x-forwarded-host`

kricsleo:
resolves https://github.com/h3js/h3/issues/1173

This PR backports https://github.com/h3js/h3/pull/993 to `v1`. ❤️


pi0 (approved):
Thanks!

PR #1176, opened 2025-07-30
feat!: `app.request`

pi0:
Followup on #1096 with `app.fetch(req)` becoming more restricted, it makes sense we make `app.request(req, init, ctx)` a public API to unblock DX limits.

Other API change: Types are now `Response | Promise<Response>` to reduce unnecessary wrappings. (please comment if anyone feels this is a bad DX)

### DIFF
--- a/docs/1.guide/900.api/1.h3.md
+++ b/docs/1.guide/900.api/1.h3.md
@@ -18,7 +18,7 @@ const app = new H3({
 
 ## `H3` Methods
 
-### `H3.fetch`
+### `H3.request`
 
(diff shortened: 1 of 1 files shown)


huseeiin:
yes i think it should be `Promise<Response>` to match `globalThis.fetch`. imagine someone making an SSR framework and they override `globalThis.fetch` so it works like SPA (similar to nuxt `useFetch`, P.S. i know `useFetch` is not overriding `globalThis.fetch`). users then will use `fetch("/route").then(res => res.json())` and get `fetch("/route").then is undefined` error

pi0:
Wrappers can easily use `Promise.resolve()` when overriding `globalThis.fetch` to match.

PR #1177, opened 2025-07-31
fix: return a `Response` for no-content

### LINKED ISSUE #1166: handleCors() returns falsy sentinel (""), conflicting with example usage.

randName:
resolves #1166
supersedes #1167

This will allow the original

```js
if (corsRes) return corsRes
```

check to work, and as a bonus this might resolve some other footguns due to `noContent` returning a falsy `""`

Not sure if this is the best way to create the `FastResponse`, and also not sure about what to do with the type for the deprecated `sendNoContent`


randName:
tests for those will be covered by #1161

pi0 (approved):

pi0:
Thanks!

PR #1180, opened 2025-08-02
feat(handler): new `jsonRpcHandler`

sandros94:
An experimental handler to support the [JSON RPC spec](https://www.jsonrpc.org/specification).

### Usage

you define it like a normal handler, but with the added `params` as its first callback argument:

```ts
const eventHandler = jsonRpcHandler({
  test: (params, event) => {
    return `Recieved \`${params}\` on path \`${event.url.pathname}\``;
  },
});
app.post("/rpc"; eventHandler); // Must be a POST method
```

then you call it with a valid body, for example:

```json
{
  jsonrpc: "2.0",
  id: 1,
  method: "test",
  params: "Hello World",
}
```

and you'll receive a json-rpc compatible response:

```json
{
  jsonrpc: "2.0",
  id: 1, // Same id from the request
  result: "Recieved `Hello World` on path `/rpc`",
}
```

it also support client-to-server notifications by simply omitting the `id` key, this will result in the server responding a `202`

### Personal considerations and future expansion

- Support built-in schema validation (probably allowing for an object instead of directly the event callback)

### DIFF
--- a/src/utils/json-rpc.ts
+++ b/src/utils/json-rpc.ts
@@ -1,0 +1,212 @@
+import type { H3Event } from "../event.ts";
+import { defineHandler } from "../handler.ts";
+import type { EventHandler } from "../types/handler.ts";
+import { HTTPError } from "../error.ts";
+import { readBody } from "./body.ts";
+
+/**
+ * JSON-RPC 2.0 Interfaces based on the specification.
+ * https://www.jsonrpc.org/specification
+ */
+
+/**
+ * JSON-RPC 2.0 Request object.
+ */
+interface JsonRpcRequest<T = unknown> {
+  jsonrpc: "2.0";
+  method: string;
+  params?: T;
+  id?: string | number | null;
+}
+
+/**
+ * JSON-RPC 2.0 Error object.
+ */
+interface JsonRpcError {
+  code: number;
+  message: string;
+  data?: any;
+}
+
+/**
+ * JSON-RPC 2.0 Response object.
+ */
+interface JsonRpcResponse<D = unknown> {
+  jsonrpc: "2.0";
+  result?: D;
+  error?: JsonRpcError;
+  id: string | number | null;
+}
+
+// Official JSON-RPC 2.0 error codes.
+/**
+ * Invalid JSON was received by the server. An error occurred on the server while parsing the JSON text.
+ */
+const PARSE_ERROR = -32_700;
+/**
+ * The JSON sent is not a valid Request object.
+ */
+const INVALID_REQUEST = -32_600;
+/**
+ * The method does not exist / is not available.
+ */
+const METHOD_NOT_FOUND = -32_601;
+/**
+ * Invalid method parameter(s).
+ */
+const INVALID_PARAMS = -32_602;
+/**
+ * Internal JSON-RPC error.
+ */
+const INTERNAL_ERROR = -32_603;
+// -32_000 to -32_099 	Reserved for implementation-defined server-errors.
+
+/**
+ * A function that handles a JSON-RPC method call.
+ * It receives the parameters from the request and the original H3Event.
+ */
+export type JsonRpcMethodHandler<T = unknown, D = unknown> = (
+  params: T,
+  event: H3Event,
+) => D | Promise<D>;
+
+/**
+ * A map of method names to their corresponding handler functions.
+ */
+export type JsonRpcMethodMap<T = unknown, D = unknown> = Record<
+  string,
+  JsonRpcMethodHandler<T, D>
+>;
+
+/**
+ * Creates an H3 event handler that implements the JSON-RPC 2.0 specification.
+ *
+ * @param methods A map of RPC method names to their handler functions.
+ * @returns An H3 EventHandler.
+ */
+export function jsonRpcHandler<T = unknown, D = unknown>(
+  methods: JsonRpcMethodMap<T, D>,
+): EventHandler {
+  return defineHandler(async (event: H3Event) => {
+    // JSON-RPC requests must be POST.
+    if (event.req.method !== "POST") {
+      throw new HTTPError({
+        status: 405,
+        message: "Method Not Allowed",
+      });
+    }
+
+    // Helper to construct and return a JSON-RPC error response.
+    const sendJsonRpcError = (
+      id: string | number | null,
+      code: number,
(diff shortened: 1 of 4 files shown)


src/utils/json-rpc.ts
```diff
@@ -0,0 +1,212 @@
...
+  JsonRpcMethodHandler<T, D>
+>;
+
+/**
+ * Creates an H3 event handler that implements the JSON-RPC 2.0 specification.
+ *
+ * @param methods A map of RPC method names to their handler functions.
+ * @returns An H3 EventHandler.
+ */
+export function jsonRpcHandler<T = unknown, D = unknown>(
```
pi0: We probably should prefix name with define

src/utils/json-rpc.ts
```diff
@@ -0,0 +1,212 @@
...
+      const error: JsonRpcError = { code, message };
+      if (data) {
+        error.data = data;
+      }
+      return { jsonrpc: "2.0", id, error };
+    };
+
+    let hasErrored = false;
+    let error = undefined;
+    const body = await readBody<JsonRpcRequest<T> | JsonRpcRequest<T>[]>(
```
pi0: We should use native event.req.json()

src/utils/json-rpc.ts
```diff
@@ -0,0 +1,212 @@
...
+      if (req.jsonrpc !== "2.0" || typeof req.method !== "string") {
+        return sendJsonRpcError(
+          req.id ?? null,
+          INVALID_REQUEST,
+          "Invalid Request",
+        );
+      }
+
+      const { id, method, params } = req;
+      const handler = methods[method];
```
pi0: It seems we might need some builtin protection against proto access

src/utils/json-rpc.ts
```diff
@@ -0,0 +1,212 @@
...
+ * A map of method names to their corresponding handler functions.
+ */
+export type JsonRpcMethodMap<T = unknown, D = unknown> = Record<
+  string,
+  JsonRpcMethodHandler<T, D>
+>;
+
+/**
+ * Creates an H3 event handler that implements the JSON-RPC 2.0 specification.
+ *
```
pi0, line 70: Can you add simple example for rendering in docs?

src/utils/json-rpc.ts
sandros94: with https://github.com/h3js/h3/pull/1180/commits/b25d03479a22b3ab0e9d513f021e19e624e0aee8 I've implemented a simple safeguard for it.

Let me know if I'm on the right track, as I do realize this is a topic which I'm weak in knowledge :/

src/utils/json-rpc.ts
```diff
@@ -0,0 +1,264 @@
...
+
+    // If it was a single notification, send 202 Accepted.
+    if (!isBatch && finalResponses.length === 0) {
+      event.res.status = 202;
+      return "";
+    }
+
+    // For a single request, return the single response object.
+    // For a batch request, return the array of response objects.
+    return isBatch ? finalResponses : finalResponses[0];
```
sandros94:
```suggestion
    return isBatch ? finalResponses : finalResponses[0];
```

@pi0 Would this block an event-stream for non-batch requests? (a final response currently is `{ jsonrpc: "2.0", id, result }`)

Doesn't have a strict approach on this, but MCP servers that do want to support SSE state that at the end of the stream a full jsonrpc response should be provided anyway

I'm having an hard time testing event streams even in normal h3 routes 😬

sandros94: I just found a better approach, much simpler and only relies on `ReadableStream`

sandros94: pushed with https://github.com/h3js/h3/pull/1180/commits/9c19de61618e7ad9a2876dd3535fb245a2c01115, this one can be marked as resolved (I see it as `Pending` and I cannot close it)

pi0:
Hi dear @sandros94 This PR is under my radar (and important!) but have to delay for more testing. In meantime you don't need to keep it updated ❤️

sandros94:
> Hi dear @sandros94 This PR is under my radar (and important!) but have to delay for more testing. In meantime you don't need to keep it updated ❤️ 

Absolutely no problem, I also imagine that we all took some time off for the summer and should focus on more important things first, this is a long term one. This feature is also easy to implement in downstream projects, so no rush.

The reason why I kept updating it for a bit was because while testing it with some Json RPC clients (mainly MCP-related ones) I've found a couple of issues and DX improvements, but now it is complete for what I know atm

src/utils/json-rpc.ts
sandros94: I've never noticed these were left as pending...

sandros94:
@pi0 coderabbit in one of its nitpicks pointed out a nice addition, and since I did already catch `HTTPError`s I decied to properly map them in https://github.com/h3js/h3/pull/1180/commits/1447b9ff8d3c1a171b5d2fb3d258db5231e43a17

I think this should be something that has to be part of h3 internal implementation of the RPC handler and it is spec compliant. I've also tested it locally with LLMs and they do understnad the mapping between `320xx` and `4xx` codes.

WDYT?

EDIT: also handled 3xx status in https://github.com/h3js/h3/pull/1180/commits/8ea2463b96e13a0e620146b79d1c26fce3a2dd96, as they were falling to `32603` which they are not

sandros94:
Development note: with https://github.com/h3js/h3/pull/1180/commits/dc503c003a03687deb20f0f62d1d942a560f3def I've dropped support for ReadableStream, something I've missinterpreted in my original implementation last august.

The JSON-RPC 2.0 spec expect a single, full, JSON object. It is the MCP spec that allows to have a GET endpoint on the same path which optionally supports ReadableStreams (eg: for server-to-client notifications), meaning it is out of scope from current PR

test/json-rpc.test.ts
```diff
@@ -0,0 +1,672 @@
...
+      expect(json).toEqual({
+        error: {
+          code: -32_603,
+          data: "Handler error",
+          message: "Internal error",
+        },
+        id: 4,
+        jsonrpc: "2.0",
+      });
+    });
```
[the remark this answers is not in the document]
sandros94: @pi0 WDYT of this? Should we remove any error information from the response? I have no personal preference on this

PR #1185, opened 2025-08-07
refactor: hide internal `event._res` and `event.res._headers`

huseeiin:
this hides props from the user that they should not use

### DIFF
--- a/src/event.ts
+++ b/src/event.ts
@@ -9,12 +9,14 @@ import type {
 } from "./types/handler.ts";
 import type { H3Core } from "./h3.ts";
 
+const kEventNS = "h3.internal.event.";
+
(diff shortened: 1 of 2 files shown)


pi0:
Thanks for PR. It breaks tests since in response handler we need to access intermediate headers and avoid triggering slow path: https://github.com/h3js/h3/blob/8972b0fc3f24b7405a04b36b16dd5d60d8d31051/src/response.ts#L72

We could use ~ instead.

huseeiin:
> Thanks for PR. It breaks tests since in response handler we need to access intermediate headers and avoid triggering slow path:
> 
> https://github.com/h3js/h3/blob/8972b0fc3f24b7405a04b36b16dd5d60d8d31051/src/response.ts#L72
> 
> We could use ~ instead.

oh this line will error?

PR #1186, opened 2025-08-08
chore(docs): lowercase `srvx`

iiio2:
I think `srvx` is always with lowercase in all letters.


pi0 (approved):
thnx!

iiio2:
Thanks @pi0 . :blush:

Issue #1187, opened 2025-08-14
forwardHeaders and filterHeaders of proxyRequest not working


mihaon:
### Environment

node 22, main branch

### Reproduction

```
return proxyRequest(event, 'http://localhost:8080/, {
	filterHeaders: ['cookie'],
})
```

### Describe the bug

The `!ignoredHeaders.has(name)` line will always be true for non-ignored headers. For example, for "coookie" header.

src/utils/proxy.ts
```
export function getProxyRequestHeaders(
  event: H3Event,
  opts?: {
    host?: boolean;
    forwardHeaders?: string[];
    filterHeaders?: string[];
  },
): Record<string, string> {
  const headers = new EmptyObject();
  for (const [name, value] of event.req.headers.entries()) {
    if (
      opts?.forwardHeaders?.includes(name) ||
      (opts?.filterHeaders && !opts.filterHeaders.includes(name)) ||
      !ignoredHeaders.has(name) || // <-- THIS LINE
      (name === "host" && opts?.host)
    ) {
      headers[name] = value;
    }
  }
  return headers;
}
```

### Logs

```sh

```

PR #1188, opened 2025-08-14
fix(proxy): fix proxy headers filter

kricsleo:
resolves https://github.com/h3js/h3/issues/1187


pi0 (approved):

pi0:
Thanks! Would be nice if we could add some coverage tests though.

PR #1191, opened 2025-08-22
chore: remove `node-ts` script

iiio2:
@pi0 , Actually I have seen your this PR. (https://github.com/nitrojs/nitro/pull/3543/files) 
That's why I thought may be we can do it also here like that.


iiio2:
Thanks @pi0 . :blush:

PR #1192, opened 2025-08-26
feat: normalize reltive URL passed to `H3.request`

huseeiin:
apply same url normalization of on to request for consistency


src/h3.ts
```diff
@@ -48,6 +48,8 @@ export const H3Core = /* @__PURE__ */ (() => {
       _init?: RequestInit,
       context?: H3EventContext,
     ): Response | Promise<Response> {
+      if (typeof _req === "string" && !URL.canParse(_req))
+        _req = new URL(_req, "http://_").pathname;
```
pi0, line 52: This is a heavy operation to do per request! Instead, users should make sure they use sanitized/encoded input. 

As an alternative, we could expose a new utility that does this sanitization for users.

pi0:
~~Not acceptable because of runtime per-request cost. PR welcome to add util.~~ (reopened, thinking more)

pi0:
Checking more, it seems `new Request` used in `toRequest` takes care of url normalziation.

Also added test for `app.request(unicode)` which seems passing https://github.com/h3js/h3/commit/1f874ee5d7f76c3ec28ba212c990b0055046f2a7

huseeiin:
what about the change to `fetchWithEvent` tho:
```ts
export async function fetchWithEvent(
  event: H3Event,
  url: string,
  init?: RequestInit,
): Promise<Response> {
  // start change
  if (URL.canParse(url)) {
    return fetch(url, init);
  }
  // end change
  return event.app!.fetch(
    createSubRequest(event, url, {
      ...init,
      headers: mergeHeaders(
        getProxyRequestHeaders(event, { host: true }),
        init?.headers,
      ),
    }),
  );
}
```
separate pr?

pi0:
global `fetch` and `createSubRequest` also use `new URL` internally that should normalize path 👍🏼 

 It is always best to first make a minimal reproduction before trying to fix issues...

huseeiin:
> global `fetch` and `createSubRequest` also use `new URL` internally that should normalize path 👍🏼
> 
> It is always best to first make a minimal reproduction before trying to fix issues...

sorry but the change to `fetchWithEvent` i mentioned above has nothing to do with normalization, its just the way to detect this is a relative url or not to use globalThis.fetch instead. previously you did `if (url[0] !== "/")`

pi0:
Sure, but what is benefit? We want to explicitly detect relative `/path` inputs. Validation/Normalization happens in the next steps always.

PR #1198, opened 2025-08-31
feat: devcontainer support

iiio2:
actually I am not sure about it. Can we add devcontainer support in h3 also ? I've seen devcontainer in nitro, that's why I thought may be we can add here also like that.


pi0:
Appreciate PR but due to low runtime complexity here, we don't need it. Nitro needed it for cross platform testing, h3 core is agnostic and based on web api.

iiio2:
Ok @pi0 . Thanks.

PR #1199, opened 2025-09-01
chore: import `ProxyOptions` type separately

iiio2:
Actually I have seen we import types separately and function separately . For example, here we have toNodeHandler function is imported separately and NodeHandler type is also imported separately from the same file. That is why I thought we can import ProxyOptions type separately as well here.


pi0 (approved):

iiio2:
Thanks @pi0 . :blush:

PR #1202, opened 2025-09-06
fix: re-export used crossws types

### LINKED ISSUE #716: Re-export websocket types
### Describe the feature

As a developer using websockets, I want to come up with generic helper functions working with websocket peers (such as for #715). However, `h3` doesn't re-export the type of peer, so I have to use introspection types:

```ts
type Peer = Parameters<NonNullable<Parameters<typeof defineWebSocketHandler>[0]["open"]>>[0]
```

Not only this is cumbersome, but also it loses "genericness" of the original type.

I propose to re-export `Peer` directly from `h3` (and some other involved types, e.g. `Message`, possibly others).

### Additional information

rijkvanzanten:
`h3` directly relies on `crossws` for it's WS handling, but doesn't make the `crossws` types available to consumers of `h3`. This in turn makes it quite hard to properly type functions that rely on the returned values in the WS handlers, as it forces the consumer of `h3` to also add `crossws` as a devDependency and ensure that the version of `crossws` exactly matches the version in use by `h3` to ensure there's no type conflicts. This is a pretty poor DX. By re-exporting the types used for the `crossws` logic that's relied on by `h3` this problem is resolved.

I've identified `Hooks`, `Peer`, `AdapterInternal`, and `Message` as the ones you immediately run into when using `defineWebSocketHandler`, but there might be others. I didn't want to go as far as to `export *` as that'll add a bunch of unused types and other noise to `h3`s global scope as well. 

Fixes #716

### DIFF
--- a/src/utils/ws.ts
+++ b/src/utils/ws.ts
@@ -3,6 +3,8 @@ import { defineHandler } from "../handler.ts";
 import type { Hooks as WSHooks } from "crossws";
 import type { EventHandler } from "../types/handler.ts";
 
+export type { Hooks, Peer, AdapterInternal, Message } from "crossws";
+
 /**
  * Define WebSocket hooks.
  *


src/utils/ws.ts
```diff
@@ -3,6 +3,8 @@ import { defineHandler } from "../handler.ts";
 import type { Hooks as WSHooks } from "crossws";
 import type { EventHandler } from "../types/handler.ts";
 
+export type { Hooks, Peer, AdapterInternal, Message } from "crossws";
```
pi0: We need to alias-prefix this types with `WebSocket` from h3

rijkvanzanten: @pi0 Do you want it prefixed with `WebSocket`, or `WS` so it matches the `WSHooks` import on line 3?

rijkvanzanten: Prefixed with `WS` now to match the other `ws.ts` usage, but happy to adjust :)

pi0: `WS` reads a little more cryptic perhaps good time we make types consitent with `WebSocket*` prefix. Are other `WS` exports exported from package publicly?

rijkvanzanten: Oki! I don't believe any other WS prefixed things are exported at the moment no

PR #1211, opened 2025-10-04
build: add missing exports

sandros94:
Just missing some exports for newer stuff like chunked cookies, allowing for downstream use


src/index.ts
```diff
@@ -50,17 +53,24 @@ export {
   toEventHandler,
 } from "./handler.ts";
 
-export { defineMiddleware } from "./middleware.ts";
+export {
+  defineMiddleware,
+  callMiddleware,
+  normalizeMiddleware,
```
pi0: normalize and call are internal. (we can only expose call* method but only when later needed really)

sandros94: Mh, indeed normalize shouldn't be exposted, my bad. While `callMiddleware` did come in handy when debugging stuff.

Will fix once back home

sandros94: Just out of curiosity: what do you mean with "when later needed really"? Once we reach a stable release to decide if we actually want to expose it or not?

Please do feel free to revert it and also not expose `callMiddleware`. My main focus for this PR was chunked cookies and defineWebSocket which I actually need for some things I'm building (auth) and testing

pi0 (approved):

Issue #1204, opened 2025-09-14
toWebRequest is not creating abort signal


IlyaSemenov:
### Environment

- h3@1.15.3
- Node v22.16.0

### Reproduction

The problem is clear by looking at the code — the [signal property](https://developer.mozilla.org/en-US/docs/Web/API/Request/signal) is not created:

https://github.com/h3js/h3/blob/ec77d6bc148e4ff7629ba56577697055cc0fcf2e/src/utils/request.ts#L354-L365

I would have created a demo reproduction with trpc-nuxt but StackBlitz currently has a problem running Nuxt apps, will do once https://github.com/nuxt/nuxt/issues/33120 is fixed.

### Describe the bug

`toWebRequest` is not creating the abort signal (which other libraries depend upon). That makes the code that uses the web requests created from h3 events to stuck forever when e.g. handling subscriptions.

Related issue: https://github.com/wobsoriano/trpc-nuxt/issues/191

### Additional context

I extended `toWebRequest` to pass the `AbortSignal` to work around the issue:

```ts
function toWebRequest(event: H3Event) {
  return event.web?.request || new Request(getRequestURL(event), {
    // @ts-ignore Undici option
    duplex: "half",
    method: event.method,
    headers: event.headers,
    body: getRequestWebStream(event),
    signal: (() => {
      const abortController = new AbortController()
      event.node.req.on("close", () => abortController.abort())
      return abortController.signal
    })(),
  })
}
```

### Logs

```sh

```

pi0:
Thanks for the report. Feel free to open a PR for adding support. 

Only we have to make sure signal is lazy created. Check https://github.com/h3js/srvx/pull/76 as reference

kricsleo:
@IlyaSemenov Do you have any plans to work on this? If not, I'd be happy to fix it ❤️

IlyaSemenov:
Sure, please do. If it were hanging like this for some time more, I could at some point pick it up, but if you're willing to fix this, please go ahead!

productdevbook:
@pi0 This is labeled v1. In v2, `toWebRequest` no longer exists — `event.req` IS already a web standard `Request` object (via srvx). The abort signal comes from the platform's native `Request.signal`.

There's also an open PR #1220 targeting this for v1. Can this be closed as not applicable to v2?

PR #1220, opened 2025-10-13
feat(toWebRequest): add request abort signal

kricsleo:
resolves https://github.com/h3js/h3/issues/1204

### DIFF
--- a/src/utils/request.ts
+++ b/src/utils/request.ts
@@ -365,9 +365,10 @@ export function toWebRequest(event: H3Event) {
       get signal() {
         if (!abortSignal) {
           abortSignal = new AbortController();
-          event.node.req.once("close", () => {
-            abortSignal?.abort();
-          });
+
+          const abort = (err?: any) => abortSignal?.abort(err);
+          event.node.req.once("close", abort);
+          event.node.req.once("end", abort);
         }
         return abortSignal.signal;
       },
--- a/src/utils/request.ts
+++ b/src/utils/request.ts
@@ -352,6 +352,8 @@ export function getRequestURL(
  * **NOTE:** This function is not stable and might have edge cases that are not handled properly.
  */
 export function toWebRequest(event: H3Event) {
+  let abortSignal: AbortController | undefined;
+
   return (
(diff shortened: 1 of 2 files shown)


pi0:
Can you add a test?

kricsleo:
Tests added 😁

Kinfe123 (approved):

src/utils/request.ts
```diff
@@ -360,6 +362,16 @@ export function toWebRequest(event: H3Event) {
       method: event.method,
       headers: event.headers,
       body: getRequestWebStream(event),
+      get signal() {
+        if (!abortSignal) {
+          abortSignal = new AbortController();
+
+          const abort = (err?: any) => abortSignal?.abort(err);
+          event.node.req.once("close", abort);
```
IlyaSemenov, line 370: This listens on the request stream, so it fires when the body has been read, not when the client
disconnects. On a POST with a streaming response the signal is aborted before the handler writes
anything. It happens to work with GET because nothing reads the body.

`srvx` (which is what `h3` v2 works on) already does this correctly in its node adapter:

```ts
  res.once("close", () => {
    if (req.errored) abort(req.errored);
    else if (!res.writableEnded) abort();
  });
```

src/utils/request.ts
```diff
@@ -360,6 +362,16 @@ export function toWebRequest(event: H3Event) {
...
+          event.node.req.once("end", abort);
```
IlyaSemenov, line 371: With proper fix (see above) this is not needed.

test/web.test.ts
```diff
@@ -66,3 +71,63 @@ describe("Web handler", () => {
...
+            controller.close();
+          },
+        });
+      }),
+    );
+
+    const controller = new AbortController();
+    const response = await client.request({
+      path: "/abort",
+      method: "GET",
```
IlyaSemenov, line 121: This should include a test for POST, too; it would surface the problems in the current implementation.

IlyaSemenov (changes requested):

PR #1222, opened 2025-10-15
feat: add `assertBodySize` util and `bodyLimit` middleware

### LINKED ISSUE #1221: Add request body size limit utility and middleware
### Describe the feature

Add utils to limit request body size.

### Additional information

aquapi:
resolves #1221

### DIFF
--- a/src/utils/body.ts
+++ b/src/utils/body.ts
@@ -71,7 +71,7 @@ export async function readValidatedBody<
  * @example
  * function validateBody(body: any) {
  *   return typeof body === "object" && body !== null;
- * };
+ * }
  *
  * app.get("/", async (event) => {
  *   const body = await readValidatedBody(event, validateBody);
--- a/docs/2.utils/1.request.md
+++ b/docs/2.utils/1.request.md
@@ -48,7 +48,7 @@ You can use a simple function to validate the body or use a Standard-Schema comp
 ```ts
 function validateBody(body: any) {
   return typeof body === "object" && body !== null;
-};
+}
 app.get("/", async (event) => {
   const body = await readValidatedBody(event, validateBody);
 });


aquapi:
I need to add tests for this feature as well
We can get some tests from Hono: <https://github.com/honojs/hono/blob/main/src/middleware/body-limit/index.test.ts>

aquapi:
@pi0 do u think I should add more tests

pi0 (approved):
Thanks ❤️

pi0:
Now we can add it a a Nitro v3 route rule even!

src/utils/body.ts
```diff
@@ -98,3 +101,60 @@ export async function readValidatedBody(
...
+): Promise<boolean> {
+  const req = event.req;
+  if (req.body === null) {
+    return true;
+  }
+
+  const bodyLen = req.headers.get("content-length");
+  if (bodyLen !== null && !req.headers.has("transfer-encoding")) {
+    return +bodyLen <= limit;
+  }
```
kaaax0815, line 146: @pi0 see https://github.com/h3js/h3/pull/1150#discussion_r2189728926

by your previous statement this code is insecure

pi0: We should add a flag to force check when no length.

But also test which runtimes allow body with length (invalid) + transfer-encoding (if they do yes indeed we should enable flag by default)

pi0: According to the HTTP/1.1 specification (RFC 7230, section 3.3.2):

A message must not include both a Content-Length header field and a Transfer-Encoding header field.


(we should actually fail in this case)

pi0: https://github.com/h3js/h3/commit/9ccd301cf0234fa351deca56935774a435f858e1

kaaax0815: but it stills allows a faked content-length

pi0: When running in real HTTP server (Node.js server for example) it should stop reading body when content-length excceds as it violated protocol.

image

pi0: (Bun and Deno also "cut off" reading body by validating `content-length`)

image

kaaax0815: Thank you very much

PR #1223, opened 2025-10-17
docs: remove beta tag

iiio2:
recently I used h3 and i did not need to write h3@beta. it automatically install h3 v.2, that is why I've thought maybe we can remove h3@beta now.


iiio2:
Thanks @pi0 .

Issue #1229, opened 2025-10-24
onError returning `new Response(404)`


pi0:
> hi, it seems that this change breaks returning custom 404 from onError handler like this:
> 
> ```
> 
> app.use(onError(event => {
> 
>    if (event.status === 404) {
> 
>       return new Response('Page not found', { status: 404 })
> 
>    }
> 
> }))
> 
> ```  

 _Originally posted by @matej21 in [f16f954](https://github.com/h3js/h3/commit/f16f954553d0d8a54d8cefbd0d7f603b9dabd4ec#r168758403)_

PR #1231, opened 2025-10-25
fix: allow returning custom `Response` with 404 code inside middleware

kricsleo:
resolves https://github.com/h3js/h3/issues/1229


pi0:
Thank you for the PR. It is kinda a wanted behavior since in Nitro for example we allow mounting a hono sub-app as middleware and if it returns `new Response(404)`, rest of the app routing works.

I have added a test to show feature here: https://github.com/h3js/h3/commit/8791818851c9f60daf5fda5e3f35e55ad6dcef4d

Now i'm thinking if we could find a way to have both.

pi0:
OK lets make it explicit for Nitro usecase ~> #1234

PR #1232, opened 2025-10-27
fix: adjust pathname for mounted sub-app routed middleware

### LINKED ISSUE #1208: Mounted middleware from subapp don't inherit base url
### Environment

| Product | Version |
|--------|--------|
| Node | `v22.20.0` |
| h3 | `v2.0.0-beta.4` |

### Reproduction

[TheBrenny/h3-middleware-bug](https://github.com/TheBrenny/h3-middleware-bug)

```js
// index.js
import {H3, serve} from "h3";
import otherRoutes from "./other.js";

const app = new H3();
app.mount("/api", otherRoutes);
serve(app);

export default app;
```

```js
// other.js
import {H3} from "h3";

const routes = new H3();
routes.use("/hello", (event) => {console.log(event)});
routes.get("/hello", () => new Response("world"));

export default routes;
```

### Describe the bug

When mounting a sub-app that has middleware inside, the middleware doesn't actually get mounted to the main app.

**What should happen:** I should expect my middleware to be mounted at the correct endpoint.

### Additional context

_No response_

### Logs

```sh
> node index.js
➜ Listening on: http://localhost:3000/ (all interfaces)
^C
```

(Note the lack of output after listening message)

amondnet:
Fixes middleware not inheriting base URL when sub-app is mounted. Middleware matchers now correctly see adjusted pathname (without base prefix), while route handlers continue to see the full pathname.

This fix adjusts `event.url.pathname` before calling child middleware, then restores it before calling the route handler (via wrapped next callback). The try/finally block ensures pathname is always restored.

Changes:
- Modified H3Core.mount() to adjust pathname for child middleware
- Added wrappedNext callback to restore pathname before route handler
- Added comprehensive tests for middleware mounting scenarios
- Updated existing test to use relative paths in mounted middleware

Test Results:
- All 712 functional tests pass
- Bundle size increased by ~150 bytes (acceptable tradeoff)

resolves #1208

### DIFF
--- a/src/utils/body.ts
+++ b/src/utils/body.ts
@@ -140,9 +140,14 @@ async function isBodySizeWithin(
     return true;
   }
 
-  const bodyLen = req.headers.get("content-length");
-  if (bodyLen !== null && !req.headers.has("transfer-encoding")) {
-    return +bodyLen <= limit;
+  const contentLength = req.headers.get("content-length");
+  if (contentLength) {
+    const transferEncoding = req.headers.get("transfer-encoding");
+    if (transferEncoding) {
+      // https://datatracker.ietf.org/doc/html/rfc7230#section-3.3.2
+      throw new HTTPError({ status: 400 });
+    }
+    return +contentLength <= limit;
   }
 
   const reader = req.clone().body!.getReader();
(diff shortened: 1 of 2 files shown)


pi0 (approved):
Thanks ❤️

PR #1235, opened 2025-10-27
build: reduce dist size by stripping comments

pi0:
Comments are already extracted to dts files.

Stripping them from `.mjs` bundle saves ~30kB!


productdevbook:
When will we reach zero kb? :D

Issue #715, opened 2024-03-21
Router params inside websocket handler


IlyaSemenov:
### Describe the feature

I am using websockets in a Nuxt app, and I put my event handler in a parameter-based route, such as `server/api/foo/[id]/ws.ts`

I would like to be able to easily access `id` in websocket hooks, similar to how I can access it in "normal" event handlers with `getRouterParam`.

The code may look like:

```ts
export default defineWebSocketHandler({
  open(peer) {
    console.log("ID is", getPeerRouterParam(peer, "id")) // <--- imaginary code, doesn't work.
  },
}
```

### Additional information

- [X] Would you be willing to help implement this feature?

dardanbujupaj:
I had the same issue when trying to use websockets in nitro.

I proposed a PR #1242 which solved this, by allowing a function as param for `defineWebSocketHandler`, which provides the `event` of the request (similar to defineEventHandler).


This would deviate from the proposed (in the issue description) usage like this:

```js
export default defineWebSocketHandler((event) => ({
  open() {
    console.log("ID is", getRouterParam(event, "id"))
  }
}))
```

@IlyaSemenov would that work for you as well? (sorry for pinging you after 1+ year, if you're not interested in this anymore, you can close the issue and I'll open another one)

IlyaSemenov:
Yes, that would work for me.

My first impression was it's overcomplicated, as it seemed to be using callbacks just for the sake of capturing the original event, but on second thought it actually made sense not only because of handling the original params (and the original query, cookies, etc.); but also possibly returning a different set of callbacks (e.g., unauthenticated requests wouldn't even return the `message` callback.)

I am not sure how it would work with non-Node adapter (Cloudflare with hibernation). I am not saying it wouldn't, just curious here.

PR #1242, opened 2025-11-16
feat(defineWebSocketHandler): support callback with `event` 

dardanbujupaj:
This PR adds an alternative parameter type for `defineWebsocketHandler`, so that we can access `event` when defining crossws hooks.
This allows us to use methods like `getRouterParam(event, name)` similar to http handlers.
It also resolves https://github.com/h3js/h3/issues/715.

I saw that there's also https://github.com/h3js/h3/pull/1149 open, which might relate to this. But I think this should _just work_ together. 

## Example
Use router param to subscribe/publish to a topic.

```js
import { defineWebSocketHandler, H3, serve, getRouterParam } from "h3-patched";
import { plugin as ws } from "crossws/server";

const app = new H3();

app.get(
  "/ws/:room",
  defineWebSocketHandler((event) => {
    const room = getRouterParam(event, "room") as string;

    return {
      open(peer) {
        console.log(`Peer ${peer.id} connected to room ${room}`);
        peer.subscribe(room)
      },
      message(peer, message) {
        console.log(`Received message from ${peer.id} for room ${room}: ${message.data}`);
        peer.publish(room, message.data);
      },
    };
  }),
);

serve(app, {
  plugins: [ws({ resolve: async (req) => (await app.fetch(req)).crossws })],
});

```

## Summary by CodeRabbit

* **New Features**
  * `defineWebSocketHandler` now accepts hooks as either static configuration or as a dynamic function that resolves hooks per request.

* **Documentation**
  * Updated API documentation to reflect enhanced parameter flexibility.

* **Tests**
  * Added test coverage for function-based hooks configuration.

### DIFF
--- a/docs/2.utils/99.community.md
+++ b/docs/2.utils/99.community.md
@@ -26,3 +26,9 @@ This section is placeholder for any new H3 version 2 compatible community librar
 [H3ravel Framework](https://h3ravel.toneflix.net) is a modern TypeScript runtime-agnostic web framework built on top of H3, designed to bring the elegance and developer experience of Laravel PHP to the JavaScript ecosystem. See the getting started guide [here](https://h3ravel.toneflix.net/guide/get-started).
 
 :read-more{to="https://github.com/h3ravel"}
+
+## `Intlify`
+
+[Intlify](https://intlify.dev/) is a project that aims to improve Developer Experience in software internationalization. That project provides server-side frameworks, middleware, and utilities. About those, see the [here](https://github.com/intlify/srvmid)
+
+:read-more{to="https://github.com/intlify/srvmid"}
--- a/package.json
+++ b/package.json
@@ -48,45 +48,45 @@
     "h3": "link:."
   },
   "dependencies": {
-    "rou3": "^0.7.9",
-    "srvx": "^0.9.1"
+    "rou3": "^0.7.10",
+    "srvx": "^0.9.7"
   },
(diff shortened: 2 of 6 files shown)


dardanbujupaj:
There's also [`defineWebSocket`](https://github.com/h3js/h3/pull/1242/files#diff-eef453364a051df059b927a5b836b233a4d34fc75b903d25c4c93ffb6438982eL17), but I wasn't sure about the use case for that, so I left it for now.
Glad for inputs about that 🙃

src/utils/ws.ts
```diff
@@ -26,15 +27,19 @@ export function defineWebSocket(
  * @see https://h3.dev/guide/websocket
  */
 export function defineWebSocketHandler(
-  hooks: Partial<WebSocketHooks>,
+  hooks:
+    | Partial<WebSocketHooks>
+    | ((event: H3Event) => Partial<WebSocketHooks>),
```
pi0: It could also return a promise for async resolution

dardanbujupaj: good point, missed that... fixed.

they-cloned-me:
Thank you, @dardanbujupaj @pi0  😍
I had a question about something like this (although in Nitro) two years ago:
[https://github.com/nitrojs/nitro/discussions/2605](https://github.com/nitrojs/nitro/discussions/2605)

PR #1260, opened 2025-12-11
fix: consume ReadableStream in serverless environments

onmax:
`sendStream()` returns early when `!event.node.res.socket` (serverless) without consuming the stream, causing empty responses and wrong Content-Type headers.

Fix: consume stream via `pipeTo()` and write data to response.

Includes tests that fail before fix, pass after.

### DIFF
--- a/src/utils/response.ts
+++ b/src/utils/response.ts
@@ -430,7 +430,8 @@ export function sendStream(
         });
     }
     // Fallback: store stream directly for environments that can handle it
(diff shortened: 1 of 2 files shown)


pi0:
Can you please provide a full reproduction/ and specify deployment platform? (if it is from nitro v2, we might do workaround per preset)

onmax:
hey, while trying to create a minimal reproduction, i found a workaround :)

PR #1262, opened 2025-12-12
fix(handleCacheHeaders): round `modifiedTime` to seconds

### LINKED ISSUE #1261: `handleCacheHeaders` compares modifiedTime with millisecond precision causing false mismatches
### Environment

nuxt4 / node22

### Reproduction

```js
const stat = fs.statSync(filePath)
handleCacheHeaders(event, { modifiedTime: stat.mtime })
// stat.mtime = 2025-01-01T12:00:00.123Z
// Last-Modified sent = Wed, 01 Jan 2025 12:00:00 GMT
// If-Modified-Since received = Wed, 01 Jan 2025 12:00:00 GMT
// Comparison fails because of .123 ms
```

### Describe the bug

`handleCacheHeaders` receives a `modifiedTime` value (typically from fs.statSync().mtime or Date from an `updateAt` field), which includes millisecond precision.
However, when setting the Last-Modified header, h3 formats the date without milliseconds (as required by HTTP date standards).

This results in the following issue:

- First request → handleCacheHeaders sends a Last-Modified header (rounded to seconds)
- Client sends back If-Modified-Since containing the rounded timestamp
- On subsequent calls, h3 compares the client’s second-precision timestamp with the server’s millisecond-precision modifiedTime

The values never match, causing unnecessary 200 responses instead of 304

***Suggested fix***

when comparing timestamps, truncate or round modifiedTime to seconds, e.g.:
`const normalized = new Date(Math.floor(modifiedTime.getTime() / 1000) * 1000)` or `setMilliseconds(0)` ?

### Additional context

_No response_

### Logs

```sh

```

amondnet:
## Summary

This PR fixes issue #1261 where `handleCacheHeaders` compares `modifiedTime` with millisecond precision, causing cache validation to always fail when the source date has non-zero milliseconds.

### The Problem

1. Server's `modifiedTime` retains millisecond precision (e.g., `2025-01-01T12:00:00.123Z`)
2. `Last-Modified` header is formatted without milliseconds per HTTP spec (e.g., `Wed, 01 Jan 2025 12:00:00 GMT`)
3. Client sends back `If-Modified-Since` with the truncated timestamp
4. Comparison fails: `1735732800000 >= 1735732800123` is `false`

This causes unnecessary 200 responses instead of 304, wasting bandwidth and server resources.

### The Fix

Truncate milliseconds from `modifiedTime` before both setting the header and making the comparison:

```typescript
const modifiedTime = new Date(opts.modifiedTime);
modifiedTime.setMilliseconds(0);  // Added line
```

### Test

Added a test case that specifically uses a date with milliseconds to verify the fix works correctly.

## Test Plan

- [x] Existing tests pass
- [x] New test case added for millisecond precision handling
- [x] `pnpm test` passes

Closes #1261

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced cache header timestamp handling for consistent comparison logic.

* **Tests**
  * Added tests to verify cache header timestamp precision.

### DIFF
--- a/src/utils/event.ts
+++ b/src/utils/event.ts
@@ -15,7 +15,7 @@ export function isEvent(input: any): input is H3Event {
 /**
  * Checks if the input is an object with `{ req: Request }` signature.
  * @param input - The input to check.
- * @returns True if the input is is `{ req: Request }`
+ * @returns True if the input is `{ req: Request }`
  */
 export function isHTTPEvent(input: any): input is HTTPEvent {
   return input?.req instanceof Request;


pi0 (approved):
Thanks!

PR #1284, opened 2026-01-20
fix(response): in errors prevent propagating corrupting headers

sandros94:
While doing debetable stuff I found myself breaking a few things, but turned out to be headers that were propagating into my error responses, minimal repro would be:

```ts
import { H3, serve, HTTPError } from "h3";

const app = new H3();

app.get("/", () => {
  throw new HTTPError("Error", {
    headers: {
      "content-length": "9999",
    }
  });
});

serve(app);
```

The current `errorResponse` would blidnly merge any headers coming from `error.headers` or `error.cause.headers` and could easily endup breaking the response altogether.

Proposed fix is to create a small set of denied headers which do not get merged in the process, tho this increased bundle size about 84bytes

## Summary by CodeRabbit

* **Bug Fixes**
  * Error responses now filter out dangerous headers, preventing unintended header propagation in error scenarios.

* **Improvements**
  * Enhanced header handling and configuration in error response processing for more robust behavior.

### DIFF
--- a/src/response.ts
+++ b/src/response.ts
@@ -8,6 +8,14 @@ import { kEventRes, kEventResHeaders, type H3Event } from "./event.ts";
 export const kNotFound: symbol = /* @__PURE__ */ Symbol.for("h3.notFound");
 export const kHandled: symbol = /* @__PURE__ */ Symbol.for("h3.handled");
 
+const DISALLOWED_ERROR_HEADERS = /* @__PURE__ */ new Set([
+  "content-length",
+  "content-type",
+  "content-encoding",
+  "content-language",
+  "transfer-encoding",
+]);
+
 export function toResponse(
   val: unknown,
   event: H3Event,
@@ -144,9 +152,12 @@ function mergeHeaders(
   base: HeadersInit,
   overrides: Headers,
   target = new Headers(base),
+  ignoreSet = new Set(),
 ): Headers {
   for (const [name, value] of overrides) {
-    if (name === "set-cookie") {
+    if (ignoreSet.has(name.toLowerCase())) {
+      continue;
+    } else if (name === "set-cookie") {
       target.append(name, value);
     } else {
       target.set(name, value);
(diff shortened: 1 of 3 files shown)


pi0:
Can you please explain more about how this can happen in realworld? There are undoubtedly ways to make invalid responses, but adding logic to core for extra validation costs.

If it is mainly on proxy utils, we can handle it there.

sandros94:
> Can you please explain more about how this can happen in realworld? There are undoubtedly ways to make invalid responses, but adding logic to core for extra validation costs.
> 
> If it is mainly on proxy utils, we can handle it there.

For example when directly returning a fetch-based utility that returns the upstream error response with its headers.

While writing a real-world reproduction based on the closed source code where I've noticed this I realized that I did this myself in [one of my personal projects](https://github.com/sandros94/uns3/blob/b48846fc2af83e6e132785154d4efe25db4bfe64/src/client.ts#L1099). In this later case I could simply fix it in my source code (afterall the content-type will always be xml), but for the issue at work I wasn't really sure how to explain proper web standards (maybe didn't search enough related topics) or why it would be better to manually manage these errors before reaching client side.

Taking in example my own lib e reproduction would be:

```ts
app.delete("/:id", (event) => {
  return s3Client.del(event.context.params?.id);
});
```

pi0:
Someone might want to rightfully return an error response with custom transform/content headers or size.

I think any usage that returns a response or throws an error with custom headers should be returning valid headers (in your case you defenitly need to wrap those and h3 proxy util does same as well).

sandros94:
> Someone might want to rightfully return an error response with custom transform/content headers or size.

For personal knowledge and understanding, could you point me to an example of such case? (if you have time)

> I think any usage that returns a response or throws an error with custom headers should be returning valid headers (in your case you defenitly need to wrap those and h3 proxy util does same as well).

Ok, in my own lib I'll update it like so today. At work I'll make sure to redirect the issue to the related team and properly address this in the upstream utility

pi0:
> For personal knowledge and understanding, could you point me to an example of such case? (

It might be for example to stream SSR response for error into chunks.

PR #1295, opened 2026-02-05
refactor(auth): enhance randomJitter function for cryptographic security

sandros94:
I love the idea of the jitter! Tho the `Math.random` is technically predictable in this type of situations, using the WebCrypto API has the same perfomance (on modern hardware and runtimes) and more secure 😬

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced authentication security by upgrading the retry mechanism to use cryptographically secure random value generation instead of standard randomization.


pi0 (approved):
thnx ❤️

PR #1298, opened 2026-02-06
feat: mcp tools

HugoRCD:
> Edited by @pi0 and Claude

## Summary

Adds native [Model Context Protocol](https://modelcontextprotocol.io) (MCP) support to h3, built directly on top of h3's JSON-RPC 2.0 infrastructure — no external SDK dependency required.

### Highlights

- `defineMcpHandler()` — creates an H3 handler implementing MCP over HTTP (JSON-RPC 2.0 wire format)
- `defineMcpTool()` / `defineMcpResource()` / `defineMcpPrompt()` — typed definition helpers for MCP primitives
- Supports: `initialize`, `ping`, `tools/list`, `tools/call`, `resources/list`, `resources/read`, `prompts/list`, `prompts/get`, and notifications
- Dynamic per-request options via `defineMcpHandler((event) => ({ ... }))`
- Fully stateless — method map built per request, no server instances or transport adapters
- Reuses h3's existing `processJsonRpcBody` for full JSON-RPC 2.0 compliance (batch, notifications, error codes)

### API

```ts
import { defineMcpHandler, defineMcpTool, defineMcpResource, defineMcpPrompt } from "h3";

const echoTool = defineMcpTool({
  name: "echo",
  description: "Echo back a message",
  inputSchema: {
    type: "object",
    properties: { message: { type: "string" } },
    required: ["message"],
  },
  handler: async (args, event) => ({
    content: [{ type: "text", text: args.message as string }],
  }),
});

app.all(
  "/mcp",
  defineMcpHandler({
    name: "my-server",
    version: "1.0.0",
    tools: [echoTool],
  }),
);
```

## Test plan

- [x] All 31 MCP tests pass (`pnpm vitest run test/mcp.test.ts`)
- [x] JSON-RPC tests unaffected (70 tests pass)
- [x] Full test suite passes (965 tests, 44 files)
- [x] Typecheck passes (`pnpm typecheck`)


src/utils/internal/mcp.ts
```diff
@@ -0,0 +1,168 @@
+import type {
+  Transport,
+  TransportSendOptions,
+} from "@modelcontextprotocol/sdk/shared/transport.js";
+import type { JSONRPCMessage, RequestId } from "@modelcontextprotocol/sdk/types.js";
+import type { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
+import type { H3Event } from "../../event.ts";
+import type { McpHandlerOptions } from "../mcp.ts";
+import { readBody } from "../body.ts";
```
pi0: in v2, we can simply use web API `event.res.json`()`

package.json
```diff
@@ -92,11 +93,15 @@
     "zod": "^4.3.6"
   },
   "peerDependencies": {
+    "@modelcontextprotocol/sdk": "^1.25.0",
```
pi0: Need to get rid of dep! https://github.com/h3js/h3/pull/1180 is start i think /cc @sandros94

sandros94:
First off, lets address the elephant in the room: the [`@modelcontextprotocol/sdk`](https://npmx.dev/package/@modelcontextprotocol/sdk) library is humongous compared to [`h3`](https://npmx.dev/package/h3).

A required step IMHO is having support for the native JsonRPC handler, which all modern MCPs are based off. I'll soon revisit #1180 PR and see if it needs something to be addressed. Regardless of MCPs this would also enable other use-cases, since it is well-known spec and vastly adopted (especially when mixing up programming languages).

But in reality the SDK reaches that size because of the following reasons, which are part of the MCP specs:
- authorization is mostly based on JWT (either direct token or via OAuth/OIDC), this is something I've started digging deep (and ended up loving) almost one year ago https://github.com/h3js/h3/issues/994
- built-in validation, not only for project-level information/data but even before accepting the incoming RPC call. It is quite hardcoded ontop of zod, its own toJsonSchema and ajv for validating it. Which they also come with their tradeoffs, not only in bundle size (you can read more about it [here](https://schemabenchmarks.dev))
- the SDK itself is meant to be ready-to-use, not really a toolset to _build-ontop_, meaning that it comes with hono, express and various adapters.

There are another couple of aspects which I do not remember atm, but even just for the first two points above I would personally consider IF h3 is even the right place.

Even with forking the SDK, bringing over the bare minimum, we would still need:

- JWT reader and validator: [`jose`](https://npmx.dev/package/jose) or [`unjwt`](https://npmx.dev/package/unjwt) (ofc I prefer the latter, but ended up being similar size and performance, just the DX is close to what h3 offers)
- Validation library agnostic: [StandardSchemaV1](https://github.com/standard-schema/standard-schema) support, which we have, and [StandardJSONSchema](https://github.com/standard-schema/standard-schema/pull/134) which not only arrived recently but also requires dedicated linking for each validation library
- A set of cryptographically secure utilities: for example to manage PKCE challanges, string comparison, randomness and so on ends up adding its own size, even forking a small library like [`unsecure`](https://npmx.dev/package/unsecure)
- JsonRPC handler

Only at this point, once we have all these tools available (forked or added as deps) we can start building the various functions and constants, and from the experience in [`h3-mcp-tools`](https://npmx.dev/package/h3-mcp-tools) (which I really miss and want to get back and develop it) I can confidently say that the final h3 size will at lest double, realistically triple, the current state.

My honest though would be to add this tooling in Nitro, which can treeshake stuff when unused anyway.

sandros94:
Also, for these very reason I'm considering to just close the JWE issue and related PR (https://github.com/h3js/h3/pull/997). Iron Crypto is not my personal favorite in its topic, but the current implementation is good-enough as is, and too many people already use it.

While on the other hand JWE (and JWS) would require to be as spec-compatible as possible, which inevitable end up increasing h3 size. Not to mention that the current `useSession` implementation is not compatible with the OAuth spec, which I would argue that it is the most important spec to consider when working with JWTs

This is why I built, maintain and use [`unjwt`](https://npmx.dev/package/unjwt). JWT compliant, h3 DX

pi0:
Thanks for your thoughts @sandros94 ❤️

Yes, lets ship json rpc utils first!

With composable approach of h3 utils, i think we can easily make authenticated wrappers for MCP tools (which can live outside of h3 as ready to use opinionated solutions). And as for validation supporting standard schema just makes sense and comes for free.

H3 provides many treeshakable utils already. Users in Nitro can benefit from auto scanning of server/mcp/* tools + h3 as base or better wrapper packages that wrap base tools.

PR #1300, opened 2026-02-10
chore(deps): upgrade crossws to v0.4.4 and align websocket

typed-sigterm:
crossws 0.3.5 seems not to work with bun 1.3.9

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
   },
   "dependencies": {
     "cookie-es": "^1.2.2",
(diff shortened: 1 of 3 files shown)


pi0:
crossws 0.4 has breaking changes h3 v1 cannot easily adopt.

Please consider upgrading to h3 v2 (or nitro v3) for best bun compatibility.

typed-sigterm:
is there a way to use h3 v2 with Nuxt?

PR #1306, opened 2026-03-01
fix(ws): expose crossws on app.fetch response

### LINKED ISSUE #1258: Type mismatch occurred when using `crossws`
### Environment

h3: 2.0.1-rc.5

### Reproduction

https://h3.dev/guide/advanced/websocket

### Describe the bug

Here is the `app.fetch`'s type

https://github.com/h3js/h3/blob/d6475e131b3a80f44bebd5b95ce3e106ed1ac258/src/types/h3.ts#L106

```ts
fetch(_request: ServerRequest): Response | Promise<Response>; 
```

But when we use `defineWebsocketHandler`, the function returns `crossws` as an extra result

https://github.com/h3js/h3/blob/d6475e131b3a80f44bebd5b95ce3e106ed1ac258/src/utils/ws.ts#L23C1-L41C2

```ts
/**
 * Define WebSocket event handler.
 *
 */
  hooks: Partial<WebSocketHooks>,
): EventHandler {
  return defineHandler(function _webSocketHandler() {
    return Object.assign(
      new Response("WebSocket upgrade is required.", {
        status: 426,
      }),
      {
        crossws: hooks,
      },
    );
  });
}
```

ahmetguness:
When using `defineWebSocketHandler`, the returned `Response` includes a `crossws` property.  
However, `app.fetch()` is typed as returning `Response`, so `res.crossws` is not visible in TypeScript without casting.

- Introduces a small `H3Response` type (`Response` with optional `crossws`)
- Updates `fetch` / `toResponse` to return `H3Response`
- Removes `(res as any).crossws` from tests
- Adds type tests to make sure this stays correct

All checks pass locally (`lint`, `typecheck`, `vitest`).

Closes #1258

## Summary by CodeRabbit

* **New Features**
  * Introduced H3Response (Response with optional WebSocket hooks) and WebSocketResponse for safer WebSocket responses.
  * Core request/response APIs and WebSocket handler creator updated to use the new response types and improved generics.
  * Re-exported the new response types from the public API surface.

* **Tests**
  * Added and updated type tests validating H3Response, WebSocketResponse and WebSocket handler typings.


pi0:
Thanks for PR. I think it would not be best idea to introduce a new `H3Response` type just to cover `.crossws`. Instead, `defineWebSocketHandler` return type could be improved.

Closing PR for maintenance as it already has lots of merge conflicts. But please feel free to open a fresh one for scoped type augmentation.

PR #1311, opened 2026-03-08
build: bundle docs as skill + `h3 docs`

pi0:
Docs discovery for LLMs is important. This PR uses an experimental approach of bundling cleaned-up docs (using [mdzilla](https://github.com/pi0/mdzilla)) into `skills/h3` with only TOC (README.md) eagerly loaded into context.

Also adding `npx h3 docs` (installs mdzilla on demand only when needed so additional install costs unless used)

image

## Summary by CodeRabbit

* **New Features**
  * Added H3 documentation command enabling users to explore comprehensive framework reference documentation directly from the command line, with fallback support for web-based documentation

* **Chores**
  * Build process now automatically generates and maintains documentation files during each compilation phase
  * Updated package configuration to ensure generated documentation is properly included in all published distributions

### DIFF
--- a/bin/h3.mjs
+++ b/bin/h3.mjs
@@ -1,6 +1,29 @@
 #!/usr/bin/env node
+import { execSync } from "node:child_process";
 import { main } from "srvx/cli";
 import meta from "../package.json" with { type: "json" };
+
+// Docs
+if (process.argv[2] === "docs") {
+  const runner = [
+    ["bun", "x"],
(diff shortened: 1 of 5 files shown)


gautemo:
The skills folder is 3x the size of the rest of the npm package, https://packagephobia.com/result?p=h3
Just asking if it is intentional and the 1.2 MB _navigation.json file is needed? 😊

PR #1319, opened 2026-03-14
fix(mount): restore pathname on error with try/finally

productdevbook:
## Summary

- The \`mount\` middleware modified \`event.url.pathname\` for child middleware but only restored it inside the \`wrappedNext\` closure
- If a child middleware threw an error, the pathname remained in its stripped form (e.g. \`/test\` instead of \`/api/test\`) when the error handler or subsequent middleware accessed \`event.url.pathname\`
- Now uses \`try/finally\` to ensure pathname is always restored, consistent with how \`withBase\` handles the same pattern

## Test plan

- [x] All existing mount tests pass (18/18)
- [x] New test verifies pathname is restored after middleware throws
- [x] Total: 20/20 passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Mounted middleware now reliably restores the original request pathname after execution, including when middleware returns without calling next, throws errors, or returns asynchronous results.

* **Tests**
  * Added tests that verify pathname restoration for middleware that returns early, throws synchronously, and throws asynchronously.


src/h3.ts
```diff
@@ -121,16 +121,20 @@ export const H3 = /* @__PURE__ */ (() => {
...
-          this["~middleware"].push((event, next) => {
+          this["~middleware"].push(async (event, next) => {
             const originalPathname = event.url.pathname;
             if (!originalPathname.startsWith(base)) {
               return next();
             }
             event.url.pathname = event.url.pathname.slice(base.length) || "/";
-            return callMiddleware(event, input["~middleware"], () => {
+            try {
+              return await callMiddleware(event, input["~middleware"], () => {
```
pi0: For performance reasons, we should avoid async/await (in microbenchmarks it adds up)

src/h3.ts
```diff
@@ -121,16 +121,20 @@ export const H3 = /* @__PURE__ */ (() => {
     mount(base: string, input: FetchHandler | FetchableObject | H3Type) {
       if ("handler" in input) {
         if (input["~middleware"].length > 0) {
-          this["~middleware"].push((event, next) => {
+          this["~middleware"].push(async (event, next) => {
```
pi0: Adding async is unacceptable

pi0 (approved):

src/h3.ts
```diff
@@ -127,10 +127,22 @@ export const H3 = /* @__PURE__ */ (() => {
...
-              return next();
-            });
+              throw err;
+            }
+            if (typeof (result as PromiseLike<unknown>)?.then === "function") {
+              return (result as Promise<unknown>).finally(() => {
+                event.url.pathname = originalPathname;
+              });
+            }
+            return result;
```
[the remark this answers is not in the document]
pi0: @productdevbook check this

pi0:
Thanks!

PR #1335, opened 2026-03-15
feat(session): support password rotation via Record<string, string>

### LINKED ISSUE #1050: FR: Allow multiple unseal passwords for sessions
### Describe the feature

When rotating secrets for use as session passwords, we'd want to continue honoring prior sealed sessions.

For the `useSession` api, can we have an additional parameter: `extraUnsealPasswords` or something similar. Passwords to attempt are the default `password` and this `extraUnsealPasswords`.

Alternatively, we can have `password` as potentially being a non empty string array, with the first `password` being used for sealing, and the rest just for unsealing.

### Additional information

productdevbook:
## Summary

Closes #1050

Enables session password rotation by allowing `SessionConfig.password` to accept `Record<string, string>` in addition to a plain string.

```ts
const session = await useSession(event, {
  password: {
    // "default" key matches sessions sealed with plain string password
    default: "old-password-at-least-32-chars-long!!",
    // first key is used for sealing new sessions
    v2: "new-password-at-least-32-chars-long!!",
  },
});
```

### How it works

- **Sealing**: Uses the first key in the record (with its ID embedded in the token)
- **Unsealing**: iron-crypto already supports `PasswordHash` (Record of passwords keyed by ID) — it reads the `passwordId` from the sealed token and looks up the matching password
- **Backward compat**: Sessions sealed with a plain `string` password use passwordId `"default"` internally, so the old password should use `"default"` as its key

### Changes

- `SessionConfig.password`: `string` → `string | Record<string, string>`
- `sealSession`: extracts first key from record as `{ id, secret }` for sealing
- `unsealSession`: passes record directly to `unseal()` (already supported)
- ~10 lines of implementation change

## Test plan

- [x] Password rotation test: seal with old string password, unseal with rotated Record
- [x] All existing session tests pass (14/14)
- [x] Typecheck clean

## Summary by CodeRabbit

* **New Features**
  * Introduced session password rotation capability. Users can now securely rotate credentials without interrupting service. The system supports multiple passwords and automatically handles unsealing of existing sessions with prior credentials, enabling seamless transitions between old and new passwords for enhanced security management.

### DIFF
--- a/src/utils/session.ts
+++ b/src/utils/session.ts
@@ -29,8 +29,13 @@ export interface SessionManager<T extends SessionDataT = SessionDataT> {
 }
 
 export interface SessionConfig {
-  /** Private key used to encrypt session tokens */
-  password: string;
+  /**
+   * Private key used to encrypt session tokens.
+   *
+   * For password rotation, pass a record of `{ id: password }` pairs.
+   * The first key is used for sealing new sessions, all keys are tried for unsealing.
+   */
+  password: string | Record<string, string>;
   /** Session expiration time in seconds */
   maxAge?: number;
   /** default is h3 */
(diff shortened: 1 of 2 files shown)


src/utils/session.ts
```diff
@@ -200,7 +205,12 @@ export async function sealSession<T extends SessionData = SessionData>(
   const session: Session<T> =
     (context.sessions?.[sessionName] as Session<T>) || (await getSession<T>(event, config));
 
-  const sealed = await seal(session, config.password, {
+  const sealPassword =
+    typeof config.password === "string"
+      ? config.password
+      : { id: Object.keys(config.password)[0], secret: Object.values(config.password)[0] };
```
pi0, line 211: we are only takin first!

PR #1237, opened 2025-11-03
feat: enhance validation capabilities for routes and handlers

productdevbook:
## Summary

cc: https://github.com/h3js/h3/issues/1088

This PR enhances H3's validation capabilities with comprehensive type-safe validation for routes and handlers:

- **Route params validation**: Added support for validating route parameters (e.g., `/users/:id`) with full type inference
- **Response validation**: Server-side response validation to ensure handlers return correctly shaped data
- **Enhanced `defineValidatedHandler`**: Now supports `params` and `response` schemas in addition to existing `body`, `headers`, and `query` validation
- **Enhanced `defineRoute`**: Full type-safe validation with proper TypeScript overloads for better DX
- **New utilities**: Exported `syncValidate` for synchronous validation and `validateResponse` for response validation
- **Comprehensive tests**: Added full test coverage including integration tests demonstrating complete validation workflows

### 1. Params Validation
```typescript
const handler = defineValidatedHandler({
  validate: {
    params: z.object({ id: z.string().uuid() })
  },
  handler: (event) => {
    // event.context.params is typed as { id: string }
    const id = event.context.params.id;
    return { userId: id };
  }
});
```

### 2. Response Validation
```typescript
const route = defineRoute({
  method: 'POST',
  route: '/api/users',
  validate: {
    response: z.object({ id: z.string(), name: z.string() })
  },
  handler: async (event) => {
    // Return type is validated and type-checked
    return { id: '123', name: 'John' };
  }
});
```

### 3. Complete Validation
```typescript
defineRoute({
  method: 'POST',
  route: '/api/users/:id',
  validate: {
    params: z.object({ id: z.string().uuid() }),
    query: z.object({ include: z.string().optional() }),
    body: z.object({ name: z.string(), email: z.string().email() }),
    response: z.object({ id: z.string(), name: z.string(), email: z.string() })
  },
  handler: async (event) => {
    // All types are inferred: params, query, body, and return type
    const { id } = event.context.params;
    const body = await event.req.json();
    return { id, name: body.name, email: body.email };
  }
});
```

## Test Plan

- [x] Unit tests for params validation in `defineValidatedHandler`
- [x] Unit tests for response validation in `defineValidatedHandler`
- [x] Integration tests for `defineRoute` with all validation schemas
- [x] Error handling tests for invalid params, query, body, headers, and response
- [x] Type inference tests demonstrating full type safety
- [x] Existing tests continue to pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

PR #1339, opened 2026-03-15
feat: add getPayload and getValidatedPayload utilities

### LINKED ISSUE #785: getPayload helper
### Describe the feature

Would be nice I think if h3 had a helper function like `getPayload`

Would work just like `getQuery` and `readBody` but would work for both depending on the request method and also include the route params

Here is a proof of concept for `getPayload` as a utility function:
https://stackblitz.com/edit/github-dbe8gu?file=server%2Futils%2Findex.ts

(uses Nuxt getRouterParams... not sure what that would be in h3)

Would also be great to have a `getValidatedPayload` version that works like the query and body counterparts.

### Additional information

productdevbook:
## Summary

Closes #785

Adds `getPayload` and `getValidatedPayload` utilities that merge route params, query params, and body into a single object — depending on the request method.

As pi0 noted: "Idea is that a utility to work with both query/body."

### `getPayload(event)`

```ts
// GET /search/books?q=h3 → merges route params + query
app.get("/search/:category", async (event) => {
  const payload = await getPayload(event);
  // { category: "books", q: "h3" }
});

// POST /users/123 with body { name: "Alice" } → merges route params + body
app.post("/users/:id", async (event) => {
  const payload = await getPayload(event);
  // { id: "123", name: "Alice" }
});
```

### `getValidatedPayload(event, schema)`

```ts
app.post("/users/:id", async (event) => {
  const payload = await getValidatedPayload(event, z.object({
    id: z.string(),
    name: z.string(),
  }));
  // Validated and typed
});
```

### Behavior

| Method | Sources merged |
|--------|---------------|
| GET, HEAD | route params + query params |
| POST, PUT, PATCH, DELETE | route params + parsed body |

Route params have lowest priority — body/query values override them on conflict.

## Test plan

- [x] GET returns query merged with route params
- [x] POST returns body merged with route params
- [x] Body overrides route params on conflict
- [x] getValidatedPayload validates with zod schema
- [x] getValidatedPayload throws 400 on invalid data
- [x] All tests pass in web + node (14/14)
- [x] Typecheck clean, formatting clean

## Release Notes

* **New Features**
  * Introduced payload utilities for extracting and validating request data. Query parameters, request bodies, and route parameters are automatically merged with configurable priority. Built-in support for validating against standard schemas or custom validators with type-safe results.

* **Tests**
  * Added comprehensive test coverage for new payload utilities.

### DIFF
--- a/src/utils/payload.ts
+++ b/src/utils/payload.ts
@@ -1,7 +1,11 @@
 import type { H3Event, HTTPEvent } from "../event.ts";
+import type { ErrorDetails } from "../error.ts";
+import type { StandardSchemaV1, FailureResult, InferOutput } from "./internal/standard-schema.ts";
+import type { ValidateResult, OnValidateError } from "./internal/validate.ts";
 import { getQuery } from "./request.ts";
 import { readBody } from "./body.ts";
 import { getRouterParams } from "./request.ts";
+import { validateData } from "./internal/validate.ts";
 
 const _payloadMethods = new Set(["PATCH", "POST", "PUT", "DELETE"]);
 
@@ -36,4 +40,36 @@ export async function getPayload<T = Record<string, unknown>>(
   }
   const query = getQuery(event);
   return { ...params, ...query } as T;
+}
+
+/**
+ * Get and validate the request payload using a Standard Schema or custom validator.
+ *
+ * @example
+ * app.post("/users/:id", async (event) => {
+ *   const payload = await getValidatedPayload(event, z.object({
+ *     id: z.string(),
+ *     name: z.string(),
+ *   }));
+ * });
+ */
+export function getValidatedPayload<Event extends HTTPEvent, S extends StandardSchemaV1<any, any>>(
+  event: Event,
+  validate: S,
+  options?: { onError?: (result: FailureResult) => ErrorDetails },
+): Promise<InferOutput<S>>;
+export function getValidatedPayload<Event extends HTTPEvent, OutputT>(
+  event: Event,
+  validate: (
+    data: Record<string, unknown>,
+  ) => ValidateResult<OutputT> | Promise<ValidateResult<OutputT>>,
+  options?: { onError?: () => ErrorDetails },
+): Promise<OutputT>;
+export async function getValidatedPayload(
+  event: H3Event | HTTPEvent,
+  validate: any,
+  options?: { onError?: OnValidateError },
+): Promise<any> {
+  const payload = await getPayload(event);
+  return validateData(payload, validate, options);
 }
--- a/src/utils/payload.ts
+++ b/src/utils/payload.ts
@@ -1,0 +1,39 @@
+import type { H3Event, HTTPEvent } from "../event.ts";
+import { getQuery } from "./request.ts";
+import { readBody } from "./body.ts";
+import { getRouterParams } from "./request.ts";
+
+const _payloadMethods = new Set(["PATCH", "POST", "PUT", "DELETE"]);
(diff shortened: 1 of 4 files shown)


src/utils/payload.ts
```diff
@@ -0,0 +1,75 @@
...
+ * });
+ */
+export async function getPayload<T = Record<string, unknown>>(
+  event: H3Event | HTTPEvent,
+  opts?: { decode?: boolean },
+): Promise<T> {
+  const params = getRouterParams(event, opts);
+  if (_payloadMethods.has(event.req.method)) {
+    const body = (await readBody(event)) || {};
+    return { ...params, ...(typeof body === "object" ? body : { body }) } as T;
```
pi0, line 39: We need an object with null proto

src/utils/payload.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+  const query = getQuery(event);
```
pi0, line 41: We can use `event.url` to get query instead of getQuery

src/utils/payload.ts
```diff
@@ -0,0 +1,75 @@
...
+ * app.get("/search/:category", async (event) => {
+ *   const payload = await getPayload(event);
+ *   // { category: "books", q: "h3" } — category from route, q from query
+ * });
+ */
+export async function getPayload<T = Record<string, unknown>>(
+  event: H3Event | HTTPEvent,
+  opts?: { decode?: boolean },
+): Promise<T> {
+  const params = getRouterParams(event, opts);
```
pi0, line 36: router params should not be included. only query+body

src/utils/payload.ts
```diff
@@ -0,0 +1,75 @@
...
+  if (_payloadMethods.has(event.req.method)) {
+    const body = (await readBody(event)) || {};
```
pi0, line 38: we should use event.req.json()

src/utils/payload.ts
danielkellyio: in my original issue #785, I did include route params. This way has been most useful for me. Any downsides?

src/utils/payload.ts
```diff
@@ -0,0 +1,75 @@
...
+ * Route params take lowest priority (body/query overrides them).
+ *
+ * @example
+ * app.post("/users/:id", async (event) => {
+ *   const payload = await getPayload(event);
+ *   // { id: "123", name: "Alice" } — id from route, name from body
+ * });
+ *
+ * @example
+ * app.get("/search/:category", async (event) => {
```
danielkellyio, line 27: It's also been useful in my experience to include query params for post requests, so that order of importance goes:
1. body
2. query params
3. route params

For example a post request to:
/users/:id?email=test@test.com

with the body:
{ name: "Alice" } 

would result in:

const { id, name, email } = await getPayload(event);

Why?
Mostly just when I'm moving fast and mix up query and body. But the idea being the payload is just everything that we pass as data to the backend

danielkellyio (changes requested):
just a couple comments based on my usage of a similar project specific utility. Thanks!

pi0x:
**Note: this analysis was performed by an AI agent (Claude Code) on behalf of the maintainer.**

Flagging for context: despite being from the same author as #1237 and touching a related area (validation), this PR is not a duplicate or subset of #1237 — it has zero file overlap and solves a different problem (unifying query/body/params access into a single call, vs. #1237's full route params/response validation pipeline). Recommending both be reviewed independently rather than closing one in favor of the other.

PR #1362, opened 2026-03-29
fix(path): handle malformed encoded pathnames without throwing

### LINKED ISSUE #1361: decodePathname throws URIError on malformed percent-encoded pathname
### Environment

Node v22.22.0

macos/linux

### Reproduction

https://stackblitz.com/edit/node-57djiu9v?file=index.js

`node index.js`

### Describe the bug

decodePathname can throw URIError when the pathname contains malformed percent-encoded bytes (for example incomplete or invalid UTF-8 sequences). This happens during H3Event construction, so the request fails before normal app-level handling can run. This was a breaking change between rc 16 and rc 19 (we stuggle a bit upgrading tanstack https://github.com/TanStack/router/pull/7044 because of it)

### Additional context

Adding a `try/catch` block might be sufficient here to return original pathname on decode failure:

```
  try {
    return decodeURI(
    );
  } catch {
    return pathname;
  }
}
```

### Logs

```sh

```

birkskyum:
resolves #1361 
- https://github.com/h3js/h3/issues/1361

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of malformed URL paths to prevent errors and ensure stability when encountering invalid path encodings.

* **Tests**
  * Added unit tests to verify URL decoding behavior and preservation of encoded sequences, reducing regression risk.


src/utils/internal/path.ts
```diff
@@ -56,7 +56,11 @@ export function getPathname(path: string = "/"): string {
  * Decode percent-encoded pathname, preserving %25 (literal `%`).
  */
 export function decodePathname(pathname: string): string {
-  return decodeURI(pathname.includes("%25") ? pathname.replace(/%25/g, "%2525") : pathname);
+  try {
+    return decodeURI(pathname.includes("%25") ? pathname.replace(/%25/g, "%2525") : pathname);
+  } catch {
+    return pathname;
```
pi0, line 62: Wouldn't this impose a security risk surface with incosistent path handling? I suggest making silent fallback behavior opt-in with a h3 global config like `strictPaths: false` and pass to decode util if really necessary.

birkskyum:
closed by https://github.com/TanStack/router/pull/7140

PR #1363, opened 2026-04-02
fix: handle URIError in decodePathname for malformed percent-encoding (#1361)

### LINKED ISSUE #1361: decodePathname throws URIError on malformed percent-encoded pathname

guoyangzhen:
### 🔗 Linked issue

Fixes #1361

### 📝 Description

`decodePathname` calls `decodeURI()` which throws `URIError` on malformed percent-encoded paths (e.g., incomplete sequences like `%XX` or invalid UTF-8). Since this runs during `H3Event` construction, the request fails before any app-level error handling can run.

This was a breaking change between rc.16 and rc.19 that blocked TanStack Router's upgrade ([TanStack/router#7044](https://github.com/TanStack/router/pull/7044)).

### 🔄 Changes

Wrap the `decodeURI` call in a try/catch block that returns the original pathname on decode failure:

```typescript
  try {
    return decodeURI(
      pathname.includes('%25') ? pathname.replace(/%25/g, '%2525') : pathname,
    );
  } catch {
    return pathname;
  }
}
```

This is the same approach suggested by the issue reporter and matches the defensive pattern used elsewhere in the codebase.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved path handling logic to correctly process edge cases involving path normalization.
  * Enhanced URL decoding reliability with better handling of encoded characters.

* **Chores**
  * Internal utilities strengthened for more robust path segment resolution.

### DIFF
--- a/src/utils/internal/path.ts
+++ b/src/utils/internal/path.ts
@@ -37,7 +37,7 @@ export function withoutBase(input: string = "", base: string = ""): string {
     return input;
   }
   const _base = withoutTrailingSlash(base);
-  if (!input.startsWith(_base)) {
+  if (!input.startsWith(_base) || (input.length > _base.length && input[_base.length] !== "/")) {
     return input;
   }
   const trimmed = input.slice(_base.length);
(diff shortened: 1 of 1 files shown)


pi0:
#1362

PR #1368, opened 2026-04-16
perf(cookie): avoid quadratic chunked cookie parsing and header rebuilds

trivikr:
### Summary

This PR improves cookie performance in the chunked-cookie path used by sessions.

Previously, `setCookie()` rebuilt and reparsed the full `set-cookie` header list on every write. Since `setChunkedCookie()` calls `setCookie()` for the main cookie and for each chunk, large chunked session writes became quadratic in chunk count. The read path had a similar issue: `getChunkedCookie()` repeatedly called `getCookie()`, which reparsed the full request cookie header each time.

### What Changed

* Cache parsed request cookies per `Headers` instance in `src/utils/cookie.ts`
* Update `getChunkedCookie()` to read from a single parsed cookie map instead of reparsing per chunk.
* Keep response-side `set-cookie` state in memory so unique cookie writes can append without reparsing existing headers.
* Preserve deduplication behavior for overwrites by rebuilding headers only when the same logical cookie key is replaced.

## Summary by CodeRabbit

* **Refactor**
  * Improved cookie handling performance by caching parsed request cookies and response Set-Cookie state to reduce repeated parsing.
  * Optimized chunked cookie reads to parse the request cookie header once.
  * Updated Set-Cookie behavior to update headers incrementally when possible.
* **Tests**
  * Added performance-focused unit tests for chunked cookie utilities (single parse on read; no extra parsing when appending).
  * Extended coverage to ensure parse results aren’t affected by mutations of returned cookie objects and that externally added Set-Cookie values aren’t removed by later calls.

### DIFF
--- a/src/utils/cookie.ts
+++ b/src/utils/cookie.ts
@@ -11,6 +11,18 @@ const CHUNKED_COOKIE = "__chunked__";
 // The limit is approximately 4KB, but may vary by browser and server. We leave some room to be safe.
 const CHUNKS_MAX_LENGTH = 4000;
 
+interface SetCookieState {
+  cookies: string[];
+  keys: Array<string | undefined>;
+  distinctKeys: Set<string>;
+}
+
+const requestCookiesCache = new WeakMap<
+  Headers,
+  { source: string; cookies: Record<string, string | undefined> }
+>();
+const responseCookiesCache = new WeakMap<Headers, SetCookieState>();
+
 /**
  * Parse the request to get HTTP Cookie header string and returning an object of all cookie name-value pairs.
  * @param event {HTTPEvent} H3 event or req passed by h3 handler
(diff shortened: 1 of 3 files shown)


pi0x:
Reviewed this branch and found two correctness issues in the new caching, both now fixed with regression tests (on the integration branch) — the core optimization is preserved (0 `parseSetCookie` on the append path, 1 `parseCookie` per read).

**1. `set-cookie` written outside `setCookie` is silently dropped (data loss).** Once `setCookie` ran, the `SetCookieState` WeakMap became the sole source of truth. Any `set-cookie` appended to `event.res.headers` elsewhere (middleware, proxy, direct `append`) was invisible, and the dedup path rebuilt the header list from cached state only — dropping the external cookie on the next overwrite.

```js
setCookie(event, "a", "1");
event.res.headers.append("set-cookie", "x=9"); // external writer
setCookie(event, "a", "2");                     // overwrite
// this branch: [ 'a=2; Path=/' ]         ← x=9 lost
// main:        [ 'x=9', 'a=2; Path=/' ]   ← preserved
```

Fix: `_getSetCookieState` reconciles the cache against live `getSetCookie()` each call — reused untouched on the all-through-h3 path (still zero `parseSetCookie`), rebuilt (re-parsing only unknown cookies) when headers diverge.

**2. `parseCookies` returned the shared cached object.** Previously each call returned a fresh record; callers mutating the result now corrupted the memoized parse for every subsequent read on that request. Fix: return a shallow copy (negligible vs. re-parsing).

Minor/no-fix: unparseable `set-cookie` entries are now *preserved* rather than dropped during dedup — a behavior change from `main` but a no-data-loss improvement; worth a note in the description.

Added regression tests for both (external-append-survives-overwrite; mutation-doesn't-corrupt-cache) — both fail before the fix, pass after; full cookie + perf suites green.

---
*This review and the accompanying fixes were generated with AI assistance (Claude) and reviewed by a human before posting.*

pi0:
Thanks for this PR! Landing via #1472 alternatively with less changes.

PR #1372, opened 2026-04-21
feat(error): coerce thrown number/string to HTTPError

onmax:
Before, throwing a 404 needed `throw new HTTPError({ status: 404 })` or the deprecated `createError`. This coerces thrown numbers at the rejection boundary in `toResponse`, so this now works:

```ts
throw 404
```

`throw new HTTPError({...})` / `throw new Error()` keep working unchanged. Returning a number from a handler is unchanged (still JSON-serialized).

Net bundle impact: ~0 (stays under existing limits).

Addresses #1371.

I know the likelihood of getting this merged is low, but fun to explore :)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of numeric values in promise rejections to properly set HTTP error status codes in asynchronous operations.

### DIFF
--- a/src/response.ts
+++ b/src/response.ts
@@ -16,7 +16,7 @@ export function toResponse(
   if (typeof (val as PromiseLike<unknown>)?.then === "function") {
     return (val as Promise<unknown>).then(
       (resolvedVal) => toResponse(resolvedVal, event, config),
-      (rejection) => toResponse(coerceThrown(rejection), event, config),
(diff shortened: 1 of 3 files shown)


pi0 (approved):
I personally really like the idea! Sharing in discord if there are real concerns against it (`throw 404` vs `return 404` might be confusing pattern)

pi0:
Lets ship it!

PR #1376, opened 2026-04-25
fix(event-stream): do not emit unhandled rejection when readable side is canceled

dyeske61283:
Per WHATWG Streams, cancelling a TransformStream's readable errors its writable with the same reason. writer.closed therefore rejects with the string 'resource closed'. 

H3's .then(() => { this._writerIsClosed = true }) has no .catch, so the rejection bubbles to a top-level unhandled promise rejection and Deno terminates the process. 

H3 does correctly .catch(...) the per-write rejections (this._writer.write(...).catch(() => { this._writerIsClosed = true })), so the design intent is clearly to swallow these — it just forgot the same .catch on .closed. Adding one to that .then would fix this regardless of runtime.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved event-stream closure handling so the stream-writer's closed state is recognized whether closure succeeds or fails, reducing unhandled promise rejections.
  * Makes downstream checks for writer closure more reliable, preventing spurious runtime errors when streams close unexpectedly.
  * No changes to public APIs; behavior is more robust without altering interfaces.

### DIFF
--- a/src/utils/internal/event-stream.ts
+++ b/src/utils/internal/event-stream.ts
@@ -21,9 +21,11 @@ export class EventStream {
   constructor(event: H3Event, opts: EventStreamOptions = {}) {
     this._event = event;
     this._writer = this._transformStream.writable.getWriter();
-    this._writer.closed.then(() => {
-      this._writerIsClosed = true;
-    });
+    this._writer.closed
+      .then(() => {
+        this._writerIsClosed = true;
+      })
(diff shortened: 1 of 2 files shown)


pi0:
Can you please add a regression test? 🙏🏼

dyeske61283:
Done @pi0  :)

pi0 (approved):
thnx!

PR #1385, opened 2026-05-11
fix(writeEarlyHints): normalize Link key to prevent hanging with Node.js

### LINKED ISSUE #1383: `writeEarlyHints` never resolves with node runtime
### Environment

h3 2.0.1-rc.22
node v24.15.0

### Reproduction

```js
// server.mjs

	return "⚡️ Tadaa!";
});

serve(app, { port: 3000 });
```

### Describe the bug

### Additional context

_No response_

### Logs

```sh

```

mixelburg:
### What

Fixes `writeEarlyHints` never resolving (hanging indefinitely) when called with a `Link` header using a non-lowercase key (e.g., `{ Link: '...' }`).

Closes #1383.

### Why

Node.js's native `res.writeEarlyHints(hints, cb)` only reads `hints.link` (lowercase). If `link` is missing or undefined, it returns **early without invoking the callback**. Since HTTP headers are case-insensitive, users naturally pass `{ Link: '...' }` — causing the h3 function to hang forever because the Promise never resolves.

### How

Before calling the native `writeEarlyHints`:
1. Normalize any `Link` / `LINK` / etc. key to lowercase `link`
2. If no `link` property exists after normalization, return immediately (no need for a Promise)

### Reproduction

```js
// This hangs forever before the fix:
  await writeEarlyHints(event, { Link: "</style.css>; rel=preload; as=style" });
  return "ok";
});
```

### Root Cause

Node.js source (`lib/_http_server.js`):
```js
function writeEarlyHints(hints, cb) {
  if (hints.link === null || hints.link === undefined) {
    return; // returns without calling cb!
  }
  // ...
  this._writeRaw(head, 'ascii', cb);
}
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved early hints handling by normalizing `link`/`Link` variations, dropping falsy/empty entries, and preventing hangs when no valid `link` hints are provided. The CDN/web fallback now emits only the collected, truthy `link` header values.
* **Tests**
  * Added regression coverage to ensure `writeEarlyHints` resolves promptly (including `{}` and empty `link`/`Link` cases), and verified correct header output when mixed values are provided and when both `link` and `Link` must be merged.


pi0x:
**Note: this analysis was performed by an AI agent (Claude Code) on behalf of the maintainer.**

This looks like the most correct fix for the `writeEarlyHints` hanging bug (#1383) among the three open PRs addressing it (this one, #1387, #1415) — it addresses the actual root cause (Node's native `writeEarlyHints` doesn't invoke its callback when the `link` key is missing/empty) rather than unconditionally dropping the await, so it preserves the "wait until flushed" guarantee for the success path. I've closed #1387 and #1415 as superseded by this one.

Would you be open to adding a regression test that reproduces the exact #1383 case (`writeEarlyHints(event, {})`)? That would make this easy to merge with confidence.

pi0x:
Pushed `1e43dbb` to this branch on behalf of the maintainer.

**Disclosure:** this commit was authored by an AI agent (Claude Code / Claude Opus), not a human, and pushed via the maintainer-can-modify permission on this PR. Please review it as you would any change.

Thanks @mixelburg for the accurate root-cause diagnosis — that Node's native `res.writeEarlyHints()` reads the exact lowercase `link` key and returns *without* invoking its callback when the resolved link value is missing/empty, leaving h3's wrapping promise (and the request) pending forever. That analysis is what made the fix straightforward.

What this commit adds on top of your fix:

1. **The regression test that was requested.** The existing `writeEarlyHints` test was `skipIf(target === "node")`, so the exact Node path this PR fixes had zero coverage. The new tests run in the `describeMatrix` node target and race `writeEarlyHints` against a short timer inside a real Node handler, so a hang surfaces as `"timeout"` rather than silently passing. I verified they are non-tautological: against the pre-fix code they fail with `expected 'timeout' to be 'resolved'` for empty hints, empty link value, and a capital-only `Link` key; against the fix they pass.

2. **A small hardening of the normalization** (addressing the double-key point CodeRabbit raised): every case-variant of `link` (e.g. both `link` and `Link`) is now merged into a single value instead of the last-processed key clobbering the others, and empty strings / empty arrays are treated as "no link" (matching Node's own callback-skipping behavior) so those inputs no longer hang either.

`pnpm vitest run test/utils.test.ts` (115 passed) and `pnpm lint` (0 errors) are green.

pi0 (approved):

PR #1386, opened 2026-05-13
fix: set explicit content-type for Uint8Array responses

### LINKED ISSUE #1158: bun sets default `content-type` for Uint8Array
### Environment

using h3 v2 directly from repo "./src/index.ts" and bun 1.2.18

### Reproduction

```ts
import { H3, serve } from "./src/index.ts";

const app = new H3();

app.get("/", (e) => {
  return new TextEncoder().encode("hello");
});

console.log(
  await fetch(serve(app).url!).then((r) => r.headers.get("content-type")),
);
```

### Describe the bug

application/octet-stream is unexpected. when opening in a browser it downloads a file.

### Additional context

in node it doesn't download and content-type is `null`

### Logs

```sh

```

mixelburg:
## Description

When returning a `Uint8Array` from an h3 handler, the `content-type` header was not set by h3. This caused inconsistent behavior across runtimes:

- **Node.js**: no `content-type` set → browsers default to `text/html` and display the content
- **Bun**: automatically sets `application/octet-stream` → browsers trigger file download instead of displaying

## Solution

Explicitly set `content-type: application/octet-stream` for `Uint8Array` responses when no `content-type` is already set. This ensures consistent behavior across runtimes and allows users to override by setting the header explicitly before returning.

```ts
app.get("/", (e) => {
  e.res.headers.set("content-type", "text/plain");
  return new TextEncoder().encode("hello");
});
```

## Changes

- `src/response.ts`: Added `content-type: application/octet-stream` header for Uint8Array responses (guarded by `!event.res.headers.has("content-type")`)

## Testing

- All 1148 functional tests pass
- Bundle size tests fail due to added bytes (expected for any code addition)

Fixes #1158

## Summary by CodeRabbit

* **New Features**
  * Added an `allowMalformedURL` option to allow malformed percent-encoded URL paths, using the raw (undecoded) pathname instead of rejecting with a 400.

* **Bug Fixes**
  * Fixed an issue where binary responses could be returned without a `Content-Type` header (now defaults to `application/octet-stream`).
  * Improved Early Hints `link` handling and prevented hangs.
  * Fixed Basic Auth colon parsing, ETag matching, CORS `vary` merging, static caching time handling, and SSE pushes after close.

* **Documentation**
  * Updated API docs for the new option and expanded session/auth guidance, plus added community documentation.

* **Chores**
  * Updated dependency versions and workspace/tooling configuration.


pi0 (approved):

pi0x:
Thanks for the PR and the clear writeup. I tested this across runtimes and I don't think we should set `application/octet-stream` here. A few reasons:

**Node *and* Deno leave it unset — Bun is the outlier, not Node.** Pure runtimes, no h3, binary body, no explicit content-type:

| Runtime | binary body → `content-type` |
| --- | --- |
| Node (`res.end(buffer)`) | *(unset)* |
| Deno (`new Response(Uint8Array)`) | *(unset)* |
| Bun (`new Response(Uint8Array)`) | `application/octet-stream` |

So the description's "Node vs Bun" framing is really "Bun vs everyone else."

**Unset is the spec-faithful behavior.** The Fetch Standard's [extract a body](https://fetch.spec.whatwg.org/#concept-bodyinit-extract) algorithm deliberately assigns *no* Content-Type to a `BufferSource` body — unlike string → `text/plain;charset=UTF-8`, `URLSearchParams` → `application/x-www-form-urlencoded`, `Blob` → its `.type`. Node and Deno follow this; Bun injects a type the spec doesn't call for. This change normalizes everyone to Bun's non-standard behavior instead of the Fetch default. (Note: even at the raw `new Response(bytes)` level, all three runtimes report `null` — Bun only adds it at its serving layer.)

**It contradicts how we already treat strings.** We intentionally *don't* set a content-type for string bodies (`// Default header is text/plain we don't set it for performance reasons`) and let the platform default apply. Hardcoding one for `Uint8Array` breaks that "let the platform decide" consistency.

**`application/octet-stream` is a pessimistic guess that often makes UX *worse*.** An unset type lets the browser sniff and usually display inline; `application/octet-stream` forces a download in most browsers. Plenty of people return a `Uint8Array` that has a *known* type (pre-encoded UTF-8 text, an in-memory PNG, …) — we can't guess that, and octet-stream is the worst-case assumption. RFC 9110 §8.3 does name `application/octet-stream` as the value a *recipient* MAY assume when Content-Type is absent — but that's the recipient's decision, and baking it in also suppresses sniffing for cases where sniffing was the desired behavior.

If a specific type matters, the handler can set it explicitly — which this PR already respects via the `!has("content-type")` guard, and that's the right place for the decision to live.

I'd rather keep h3 aligned with the Fetch default (unset) and, if anything, document that binary responses should carry an explicit content-type when it matters. The Bun inconsistency is real, but the fix for "runtimes disagree" shouldn't be "match the one that diverges from the spec." Leaning towards not merging this — open to being convinced otherwise if there's a concrete case where octet-stream is clearly the right default.

PR #1387, opened 2026-05-15
fix: do not await writeEarlyHints callback in Node.js

### LINKED ISSUE #1383: `writeEarlyHints` never resolves with node runtime

ruguoba:
Fixes #1383

## Problem

In Node.js v24,  does not call the callback when passed an empty hints object (or possibly in other cases). This causes the returned Promise to never resolve, and the request handler hangs indefinitely.

## Root Cause

The original code wrapped  in a Promise:

If the callback is never called, the Promise never resolves.

## Fix

Since early hints are fire-and-forget by nature, we now call  without wrapping it in a Promise:

This prevents the handler from hanging while still sending the early hints.

## Summary by CodeRabbit

* **Refactor**
  * Streamlined early hints handling for Node.js runtime.

[Review Change Stack](https://app.coderabbit.ai/change-stack/h3js/h3/pull/1387)

### DIFF
--- a/src/utils/response.ts
+++ b/src/utils/response.ts
@@ -114,9 +114,8 @@ export function writeEarlyHints(
 ): void | Promise<void> {
   // Use native early hints if available (Node.js)
   if (event.runtime?.node?.res?.writeEarlyHints) {
-    return new Promise((resolve) => {
-      event.runtime?.node?.res?.writeEarlyHints(hints, () => resolve());
-    });
+    event.runtime.node.res.writeEarlyHints(hints);
+    return;
   }
 
   // Fallback: Set Link headers for CDN support (only Link headers to avoid leaking sensitive headers)


pi0x:

This PR and #1415 both address the same `writeEarlyHints` hanging bug reported in #1383, but #1385 diagnoses the actual root cause (Node's native `writeEarlyHints` never invokes its callback when the `link` key is missing/empty) rather than unconditionally dropping the await. Dropping the await, as this PR does, fixes the hang but also removes the "wait until flushed" guarantee for the success path.

Closing in favor of #1385. Thanks for digging into this — the repro and diagnosis helped confirm the root cause.

PR #1391, opened 2026-05-23
fix(serve-static): check the response (not request) for an existing content-length

francisjohnjohnston-web:
## What

In `serveStatic` (`src/utils/static.ts`), the `content-length` response header is set from `meta.size`
only when one isn't already present. That guard checks the **request** headers:

```ts
if (meta.size !== undefined && meta.size > 0 && !event.req.headers.get("content-length")) {
  event.res.headers.set("content-length", meta.size + "");
}
```

Every sibling guard in the same function checks the **response** headers instead — `content-type`,
`content-encoding`, and `last-modified` are all set only `if (!event.res.headers.get(...))`. This brings
`content-length` into line:

```ts
if (meta.size !== undefined && meta.size > 0 && !event.res.headers.get("content-length")) {
```

## Why (asymmetric — easy to miss in review)

A GET/HEAD request almost never carries a `content-length` of its own, so
`!event.req.headers.get("content-length")` is **"accidentally true"** in normal use and the derived size
is still applied — the function looks correct in every ordinary test. The bug only surfaces in two edge
cases:

1. A `content-length` deliberately set by the caller via `options.headers` is **silently overwritten** by
   `meta.size` (the request guard never sees the response header that's already there).
2. A request that *does* carry a `content-length` (e.g. behind a proxy, or a GET with a body) wrongly
   **suppresses** the response `content-length`.

Because the common path is unaffected, a reviewer scanning the logic sees `event.req.headers` and
`event.res.headers` a line apart and the slip reads as intentional — until the three sibling guards are
lined up next to it.

## Safety / blast radius

- One-token change to internal logic; no public API, signature, or routing change.
- Behaviour-preserving on the common path (no pre-set response `content-length` ⇒ identical result; all
  four existing `content-length` assertions in `test/static.test.ts` stay green).
- Brings `content-length` into line with the function's own "set only when absent" contract.

## Testing

- Added a regression test: an explicit response `content-length` (via `options.headers`) must be
  preserved, not overwritten by `meta.size`. It **fails on the current code** (returns the derived `18`
  instead of the explicit `999`) on both the `web` and `node` matrix targets, and passes with the fix.
- Full suite green with the fix: **1152 passed / 14 skipped, 0 type errors** (`vitest run` + `tsc --noEmit`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved static file responses to preserve an explicitly set `content-length` header instead of recalculating it.

* **Tests**
  * Added coverage ensuring `HEAD` responses keep a configured `content-length` value unchanged.


pi0x (approved):

pi0 (approved):

PR #1392, opened 2026-05-24
refactor(validate): drop always-true `if (validate.body)` guard in body proxy

francisjohnjohnston-web:
## What

Removes an always-true `if (validate.body)` guard inside the lazy body-validation
`Proxy` created by `validatedRequest`, flattening one level of nesting and a dead
branch.

## Why it's safe / why the guard is redundant

`validatedRequest` returns the **unwrapped** request when `validate.body` is unset:

```ts
if (!validate.body) {
  return req;
}
// Proxy created here — reached only when validate.body is truthy
return new Proxy(req, { get(_t, prop) { if (validate.body) { … } } });
```

So by the time the `Proxy`'s `get` handler runs, `validate.body` is always truthy
— the inner `if (validate.body)` can never be false. The handler already relies on
this (it uses the non-null assertion `validate.body!` inside). Removing the guard
is behaviour-preserving and lets the body-prop checks sit one level shallower.

## Why a reviewer might not have caught it

The redundant guard reads as ordinary defensiveness; it's only visible as dead once
you trace that the enclosing function early-returns for the `!validate.body` case,
so the proxy is conditionally constructed.

## Scope

Internal helper only — no public API, signature, or behaviour change. All 38
`validate` tests pass; `tsc --noEmit`, `oxlint`, and `oxfmt` are clean.

## Release Notes

* **Bug Fixes**
  * Improved request body validation error handling with more consistent enforcement of the `.json()` method for accessing request bodies.

* **Performance**
  * Optimized request validation proxy handler by removing unnecessary runtime checks, reducing overhead for request processing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/h3js/h3/pull/1392?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/utils/internal/validate.ts
+++ b/src/utils/internal/validate.ts
@@ -106,30 +106,31 @@ export function validatedRequest<
   // Create proxy for lazy body validation
   return new Proxy(req, {
     get(_target, prop: keyof ServerRequest) {
-      if (validate.body) {
-        if (prop === "json") {
-          return function _validatedJson() {
-            return req
-              .json()
-              .then((data) => validate.body!["~standard"].validate(data))
-              .then((result) => {
-                if (result.issues) {
-                  throw createValidationError(
-                    validate.onError?.({ _source: "body", ...result }) || {
-                      message: VALIDATION_FAILED,
-                      issues: result.issues,
-                    },
-                  );
-                }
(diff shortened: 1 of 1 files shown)


pi0x (approved):

PR #1394, opened 2026-05-24
fix(serve-static): compare If-Modified-Since at whole-second precision

francisjohnjohnston-web:
## What
`serveStatic` now compares the `If-Modified-Since` request value against the file mtime **truncated to whole-second precision**, matching the granularity of the `Last-Modified` response header it emits.

## Why
`Last-Modified` is serialized via `Date.toUTCString()`, which has whole-second granularity. The conditional-request check, however, compared the client's `If-Modified-Since` against the **full-precision** `mtime`. When a file's mtime carries sub-second milliseconds, a conditional request echoing back the server's own `Last-Modified` value never satisfied `ifModifiedSince >= mtime`, so the server returned a full **200** instead of **304** — silently defeating conditional-request caching for any asset whose mtime is not exactly on a second boundary (the common case on many filesystems).

## Safety
Behaviour-preserving for whole-second mtimes. The only change is that sub-second-mtime assets now correctly return `304` on a matching conditional request. This mirrors the existing whole-second handling already used in `handleCacheHeaders` (`cache.ts`), so the two paths are now consistent.

## Testing
Added "Handles cache (if-modified-since) for sub-second mtime" to `test/static.test.ts` (both web and node targets). It fails on the current code (`expected 200 to deeply equal 304`) and passes after the fix. Full test suite green (**1154 passed / 14 skipped**), typecheck clean, oxlint clean on changed files.

## Impact
Restores correct `304 Not Modified` responses for conditional requests on static assets with sub-second mtimes — fixing avoidable full-body re-transfers.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed timestamp precision handling in HTTP conditional requests to ensure proper cache validation. Timestamps are now normalized to second-level granularity, preventing mismatches when clients check if cached resources remain valid.

[Review Change Stack](https://app.coderabbit.ai/change-stack/h3js/h3/pull/1394?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


pi0x (approved):

pi0 (approved):
Thanks!

PR #1398, opened 2026-05-25
fix(response): resolve writeEarlyHints when Node callback is omitted

### LINKED ISSUE #1383: `writeEarlyHints` never resolves with node runtime

wyf027:
## Summary
Prevent `await writeEarlyHints()` from hanging Node.js request handlers when the native callback is never invoked.

## Problem
Fixes #1383. Handlers like:

```js
return "ok";
```

never completed because `ServerResponse.writeEarlyHints()` does not reliably call its callback (reproducible with `{}` and with Link hints on Node 24).

## Fix
- Return early for empty hint objects (no-op)
- Resolve the returned promise via `queueMicrotask` when the native callback is missing
- Guard with a settled flag so a late callback cannot double-resolve

## Test plan
- [x] `pnpm test test/utils.test.ts` — 108 passed
- [x] New node-runtime test covers empty hints and Link hints without hanging

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of early hints processing with enhanced error handling and fallback mechanisms.
  * Early hints now properly handles empty configurations.

* **Tests**
  * Added test coverage for early hints functionality in Node.js environments.

[Review Change Stack](https://app.coderabbit.ai/change-stack/h3js/h3/pull/1398?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/utils/response.ts
+++ b/src/utils/response.ts
@@ -114,8 +114,28 @@ export function writeEarlyHints(
 ): void | Promise<void> {
   // Use native early hints if available (Node.js)
   if (event.runtime?.node?.res?.writeEarlyHints) {
+    if (Object.keys(hints).length === 0) {
+      return;
+    }
     return new Promise((resolve) => {
-      event.runtime?.node?.res?.writeEarlyHints(hints, () => resolve());
+      let settled = false;
+      const done = () => {
+        if (settled) {
+          return;
+        }
+        settled = true;
+        resolve();
+      };
+
+      try {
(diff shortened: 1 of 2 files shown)


pi0:
Too complex.

PR #1406, opened 2026-06-03
fix(event): don't crash the process on a malformed request URI

davidz627:
### 🔗 Linked issue

Refs #1361 (and the previously-closed #1363).

### 📝 Description

`decodePathname` calls `decodeURI`, which throws a `URIError: "URI malformed"` on invalid percent-encoding — a bare `%`, or an invalid UTF-8 byte sequence such as `%C0`:

```ts
  return decodeURI(
  );
}
```

It's called from the `H3Event` constructor:

```ts
if (url.pathname.includes("%")) {
  url.pathname = decodePathname(url.pathname);
}
```

The `new H3Event(...)` call in `H3Core["~request"]` runs **before** the per-request `try/catch`, so the throw can't be caught by `onError`, Nitro, or consumer middleware. With the `srvx` Node adapter, `serve()` invokes the app's `fetch` directly inside the HTTP request listener, and h3's `fetch` throws **synchronously** (the event is constructed outside the `try`) — so it escapes as an **uncaughtException** and crashes the process. Any scanner/bot hitting a junk URL (`GET /%`) takes down a pod.

```js
import { H3 } from "h3";

const app = new H3();
app.get("/**", () => "ok");

// Throws URIError: URI malformed (uncaught in a real server -> process crash)
await app.fetch(new Request("http://localhost/%"));
```

### 🔄 Changes

Guard the decode and fall back to the raw, undecoded pathname when `decodeURI` throws. Routing then simply won't match the un-decodable path (404) instead of crashing. Valid encoded paths still decode normally, and the existing `%25`-preservation behaviour is unchanged.

```ts
  try {
    return decodeURI(
    );
  } catch {
    // Fall back to the raw, undecoded pathname; routing then won't match it.
    return pathname;
  }
}
```

This is the same guard the issue reporter proposed in #1361. It was previously included in #1363, which was closed in favour of landing the path-traversal hardening (`withoutBase` / `resolveDotSegments`) separately — those landed on `main`, but this crash-guard did not. This PR is the minimal, guard-only remainder.

### ✅ Tests

- `test/unit/path.test.ts` — `decodePathname` decodes valid input, preserves `%25` (no double-decode), and returns malformed input unchanged without throwing.
- `test/security.test.ts` — matrix (web + node) regression asserting `/%`, `/%C0%AF`, `/foo%`, `/%E0%A4%A` route normally instead of crashing.

Both fail against the current code (the node-mode cases time out — the crash signature) and pass with the fix. `pnpm lint` is clean.

## Release Notes

* **Bug Fixes**
  * Improved robustness of URL path handling for malformed percent-encoded pathnames. Invalid encoding is now rejected with the correct client error response (HTTP 400), and failures during request initialization no longer bypass normal request processing.

* **Tests**
  * Added security coverage to verify malformed URIs are consistently rejected and valid percent-encoding is accepted.
  * Expanded unit tests for pathname decoding, including proper handling of preserved encoded sequences and expected error behavior.

### DIFF
--- a/src/utils/internal/path.ts
+++ b/src/utils/internal/path.ts
@@ -58,7 +58,16 @@ export function getPathname(path: string = "/"): string {
  * Decode percent-encoded pathname, preserving %25 (literal `%`).
  */
 export function decodePathname(pathname: string): string {
-  return decodeURI(pathname.includes("%25") ? pathname.replace(/%25/g, "%2525") : pathname);
+  try {
+    return decodeURI(pathname.includes("%25") ? pathname.replace(/%25/g, "%2525") : pathname);
+  } catch {
+    // A malformed percent-encoding (a bare "%", or an invalid UTF-8 byte
+    // sequence such as "%C0") makes decodeURI throw a URIError. This runs in
+    // the H3Event constructor, before the per-request try/catch, so an
+    // unguarded throw escapes as an uncaughtException and crashes the process.
+    // Fall back to the raw, undecoded pathname; routing then won't match it.
+    return pathname;
+  }
 }
 
 export function resolveDotSegments(path: string): string {
--- a/src/utils/internal/path.ts
+++ b/src/utils/internal/path.ts
@@ -1,3 +1,5 @@
+import { HTTPError } from "../../error.ts";
+
 export function withLeadingSlash(path: string | undefined): string {
   if (!path || path === "/") {
     return "/";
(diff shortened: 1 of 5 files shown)


davidz627:
hi @pi0! Hope you're doing well! What do you think of this fix? It's something that has trigged in production for my system a few times so I've already monkey-patched but would love to contribute the fix upstream

src/utils/internal/path.ts
```diff
@@ -58,7 +58,16 @@ export function getPathname(path: string = "/"): string {
...
-  return decodeURI(pathname.includes("%25") ? pathname.replace(/%25/g, "%2525") : pathname);
+  try {
+    return decodeURI(pathname.includes("%25") ? pathname.replace(/%25/g, "%2525") : pathname);
+  } catch {
+    // A malformed percent-encoding (a bare "%", or an invalid UTF-8 byte
+    // sequence such as "%C0") makes decodeURI throw a URIError. This runs in
+    // the H3Event constructor, before the per-request try/catch, so an
+    // unguarded throw escapes as an uncaughtException and crashes the process.
+    // Fall back to the raw, undecoded pathname; routing then won't match it.
+    return pathname;
```
pi0: It is unsafe to ignore parse errors. (routing mismatches can cause some middleware like auth to be bypassed in certain conditions for example

pi0x:
> **Disclosure:** This comment and the commit it describes (f437f70) were authored by an AI agent (Claude Code / Claude Opus) acting on behalf of the maintainer, in response to @pi0's review feedback. Please review as you would any change.

Thanks a lot @davidz627 for the crash repro and diagnosis — the root cause analysis (an unguarded `decodeURI` throwing a `URIError` out of the `H3Event` constructor, before the per-request `try/catch`, taking down the process) was spot on. 🙏

Following up on @pi0's review note:

> It is unsafe to ignore parse errors. (routing mismatches can cause some middleware like auth to be bypassed in certain conditions)

I pushed a follow-up commit that keeps your crash fix but replaces the **silent fallback to the raw, undecoded pathname** with an explicit **HTTP 400** rejection, so a malformed URI can never be routed/matched against a different pathname than the one the decode is meant to normalize.

**What changed (f437f70):**

- `decodePathname` now **throws `HTTPError({ status: 400 })`** on malformed percent-encoding instead of returning the raw string.
- The `H3Event` constructor **captures** that error into a `kEventInitError` symbol rather than letting it escape (the constructor still runs before the per-request `try/catch`, so it must not throw — that was the original crash).
- `H3Core["~request"]` **re-throws the captured error at the very start of the request `try/catch`**, before any `onRequest` hook, handler, or route matching. So it flows through the normal `toResponse` error pipeline as a clean 400 and **never reaches routing/middleware** — no bypass risk.

**Tests:**

- `test/unit/path.test.ts` — updated the malformed-input case to assert `decodePathname` throws a 400 `HTTPError` (instead of the previous "returns it unchanged").
- `test/security.test.ts` — updated the malformed-URI matrix cases to assert a **400 response** and that the route handler is **never reached**, plus a new case confirming valid percent-encoded paths still route normally. Runs green in both `web` and `node` modes.

`pnpm vitest run test/security.test.ts test/unit/path.test.ts` and `pnpm lint` both pass (the one remaining lint warning is pre-existing in `test/bench/bench.ts`, unrelated to this change).

pi0:
Thanks for PR. Landing https://github.com/h3js/h3/pull/1424 with a slightly different method

PR #1409, opened 2026-06-05
docs(session): note secure cookie limitation over local HTTP

### LINKED ISSUE #977: Safari by default doesn't persist cookies created by H3 session
### Environment

Latest Safari and MacOS.

This library is being used by TanStack Start and I reported the issue there:

https://github.com/TanStack/router/issues/3492

This affects both the v1 and v2 branches.

### Reproduction

The reproduction is given on TanStack Start example here: https://github.com/TanStack/router/issues/3492

But basically cookies will persist by default in Chrome and Firefox on MacOS but not Safari (and I suspect this also happens on iOS.)

### Describe the bug

The culprit is in H3 on this line: https://github.com/unjs/h3/blob/67e12ae157876c8e5cd8e686ccb6efb1e467861c/src/utils/session.ts#L37

Basically DEFAULT_COOKIE is used as defaults in updateSession: https://github.com/unjs/h3/blob/67e12ae157876c8e5cd8e686ccb6efb1e467861c/src/utils/session.ts#L153 and also in https://github.com/unjs/h3/blob/67e12ae157876c8e5cd8e686ccb6efb1e467861c/src/utils/session.ts#L217

DEFAULT_COOKIE has secure = true.

But if you use http localhost on Safari in dev mode, cookies will not persist by default not, they will fail silently.

So if you do not specify a value for "cookie.secure" it will end up as true and thus Safari development mode will fail to save cookies.

> Note be careful when setting 'secure' to true, as compliant clients will not send the cookie back to the server in the future if the browser does not have an HTTPS connection.

Maybe automatically change cookie secure setting based on ```process.env.NODE_ENV === 'production'``` or something similar?  That is what I am suggesting to TanStack Start.

### Additional context

Again, you can read my original issue over on TanStack Start here describing the issue: https://github.com/TanStack/router/issues/3492

### Logs

```sh

```

Aimee1608:
## Summary

The session example in `docs/4.examples/handle-session.md` configures the cookie with `secure: true`, but the docs don't mention that this silently breaks sessions during local **HTTP** development: compliant browsers (notably Safari/iOS, and Chrome on some local domains) drop `secure` cookies sent over plain HTTP, so the session never persists.

This adds a short note after the **Options** section explaining the behavior and the `cookie: { secure: false }` workaround for local development.

As suggested by @pi0 in #977:

> Overriding is fairly easy with session config `{ cookie: { secure: false } }`, i think we can make the documentation more clear about this.

## Changes

- `docs/4.examples/handle-session.md`: add a `> [!NOTE]` after the Options example documenting the `secure` cookie limitation over local HTTP and the workaround.

Docs-only change, no behavior change. It intentionally does **not** add `NODE_ENV`-based auto-switching of `secure`, since that would be a behavior change that hasn't been agreed on in the issue.

Closes #977

---

> AI-assisted; I reviewed and verified this change before submitting.

## Summary by CodeRabbit

* **Documentation**
  * Updated session handling documentation with enhanced guidance on secure cookie configuration. Clarifies how strict HTTPS-only settings may affect session persistence in local HTTP development and provides practical recommendations for configuring cookies appropriately across different development and production scenarios.


pi0x (approved):

pi0 (approved):

PR #1413, opened 2026-06-12
fix(request): parse first entry of comma-list x-forwarded-proto header

LeSingh1:
Proxies can stack `x-forwarded-proto` as a comma-separated list, for example `"https,http"` where the leftmost entry is the scheme the original client used. `getRequestProtocol` compared the raw header value with strict equality, so any comma-list value fell through and the function returned the scheme from the underlying request URL instead of the client-facing protocol. `toRequest` had the same problem on the same header.

The fix applies the same `split(",")[0].trim()` pattern already used by `getRequestHost` and `getRequestIP` for their respective forwarded headers.

Two failing tests were added before the fix and confirmed broken: one in `test/unit/request.test.ts` that exercises `getRequestProtocol` directly with a fake event, and two in `test/utils.test.ts` under `getRequestURL` that test the comma-list and comma-with-space cases end-to-end through a live app handler. All 48 test files and 1159 tests pass after the fix. Lint and `oxfmt` are clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved protocol detection for proxied requests by correctly parsing `x-forwarded-proto` when it contains a comma-separated list (using the first value with whitespace trimmed), while respecting the option to ignore proxy headers.
* **Tests**
  * Added unit and utility coverage for `x-forwarded-proto` parsing, including comma-separated values, whitespace handling, and the disable-proxy-header option.
* **Documentation**
  * Updated request utility docs to clarify comma-separated `x-forwarded-proto` behavior and the `xForwardedProto: false` option.


pi0x (approved):

pi0 (approved):
Thanks!

PR #1415, opened 2026-06-17
fix: writeEarlyHints resolves immediately instead of hanging (#1383)

### LINKED ISSUE #1383: `writeEarlyHints` never resolves with node runtime

rafaumeu:
Fixes #1383

The callback for Node.js writeEarlyHints is optional and may not
be invoked in all cases. By resolving the promise immediately after
calling writeEarlyHints, we avoid hanging requests while still
sending early hints to the client.

**Changes:**

- Remove unused Promise wrapper that could hang
- Resolve promise immediately after calling writeEarlyHints
- Still sends early hints to the client via native Node.js API

**Testing:**

- Existing writeEarlyHints tests pass (web fallback)
- Manual test with reproduction case works (no hanging)

## Summary by CodeRabbit

* **Tests**
  * Added test script for early hints functionality validation

* **Refactor**
  * Optimized early hints execution for improved efficiency

### DIFF
--- a/src/utils/response.ts
+++ b/src/utils/response.ts
@@ -114,9 +114,11 @@ export function writeEarlyHints(
 ): void | Promise<void> {
   // Use native early hints if available (Node.js)
   if (event.runtime?.node?.res?.writeEarlyHints) {
-    return new Promise((resolve) => {
-      event.runtime?.node?.res?.writeEarlyHints(hints, () => resolve());
-    });
+    // Note: Node.js writeEarlyHints callback is optional and may not be invoked
+    // in all cases. We wrap in a promise but resolve immediately after the call
+    // to avoid hanging when the callback isn't invoked.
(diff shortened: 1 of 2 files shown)


pi0x:

This PR and #1387 both fix the same `writeEarlyHints` hanging bug (#1383) by dropping the await on the native callback entirely. #1385 fixes the same issue at the root cause instead (Node's `writeEarlyHints` doesn't invoke its callback when `link` is missing/empty), while preserving the intended "wait until flushed" behavior for the working case.

Closing in favor of #1385. Thanks for the reproduction and fix attempt — it helped pin down the actual bug.

PR #1417, opened 2026-06-18
feat(proxy): support client aborts

iainsproat:
- client abort signals are propagated upstream
- 502 response on abort remains the default to prevent breaking existing behaviour
- new option `propogateAbortError` to throw client abort error instead of 502 response

## Summary by CodeRabbit

* **New Features**
  * Added a `propagateAbortError` option to control whether client disconnects surface as an `AbortError` or are handled quietly.
  * Proxying now forwards the client abort signal for upstream requests, and combines it with any provided `fetch` signal.
  * Client disconnects now result in HTTP **499** by default; enabling `propagateAbortError` surfaces the `AbortError`. Mid-stream disconnects no longer trigger gateway errors.
* **Tests**
  * Added web-only coverage for default **499** behavior, `propagateAbortError: true`, and mid-stream abort outcomes.

### DIFF
--- a/src/utils/proxy.ts
+++ b/src/utils/proxy.ts
@@ -179,54 +179,16 @@ export async function proxy(
     await opts.onResponse(event, response);
   }
 
-  // A client disconnect during response streaming aborts the forwarded
-  // `event.req.signal` and errors the upstream body stream *after* the
-  // try/catch above has returned. By default, swallow that abort quietly (the
-  // bytes are never delivered — the client is already gone) instead of letting
-  // it surface as an unhandled streaming error. When opted in via
-  // `propagateAbortError`, leave the stream untouched so the error propagates.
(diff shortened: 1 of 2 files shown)


src/utils/proxy.ts
```diff
@@ -97,6 +106,7 @@ export async function proxy(
   opts: ProxyOptions = {},
 ): Promise<HTTPResponse> {
   const fetchOptions: RequestInit = {
+    signal: event.req.signal,
```
iainsproat: should this also be gated behind an option? e.g. `signal: opts.propagateRequestSignal ? event.req.signal : undefined`

pi0 (approved):

pi0 (approved):

pi0:
Thanks!

PR #1422, opened 2026-06-30
fix(sanitizeStatusCode): return default for non-numeric input instead of NaN

mahirhir:
`sanitizeStatusCode` accepts `string | number` and parses strings with `Number.parseInt`. For a non-numeric string `Number.parseInt` returns `NaN`, and `NaN` slips through the range check because every comparison with `NaN` is `false`:

```ts
sanitizeStatusCode("abc"); // -> NaN  (expected 200)
```

So an invalid code can reach `createError` (`err.statusCode = sanitizeStatusCode(input.statusCode, ...)`) and the response helpers (`event.node.res.statusCode = sanitizeStatusCode(...)`), where a `NaN` status is no longer a valid HTTP status.

This is the same issue that was fixed for v2 on `main` in #1420; this backports the guard to the v1 line. The `Number.parseInt` behaviour and the `100..999` range are unchanged, I only added the `Number.isNaN` check:

```ts
if (Number.isNaN(statusCode) || statusCode < 100 || statusCode > 999) {
  return defaultStatusCode;
}
```

Added a unit test covering valid codes, out-of-range codes and non-numeric strings. It fails before the change (returns `NaN`) and passes after.

### DIFF
--- a/src/utils/sanitize.ts
+++ b/src/utils/sanitize.ts
@@ -24,7 +24,10 @@ export function sanitizeStatusCode(
   if (typeof statusCode === "string") {
     statusCode = Number.parseInt(statusCode, 10);
   }
-  if (statusCode < 100 || statusCode > 999) {
+  // `Number.parseInt` returns `NaN` for non-numeric strings, and `NaN` passes
+  // the range check below (every comparison with `NaN` is `false`), so an
+  // invalid status code would leak through unchanged. Guard against it.
+  if (Number.isNaN(statusCode) || statusCode < 100 || statusCode > 999) {
     return defaultStatusCode;
   }
   return statusCode;
--- a/test/sanitize.test.ts
+++ b/test/sanitize.test.ts
@@ -1,0 +1,26 @@
+import { describe, it, expect } from "vitest";
+import { sanitizeStatusCode } from "../src/utils/sanitize";
+
+describe("sanitizeStatusCode", () => {
+  it("returns valid status codes unchanged", () => {
+    expect(sanitizeStatusCode(200)).toBe(200);
+    expect(sanitizeStatusCode(404)).toBe(404);
+    expect(sanitizeStatusCode("301")).toBe(301);
+  });
+
+  it("returns the default for missing or out-of-range input", () => {
+    expect(sanitizeStatusCode(undefined)).toBe(200);
+    expect(sanitizeStatusCode(0)).toBe(200);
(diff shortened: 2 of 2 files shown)


mahirhir:
This targets `v1`, which last moved on 2026-05-19, so I checked whether it is still worth having before asking anyone to look. I think it is, and `main` is the reason.

`main` already carries this guard, with a test file beside it:

```ts
// main, src/utils/sanitize.ts:28
if (!Number.isInteger(statusCode) || statusCode < 100 || statusCode > 599) {
  return defaultStatusCode;
}
```

`v1` does not:

```ts
// v1, src/utils/sanitize.ts:27
if (statusCode < 100 || statusCode > 999) {
  return defaultStatusCode;
}
```

```console
$ curl -o /dev/null -w '%{http_code}' .../h3/main/test/unit/sanitize.test.ts
200
$ curl -o /dev/null -w '%{http_code}' .../h3/v1/test/sanitize.test.ts
404
```

So the decision has already been made on `main`. This PR is the same decision applied to the branch that did not get it, and it adds the test file `v1` has never had.

Why a bare range check does not catch it, measured rather than argued:

```
input     parseInt   < 100    > 999    passes the range check
"abc"          NaN   false   false   true
"ok"           NaN   false   false   true
"200px"        200   false   false   true
""             NaN   false   false   true
```

Every comparison with `NaN` is `false`, so `NaN` satisfies neither bound and is returned as the status code. It then reaches whatever constructs the response, which is not where you want to discover it. (That table is Node measuring the premise; it is not a run of h3.)

On whether `v1` is worth patching at all, since a five-month-old branch usually means no:

```
npm dist-tags:  latest = 2.0.1-rc.31    1x = 1.15.11
1.x newest      1.15.11   2026-04-01
2.x newest      2.0.1-rc.31   2026-09-03
```

`latest` is still a release candidate, and there is a maintained `1x` tag. So the 1.x line is what most people are actually running, which is the argument for the backport rather than against it. If you see it differently and `v1` is closed to fixes, say so and I will close this rather than leave it open.

State as your own CI reports it rather than as I claim it:

```
tests                             = pass, 41s
autofix                           = pass
Socket Security (both checks)     = pass
mergeable                         = MERGEABLE
```

So the added test file runs and passes there, which is the one thing in this PR I would otherwise be asking you to take on trust. Happy to rebase onto the current `v1` head if you would rather have a fresh run, or to take the two-line change and close this. No attribution needed.

PR #1435, opened 2026-07-04
types: expose `.crossws` on `defineWebSocketHandler` return type

### LINKED ISSUE #1258: Type mismatch occurred when using `crossws`

momomuchu:
`defineWebSocketHandler()` attaches `crossws` hooks to the returned Response at runtime, but the declared return type resolved to `unknown`, so reading `res.crossws` needed a cast (#1258).

This adds a scoped `WebSocketResponse` type and typed overloads (hooks-only and http-fallback) so `crossws` is visible. Type-only change, runtime behaviour is unchanged. Kept scoped rather than introducing a global response type, per the maintainer note on #1306.

Tests: a runtime check that `res.crossws` equals the hooks without a cast, plus `expectTypeOf` assertions. Full suite passes, tsc and lint clean.

Closes #1258

## Summary by CodeRabbit

* **New Features**
  * Exposed a typed `WebSocketResponse` from the package entrypoint.
  * Enhanced `defineWebSocketHandler` typing so WebSocket responses include `crossws` in a type-safe way.
* **Bug Fixes**
  * When using async hook factories, `crossws` is now attached only after hooks resolve, ensuring correct values on `426 Upgrade Required`.
  * Kept accurate return-type inference when combining WebSocket handling with an HTTP fallback.
* **Tests**
  * Added type-level and runtime coverage for `crossws` presence and async hook behavior.
* **Documentation**
  * Removed outdated `defineWebSocketHandler` behavioral details from the docs.

### DIFF
--- a/docs/2.utils/9.more.md
+++ b/docs/2.utils/9.more.md
@@ -93,6 +93,37 @@ const hooks = defineWebSocket({
 
 ### `defineWebSocketHandler(http?)`
 
+Define WebSocket event handler.
+
+By default, non-upgrade (plain HTTP) requests receive a `426 Upgrade Required` response. Pass an `http` handler to serve those requests instead, allowing the same route to handle both WebSocket upgrades and regular HTTP requests. WebSocket upgrade requests always go to `hooks`.
+
+Note: the `http` handler only handles non-upgrade requests. To reject or customize the upgrade handshake itself, use the crossws `upgrade` hook instead.
+
+**Example:**
+
+```ts
+// WebSocket-only route (non-upgrade requests get `426 Upgrade Required`)
+app.get(
+  "/_ws",
+  defineWebSocketHandler({
+    message: (peer, message) => peer.send(message.text()),
+  }),
+);
+```
+
+**Example:**
+
+```ts
+// Handle both WebSocket upgrades and plain HTTP on the same route
+app.get(
+  "/_ws",
+  defineWebSocketHandler(
+    { message: (peer, message) => peer.send(message.text()) },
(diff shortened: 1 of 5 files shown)


pi0 (approved):

momomuchu:
The async-factory await concern was fixed in 892bfe0 (the factory result is awaited before crossws is attached, covered by the awaits-an-async-hooks-factory test). The follow-up review's return-type point is also addressed now: defineWebSocketHandler's factory overload returns WebSocketResponse | Promise<WebSocketResponse>, with a type test that fails on the sync-only type. Resolving as addressed.

PR #1439, opened 2026-07-06
fix: support plain object request headers

onmax:
Keeps deprecated `getRequestHeaders()` tolerant of plain object header bags while preserving the normal `Headers.entries()` path. This does not change the primary h3 request-header contract.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved request header handling so headers are read more reliably across different request shapes.
  * Added a safer fallback when header entries aren’t available, reducing errors in edge cases.

* **Tests**
  * Added coverage for header extraction from plain request objects to verify consistent behavior.

### DIFF
--- a/test/unit/request.test.ts
+++ b/test/unit/request.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it } from "vitest";
 import { requestWithURL, requestWithBaseURL } from "../../src/utils/request.ts";
-import { getRequestProtocol } from "../../src/index.ts";
+import { getRequestHeaders, getRequestProtocol } from "../../src/index.ts";
 
 // Minimal fake HTTPEvent for unit-testing getRequestProtocol without a live server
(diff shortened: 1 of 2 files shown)


test/unit/request.test.ts
```diff
@@ -90,6 +90,13 @@ describe("requestWithBaseURL", () => {
   });
 });
 
+describe("getRequestHeaders", () => {
+  it("supports plain object request headers", () => {
+    const event = { req: { headers: { "x-test": "works" } } } as any;
```
pi0, line 95: This is not a valid H3 event in v2. you should _never_ pass h3 v1 event.

pi0:
Appreciate the pr. while it is only applied to deprecated utils, combining h3 v1 event with v2 utils can be inheringly dangrous and small patches like this to workaround makes it more normal while it shouldn't be.

PR #1462, opened 2026-07-14
fix(cookie): dedup cookies with leading-dot / mixed-case domains

pi0x:
Cookie dedup in `setCookie` keyed new cookies on the raw `options.domain`, but keyed existing cookies on the normalized domain from `parseSetCookie` (leading `.` stripped, lowercased). For domains like `.Example.com` the keys never matched, so two `Set-Cookie` headers were emitted instead of a replacement.

This normalizes the domain (strip leading `.`, lowercase) on both sides in `_getDistinctCookieKey`, per RFC 6265 where Domain is case-insensitive and a leading `.` is ignored. Adds a regression test covering setting the same cookie twice with `domain: ".Example.com"`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cookie deduplication by treating domain names consistently, regardless of leading dots or letter casing.
  * Ensured the latest value is retained when equivalent cookies are set multiple times.

### DIFF
--- a/src/utils/cookie.ts
+++ b/src/utils/cookie.ts
@@ -243,7 +243,9 @@ export function deleteChunkedCookie(
  * @see https://httpwg.org/specs/rfc6265.html#rfc.section.4.1.2
  */
 function _getDistinctCookieKey(name: string, options: { domain?: string; path?: string }) {
-  return [name, options.domain || "", options.path || "/"].join(";");
+  // Domain is case-insensitive and a leading "." is ignored (RFC 6265).
+  const domain = (options.domain || "").replace(/^\./, "").toLowerCase();
+  return [name, domain, options.path || "/"].join(";");
 }
 
 // Maximum number of chunks allowed for chunked cookies.
--- a/test/cookies.test.ts
+++ b/test/cookies.test.ts
@@ -130,6 +130,17 @@ describeMatrix("cookies", (t, { it, expect, describe }) => {
       expect(await result.text()).toBe("200");
     });
 
+    it("deduplicates cookies with leading-dot / mixed-case domains", async () => {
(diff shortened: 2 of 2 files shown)


pi0 (approved):

pi0x (commented):
**Review verdict: LGTM (approve)** — posting as a comment since GitHub blocks self-approval on this account.

Clean, correct, minimal fix.

**Correctness**
- Normalizing inside `_getDistinctCookieKey` fixes both sides at once: the new-cookie key (`options.domain`) and the existing-cookie key (from `parseSetCookie`) now go through identical normalization, so they match regardless of leading dot or casing. It's the single shared chokepoint, so no missed call sites — `setCookie` is the only consumer, and `deleteCookie`/chunked helpers route through it.
- `(options.domain || "").replace(/^\./, "").toLowerCase()` handles empty/undefined domain (→ `""`) and idempotently re-normalizes the already-normalized `parseSetCookie` output. Path and cookie name are left untouched, which is correct: RFC 6265 treats path and name as case-sensitive, only Domain as case-insensitive with a single leading dot ignored.
- The actual emitted `Set-Cookie` header still carries the user's raw `Domain=.Example.com` — only the dedup key is normalized. Correct: we dedup without mutating the requested cookie.

**Verification**
- Reverted the source change and confirmed the new test fails as expected (two `Set-Cookie` headers emitted); restored, and all 98 cookie tests pass in both web and node matrix modes. `pnpm lint` (oxlint + oxfmt) clean.
- Test quality is good: covers both leading-dot and mixed-case in one case, asserts exact single-header output.

Matches AGENTS.md conventions and RFC 6265 §4.1.2/§5.1.3.

PR #1463, opened 2026-07-14
fix(auth): decode Basic-auth credentials as UTF-8

pi0x:
### Bug

`requireBasicAuth` decoded the base64 `Authorization` header with `atob()`, which returns a Latin-1 string. UTF-8 credential bytes were therefore never decoded, so a user with a non-ASCII password (RFC 7617 `charset="UTF-8"`) could never authenticate — the server compares against `TextEncoder`/UTF-8 values, guaranteeing a mismatch.

### Fix

Decode the base64 payload through `TextDecoder("utf-8")` so multi-byte credentials round-trip correctly, and advertise `charset="UTF-8"` in the `WWW-Authenticate` challenge per RFC 7617. Added regression tests covering a non-ASCII password and the challenge header.

### DIFF
--- a/src/utils/auth.ts
+++ b/src/utils/auth.ts
@@ -58,7 +58,10 @@ export async function requireBasicAuth(event: HTTPEvent, opts: BasicAuthOptions)
   }
   let authDecoded: string;
   try {
-    authDecoded = atob(b64auth);
+    // RFC 7617: credentials are base64-encoded and may use UTF-8 (charset="UTF-8").
+    authDecoded = new TextDecoder("utf-8", { fatal: false }).decode(
+      Uint8Array.from(atob(b64auth), (c) => c.charCodeAt(0)),
+    );
   } catch {
     throw authFailed(event, opts?.realm);
   }
@@ -109,7 +112,7 @@ function authFailed(event: HTTPEvent, realm: string = "") {
     status: 401,
     statusText: "Authentication required",
     headers: {
-      "www-authenticate": `Basic realm=${JSON.stringify(realm)}`,
+      "www-authenticate": `Basic realm=${JSON.stringify(realm)}, charset="UTF-8"`,
     },
   });
 }
--- a/src/utils/auth.ts
+++ b/src/utils/auth.ts
@@ -58,12 +58,17 @@ export async function requireBasicAuth(event: HTTPEvent, opts: BasicAuthOptions)
   }
   let authDecoded: string;
   try {
-    // RFC 7617: credentials are base64-encoded and may use UTF-8 (charset="UTF-8").
-    authDecoded = new TextDecoder("utf-8", { fatal: false }).decode(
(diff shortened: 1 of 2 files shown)


pi0x (commented):
## Review: Approve (posted as comment — GitHub blocks self-approval on this account)

Clean, correctly-scoped fix. Verified locally on the branch: `pnpm vitest run test/auth.test.ts` → 36/36 pass, `pnpm lint` (oxlint + oxfmt) clean.

**Correctness**
- The decode idiom `new TextDecoder("utf-8", { fatal: false }).decode(Uint8Array.from(atob(b64auth), c => c.charCodeAt(0)))` is the standard, correct way to round-trip UTF-8 credential bytes that `atob()` would otherwise leave as Latin-1.
- `fatal: false` is the right call: invalid UTF-8 produces U+FFFD replacements → auth mismatch → 401, never an unhandled 500. Invalid base64 still throws from `atob()` and is caught → 401 (covered by the existing "responds 401 for invalid base64" test).
- Colon-split via `indexOf(":")`/`slice` still keeps `:` characters in passwords (existing multi-colon tests pass).
- Challenge `Basic realm="...", charset="UTF-8"` matches the RFC 7617 §2.1 example verbatim and is valid per RFC 7235 auth-param grammar.

**Tests**
- The UTF-8 regression test exercises a genuine multi-byte password (`pä$$wörd🔒`, including a 4-byte emoji) end-to-end and asserts a 200. The challenge test asserts `charset="UTF-8"` on the 401. `Buffer` usage is consistent with the rest of the file.

**Scope / style**
- Change is limited to the auth decode, the challenge header, and tests — no scope creep. Style, formatting, and comments follow AGENTS.md conventions.

pi0x:
Good idea — implemented in 8face06. The decode is now: strict UTF-8 first (`TextDecoder("utf-8", { fatal: true })`), and on failure keep the raw `atob()` Latin-1 string, so legacy (pre-RFC 7617) clients sending Latin-1 credentials still authenticate.

Notes from evaluating it:

- **Correctness:** ASCII is identical in both encodings, so ambiguity only exists when a Latin-1 client sends bytes that happen to form valid UTF-8 (e.g. Latin-1 `"Ã©"` = `C3 A9` = UTF-8 `"é"`). That's the inherent cost of any charset sniff, is vanishingly rare for real passwords, and fails closed (401), never open.
- **Security:** No credential-equivalence risk — strict UTF-8 decoding and Latin-1 are each injective, so the only cross-path overlap is that the UTF-8 and Latin-1 encodings of the *same* password both authenticate, which is the point. It's actually safer than the previous `fatal: false` decode, where distinct invalid byte sequences collapsed into U+FFFD replacement characters.
- The challenge still advertises `charset="UTF-8"` (RFC 7617); the fallback is receive-side leniency only.

Added a regression test for a genuinely Latin-1-encoded credential (`0xE4`, invalid as UTF-8) and confirmed it failed before the change.

pi0 (approved):

PR #1464, opened 2026-07-14
fix(cors): warn on credentials with `null` origin

pi0x:
`origin: "null"` combined with `credentials: true` is dangerous: sandboxed iframes, `data:`/`file:` documents, and other opaque origins all send `Origin: null`, so credentials would be shared across untrusted contexts. This adds a warn-once message in `resolveCorsOptions`, mirroring the existing wildcard-origin and wildcard-`exposeHeaders` warnings. The hazard is also documented on the `origin` option JSDoc, and a unit test asserts the warning fires.

## Summary by CodeRabbit

* **Documentation**
  * Clarified the security risks of combining credentialed CORS requests with `origin: "null"`.

* **Bug Fixes**
  * Added warnings when credentialed CORS configuration allows opaque origins, including direct or list-based `"null"` origins.
  * Warnings are deduplicated to avoid repeated notifications.

* **Tests**
  * Added coverage verifying warnings for both supported `"null"` origin configurations.

### DIFF
--- a/src/utils/cors.ts
+++ b/src/utils/cors.ts
@@ -20,6 +20,10 @@ export interface CorsOptions {
    * If an array of strings or regular expressions, it can be used with origin matching.
    * If a custom function, it's used to validate the origin. It takes the origin as an argument and returns `true` if allowed.
    *
+   * Avoid `"null"` together with `credentials: true`. Sandboxed iframes, `data:`/`file:` documents,
+   * and other opaque origins all send `Origin: null`, so allowing it with credentials would share
+   * them across untrusted contexts.
+   *
    * @see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin
    * @default "*"
    */
--- a/src/utils/internal/cors.ts
+++ b/src/utils/internal/cors.ts
@@ -44,6 +44,12 @@ export function resolveCorsOptions(options: CorsOptions = {}): ResolvedCorsOptio
     );
   }
 
+  if (resolved.credentials && resolved.origin === "null") {
+    warnOnce(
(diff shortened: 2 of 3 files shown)


pi0x (commented):
Independent review (did not author this change). Tests pass (`test/unit/cors.test.ts`, 44 passed) and `pnpm lint` is clean.

**LGTM** — well-scoped DX warning with no behavior change to actual CORS resolution:

- Trigger `resolved.credentials && resolved.origin === "null"` is correct and properly gated.
- Reuses the existing `warnOnce` helper, so dedup semantics match the wildcard warnings; the distinct message string dedupes independently.
- Test is well-isolated: `exposeHeaders: ["X-Custom"]` + non-wildcard origin ensures exactly one warning fires, so `toHaveBeenCalledOnce()` is meaningful.
- JSDoc is accurate — sandboxed iframes and `data:`/`file:`/opaque origins do serialize `Origin: null`.

**Non-blocking suggestion:** the warning only fires for the top-level string `origin: "null"`, not for an array containing it (e.g. `origin: ["null"]`). Unlike the wildcard case — where `["*"]` is inert because no real request sends `Origin: *` — an array containing `"null"` is a *live* hazard: `isCorsOriginAllowed` does `origin === _origin` over the array, so `Origin: null` matches and gets reflected with credentials. Consider also covering `Array.isArray(resolved.origin) && resolved.origin.includes("null")`. Fine to defer since it matches the current wildcard-warning pattern and is warn-only.

pi0x:
Addressed the review gap in e66906d: the warning now also fires when the `origin` array contains `"null"` (the array path does an exact string match, so `origin: ["null"]` is a live hazard with credentials). Added a unit test for the array case; tests and lint pass.

pi0 (approved):

PR #1465, opened 2026-07-14
fix(validate): convert malformed JSON to 400 in validated-handler path

pi0x:
A rejected `req.json()` in the validated-request proxy (`_validatedJson`) was not caught, so a malformed JSON body sent to a `defineValidatedHandler` surfaced as an unhandled **500** — inconsistent with `readBody`, which returns a **400** "Invalid JSON body".

This wraps the `.json()` call so a parse failure becomes a `400` `HTTPError` with message "Invalid JSON body". Adds a regression test covering malformed JSON to a validated handler.

### DIFF
--- a/src/utils/internal/validate.ts
+++ b/src/utils/internal/validate.ts
@@ -113,6 +113,13 @@ export function validatedRequest<
         return function _validatedJson() {
           return req
             .json()
+            .catch(() => {
+              throw new HTTPError({
+                status: 400,
+                statusText: "Bad Request",
+                message: "Invalid JSON body",
+              });
+            })
             .then((data) => validate.body!["~standard"].validate(data))
             .then((result) => {
               if (result.issues) {
--- a/test/handler.test.ts
+++ b/test/handler.test.ts
@@ -214,6 +214,21 @@ describe("handler.ts", () => {
       expect(res.status).toBe(400);
     });
 
+    it("malformed JSON body", async () => {
+      const res = await handler.fetch(
+        toRequest("/?id=123", {
+          method: "POST",
+          headers: { "x-token": "abc" },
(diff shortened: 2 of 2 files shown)


pi0x (commented):
Independent review — LGTM (posting as comment since GitHub blocks self-approval on this account).

Clean, minimal fix that closes a real gap: a malformed JSON body to a `defineValidatedHandler` previously surfaced as an unhandled 500, now a 400.

Verified locally on the branch:
- Error shape is byte-for-byte consistent with `readBody` (`src/utils/body.ts`): `status: 400`, `statusText: "Bad Request"`, `message: "Invalid JSON body"`.
- The `.catch()` sits only on `req.json()`, before the validation `.then()` chain, so schema/validation errors are untouched and keep their existing `createValidationError` shape/status.
- The regression test genuinely fails without the fix (asserted 500 before, 400 after) and passes with it. It correctly satisfies header/query validation first so the JSON-parse path is actually exercised.
- `pnpm vitest run test/handler.test.ts` and `pnpm lint` both green.

Two minor, non-blocking notes (both edge cases and strict improvements over the prior 500):
1. `req.json()` bundles body-reading + parsing, so a genuine body/stream read error would also be reported as 400 "Invalid JSON body". `readBody` wraps only `JSON.parse` (not `req.text()`), so it would surface such a read error as 500. Rare, but a slight semantic divergence if you want the two paths to treat internal read failures identically.
2. Empty body now returns 400 "Invalid JSON body" here, whereas `readBody` treats an empty body as `undefined` and hands it to the schema. Not wrong, just worth being aware the two paths diverge for empty bodies.

Neither blocks merge.

pi0 (approved):

PR #1466, opened 2026-07-14
fix(cors): set single-valued CORS headers instead of appending

pi0x:
Applying CORS headers more than once (e.g. middleware plus handler) duplicated single-valued response headers, producing invalid values like `*, *` for `access-control-allow-origin`, `access-control-allow-methods`, `access-control-max-age`, and `access-control-allow-credentials`.

This switches `appendCorsHeaders`/`appendCorsPreflightHeaders` to use `.set` for CORS response headers, keeping `.append` only for the legitimately multi-valued `vary` header. Added regression tests that apply CORS twice and assert single values.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed CORS responses to prevent duplicate single-value headers when headers are applied more than once.
  * Preserved multiple `Vary` header values correctly for preflight and standard requests.

* **Tests**
  * Added coverage verifying consistent CORS headers across repeated applications.

### DIFF
--- a/src/utils/cors.ts
+++ b/src/utils/cors.ts
@@ -115,10 +115,7 @@ export function appendCorsPreflightHeaders(event: H3Event, options: CorsOptions)
   if (varyValues.length > 0) {
     headers.vary = varyValues.join(", ");
   }
-  for (const [key, value] of Object.entries(headers)) {
-    event.res.headers.append(key, value);
-    event.res.errHeaders.append(key, value);
-  }
+  setCorsHeaders(event, headers);
 }
 
 /**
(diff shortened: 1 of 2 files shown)


pi0x (commented):
**Review: LGTM** (posting as comment — GitHub blocks self-approval on own PR)

The change is correct and well-targeted.

**Correctness**
- All single-valued CORS response headers now use `.set`: `access-control-allow-origin`, `access-control-allow-methods`, `access-control-allow-credentials`, `access-control-allow-headers`, `access-control-expose-headers`, `access-control-max-age`. Each is produced by the `create*` helpers as a single fully-computed value (comma-joined lists built in one shot), so `.set` is the right semantics — including for `allow-headers`/`expose-headers`, where a repeated application should replace rather than concatenate.
- `vary` correctly stays on `.append` (legitimately multi-valued; distinct code paths contribute different vary tokens).
- Both `res.headers` and `res.errHeaders` are handled identically, matching prior behavior.
- Single-application output (headers emitted and their values) is unchanged — verified by the existing unit tests passing.

**API/types**: `setCorsHeaders` is a non-exported internal helper; no public API or type surface change.

**Tests**: The two regression tests apply CORS twice and assert single values, directly covering the `*, *` bug.

**Verification**: `pnpm vitest run test/unit/cors.test.ts` → 46 passed; `pnpm lint` clean.

**Minor, non-blocking**: Applying preflight twice still duplicates `vary` tokens (e.g. `origin, origin`). Harmless/valid per HTTP and out of scope for this fix; noting in case dedup is later desired.

pi0 (approved):

PR #1467, opened 2026-07-14
fix(proxy): forwardHeaders must not override framing headers

pi0x:
`forwardHeaders` is documented as an escape hatch for the "soft" denylist drops (`host`, `accept-encoding`), but it previously won over the *entire* hop-by-hop denylist and `Connection` nominations. That let a caller force-forward true framing headers (`connection`, `keep-alive`, `transfer-encoding`, `te`, `trailer`, `upgrade`, `proxy-authorization`, `proxy-connection`) upstream, which can desync request framing or leak the inbound proxy's credentials.

This partitions the denylist into a dedicated `framingHeaders` set: `forwardHeaders` may now only override the soft drops, never the framing set or `Connection`-nominated tokens. The `forwardHeaders` JSDoc is updated to document the constraint, and regression tests cover both that framing headers stay dropped and that soft drops (e.g. `accept-encoding`) still forward.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved proxy header forwarding to prevent unsafe hop-by-hop framing headers from being forwarded.
  * Preserved the ability to explicitly forward permitted headers, such as `Accept-Encoding`.
  * Added safeguards for `Transfer-Encoding`, `Connection`, and `Keep-Alive` while continuing to forward valid headers.

### DIFF
--- a/src/utils/internal/proxy.ts
+++ b/src/utils/internal/proxy.ts
@@ -22,6 +22,24 @@ export const ignoredHeaders: Set<string> = /* @__PURE__ */ new Set([
 ]);
 
 /**
+ * The subset of `ignoredHeaders` carrying true hop-by-hop framing/connection
+ * semantics (RFC 9110 §7.6.1). Unlike the "soft" drops (`host`,
+ * `accept-encoding`), forwarding these upstream can desync request framing or
+ * leak the inbound proxy's credentials — so `forwardHeaders` must never
+ * override them.
+ */
+export const framingHeaders: Set<string> = /* @__PURE__ */ new Set([
+  "connection",
+  "keep-alive",
+  "transfer-encoding",
+  "te",
+  "trailer",
+  "upgrade",
+  "proxy-authorization",
+  "proxy-connection",
+]);
+
+/**
  * Parse a `Connection` header value into a lowercased set of the field names it
  * nominates as hop-by-hop (RFC 9110 §7.6.1). Those fields must not be relayed
  * across the proxy in either direction.
--- a/src/utils/internal/proxy.ts
+++ b/src/utils/internal/proxy.ts
@@ -24,7 +24,7 @@ export const ignoredHeaders: Set<string> = /* @__PURE__ */ new Set([
 /**
  * The subset of `ignoredHeaders` carrying true hop-by-hop framing/connection
  * semantics (RFC 9110 §7.6.1). Unlike the "soft" drops (`host`,
- * `accept-encoding`), forwarding these upstream can desync request framing or
+ * `accept-encoding`, `expect`), forwarding these upstream can desync request framing or
  * leak the inbound proxy's credentials — so `forwardHeaders` must never
  * override them.
  */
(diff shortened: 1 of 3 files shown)


pi0x (commented):
**Verdict: APPROVE** (one minor, non-blocking doc nit)

Reviewed independently. Fetched the branch, `pnpm install`, `pnpm vitest run test/proxy.test.ts` (104 passed / 24 skipped), and `pnpm lint` all green.

**Correctness — looks right:**
- The framing guard is effective: `framingHeaders` is a strict subset of `ignoredHeaders`, so a blocked framing header still falls through to the `ignoredHeaders` drop rather than leaking. (Worth keeping in mind: if a future entry is added to `framingHeaders` but *not* `ignoredHeaders`, the `!framingHeaders.has(name)` block would be a no-op since the header would forward via the `!ignoredHeaders.has(name)` branch. A brief invariant comment could help, but current state is fine.)
- Case-insensitivity holds end to end: `forwardHeaders` is lowercased up front, `framingHeaders` entries are lowercase, `Headers.entries()` yields lowercase names, and `connectionTokens()` lowercases. All comparisons align.
- `Connection`-nominated tokens are now correctly excluded from the escape hatch (`!connectionNominated.has(name)`) and still dropped below.
- No behavior change when `forwardHeaders` is unset: `forwardHeaders?.includes(name)` short-circuits to falsy, so the added `&&` clauses never run.

**Tests — solid:**
- Confirmed the framing regression test genuinely fails without the fix (`expected 'chunked' to be undefined` → received `'chunked'`), so it's a real guard, not a tautology.
- The web-only skip on the framing test is justified: on the node target `t.fetch` uses real `fetch`, which strips forbidden request headers (`connection`/`transfer-encoding`/`keep-alive`) before they ever reach the proxy, so the assertion can't be exercised there. The soft-drop (`accept-encoding`) test correctly runs on both targets.

**Minor nit (non-blocking):** the partition leaves three soft drops overridable by `forwardHeaders` — `host`, `accept-encoding`, and also `expect` (it's in `ignoredHeaders` but not `framingHeaders`). The `ProxyOptions` JSDoc says *"Only the 'soft' drops (`host`, `accept-encoding`) can be overridden this way"*, which reads as exhaustive and omits `expect`. Leaving `expect` soft is defensible (it isn't in the classic RFC 9110 §7.6.1 hop-by-hop set), but consider either mentioning `expect` in the doc or moving it into `framingHeaders` for consistency. Behavior for `expect` is unchanged by this PR either way.

Style matches AGENTS.md conventions (`/* @__PURE__ */ new Set`, `Set<string>` for isolatedDeclarations, RFC-referenced JSDoc).

pi0x:
Addressed the reviewer's doc nit in ebe4157: the `forwardHeaders` JSDoc (and the matching `framingHeaders` comment) now lists `expect` in the soft-drop set alongside `host` and `accept-encoding`. `expect` stays soft — it was not moved into `framingHeaders`.

pi0 (approved):

PR #1468, opened 2026-07-14
fix(auth): harden basic-auth realm handling and credential timing

pi0x:
Hardens `requireBasicAuth`/`basicAuth` for realm consistency and timing defense-in-depth. The first challenge (missing `Authorization` header) now carries the configured realm, and the documented `"auth"` default is applied consistently to both the challenge and `event.context.basicAuth.realm`. Username and password comparisons are now evaluated unconditionally so failure timing no longer distinguishes an unknown user from a wrong password, and empty user-id/password are permitted when a custom `validate` function is supplied (RFC 7617).

Parsing nits: accept `1*SP` between the `Basic` scheme and token (RFC 9110), and sanitize the realm to the RFC 9110 §5.6.4 quoted-string charset instead of `JSON.stringify`, which previously threw on non-Latin-1 realms. Added regression tests for all three observable behaviors; full suite and lint pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Basic Authentication challenges by always using the configured realm, safely escaping it for the `WWW-Authenticate` header.
  * Refined custom credential validation so empty usernames/passwords are handled correctly when a validator is provided.
  * Enhanced authentication failure handling and robustness, including support for non-Latin realm values.
  * Accepted `Authorization` headers that include extra spaces between the scheme and credentials.
* **Tests**
  * Added authentication coverage for realm in challenges, spacing variations, and non-Latin realm values.

### DIFF
--- a/src/utils/auth.ts
+++ b/src/utils/auth.ts
@@ -16,6 +16,11 @@ type _BasicAuthOptions = {
 
   /**
    * Custom validation function for basic auth.
+   *
+   * When provided, the built-in non-empty check is skipped and this function
+   * receives the decoded `username`/`password` as-is, including empty strings
+   * (RFC 7617 permits an empty user-id and/or password). It must return `false`
+   * to reject empty or otherwise invalid credentials.
    */
   validate: (username: string, password: string) => boolean | Promise<boolean>;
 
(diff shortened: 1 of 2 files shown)


pi0x (commented):
Verdict: APPROVE (independent review — I did not author this change).

Reviewed the diff, checked out the branch, and ran the suite: `pnpm vitest run test/auth.test.ts` → 42 passed; `pnpm lint` (oxlint + oxfmt) clean. I also reverted only the source (keeping the new tests) and confirmed all three regression tests fail against `main`, so they are genuine regressions, not tautologies.

What I verified:

1. Constant-shape timing. `usernameOk`/`passwordOk`/`validateOk` are each computed on their own line before the `if`, so both `timingSafeEqual` calls always execute — the only short-circuit (`!opts.username || …`) depends on static config, not request data, so the timing shape is invariant across requests for a given config. The final `if (!usernameOk || !passwordOk || !validateOk)` short-circuits only over already-computed booleans. This correctly removes the old `||`-chain side-channel (unknown-user skipping the password compare). `randomJitter()` still fires on failure. Correct.

2. Realm sanitizer. `quoteRealm` matches RFC 9110 §5.6.4: strips everything outside `HTAB / SP / %x21-7E / obs-text(%x80-FF)` (so control chars, DEL 0x7F, and >0xFF code points go), then escapes DQUOTE and backslash as quoted-pairs. This is the correct qdtext + quoted-pair charset and avoids the `JSON.stringify` non-Latin-1 header throw. Correct.

3. Empty-credential change does not weaken the default path. `if (!opts.validate && (!username || !password))` still rejects empty user-id/password whenever no custom `validate` is supplied (including the common `basicAuth({ password })` case). Empty creds only reach a user-provided `validate`, per RFC 7617. Correct.

4. Default realm `"auth"` is now consistent across the missing-header challenge, all other challenges, `context.basicAuth.realm`, and the JSDoc (`Defaults to "auth"`). No existing test regressed.

Minor (non-blocking) notes:

- The scheme regex `/^basic\s+(.+)$/i` is commented as "one or more spaces" but `\s` also matches TAB/CR/LF/FF/VT — broader than RFC 9110 `1*SP`. Harmless in practice (the header is already line-parsed by the HTTP layer and token68 is base64, so atob rejects anything odd), but the comment slightly overstates strictness. Consider `[ ]+` or `[\t ]+` if you want to match the comment, or reword the comment.

- Expected semantic overlap with #1463 (UTF-8 decode): both edit the `authFailed` `www-authenticate` line and the decode block. A merge conflict is expected there. Whichever lands second should reconcile so the final code keeps all three: the `quoteRealm` sanitizer, the `TextDecoder("utf-8")` decode, and `charset="UTF-8"` in the challenge. Flagging only — not counted against this PR.

Solid, well-tested hardening. Approving.

pi0x:
Addressed the review nit in 6d3be92: tightened the scheme separator regex to `/^basic +(.+)$/i` (literal spaces only, matching RFC 9110's `1*SP`) instead of `\s+`. Auth tests and lint pass.

pi0 (approved):

PR #1471, opened 2026-07-14
fix(base): collapse leading-slash run in all base-stripping sites

pi0x:
Both the `mount` middleware (`src/h3.ts`) and `requestWithBaseURL` (`src/utils/request.ts`) stripped the base with a bare `.slice(base.length)`, so a request like `/api//evil.com` stripped to a protocol-relative `//evil.com`. A downstream handler that redirects to that pathname yields a protocol-relative open redirect (`withBase` already guarded against this via `withoutBase`).

This routes both sites through a shared internal `stripBase` helper that enforces a segment boundary and collapses the leading-slash run to a single slash. `requestWithBaseURL` now also skips stripping when `base` doesn't match on a boundary, aligning it with the mount middleware.

Regression tests added for the mount fetch path, the mounted-H3 middleware path, and `requestWithBaseURL`. The H3 gzip bundle limit is nudged (6.6kb -> 6.65kb) to fit the fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved base-path URL rewriting for mounted applications to safely normalize paths and avoid malformed/protocol-relative path outcomes.
  * Collapsed repeated leading slashes after base stripping to ensure consistent request handling.
  * Preserved pathnames unchanged when they don’t match the configured base.
* **Tests**
  * Added mount and request edge-case coverage for slash normalization and base-stripping behavior.
  * Slightly increased the H3 gzipped bundle-size benchmark allowance.

### DIFF
--- a/src/h3.ts
+++ b/src/h3.ts
@@ -4,6 +4,7 @@ import { HTTPError } from "./error.ts";
 import { toResponse, kNotFound } from "./response.ts";
 import { callMiddleware, normalizeMiddleware } from "./middleware.ts";
 import { requestWithBaseURL } from "./utils/request.ts";
+import { stripBase } from "./utils/internal/path.ts";
 
 import type { ServerRequest } from "srvx";
 import type { H3Config, H3CoreConfig, MatchedRoute, RouterContext } from "./types/h3.ts";
@@ -139,7 +140,10 @@ export const H3 = /* @__PURE__ */ (() => {
             ) {
               return next();
             }
-            event.url.pathname = event.url.pathname.slice(base.length) || "/";
+            // `stripBase` collapses the leading-slash run so `/base//evil.com`
+            // cannot strip to a protocol-relative `//evil.com` a downstream
+            // redirect could abuse (the boundary is already checked above).
+            event.url.pathname = stripBase(originalPathname, base);
             const restore = () => {
               event.url.pathname = originalPathname;
             };
--- a/src/utils/internal/path.ts
+++ b/src/utils/internal/path.ts
@@ -25,30 +25,26 @@ export function joinURL(base: string | undefined, path: string | undefined): str
   return base + path;
 }
 
-export function withoutBase(input: string = "", base: string = ""): string {
-  if (!base || base === "/") {
-    return input;
-  }
-  const _base = withoutTrailingSlash(base);
-  if (!input.startsWith(_base) || (input.length > _base.length && input[_base.length] !== "/")) {
-    return input;
-  }
-  // Collapse leading slashes to prevent protocol-relative URL injection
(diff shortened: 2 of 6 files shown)


pi0x (commented):
**APPROVE** — correct, minimal, well-tested fix for the protocol-relative open-redirect vector. Verified locally on the PR head.

### What I checked
- **Regression tests genuinely fail without the fix**: reverting only `src/` leaves all 6 new assertions red (mount fetch, mounted-H3 child middleware, `requestWithBaseURL` collapse + no-match), green with it. Both stripping sites covered.
- **Full suite green**: `pnpm vitest run test/` → 1463 passed / 39 skipped, `pnpm lint` clean (oxlint + oxfmt + typecheck).
- **Edge cases** (verified against a reference impl): `/api`+`/api`→`/`, `/api/`→`/`, `/apifoo`→unchanged (segment boundary holds), `/api//evil.com`→`/evil.com`, `/api///x`→`/x`, and crucially `/api/a//b`→`/a//b` (only the *leading* slash run collapses — deep multi-slash paths are preserved). Root/empty base are benign.
- **Query/URL consistency**: mount only rewrites `event.url.pathname` (search untouched); `requestWithBaseURL` re-serializes via `url.href`, so query survives. Good.

### Bundle-size bump is justified
Measured H3 minimal bundle: pre-fix **17334 / gzip 6594** → PR **17385 / gzip 6627**. New gzip exceeds the old 6600 guard, so the bump to 6650 is warranted; raw limit stays 17400 (now 15 bytes of headroom — tight but passing). Nit: the PR description says pre-fix gzip was "exactly 6600" — it actually measured 6594, but the bump is still correct.

### On `stripBase` vs the existing `withoutBase` (same file)
The two overlap heavily and I initially expected reuse. But I measured it: routing both sites through `withoutBase` is **larger** (17527 / gzip 6657) because it drags `withoutTrailingSlash` into the core bundle. So the compact purpose-built helper is the size-conscious choice here — the duplication is defensible. Two minor notes for the author's awareness (non-blocking):
- `stripBase` does **not** normalize a trailing-slash base (`stripBase('/api/foo', '/api/')` returns `/api/foo` unstripped, where `withoutBase` yields `/foo`). This is unreachable via the mount middleware (its outer guard skips trailing-slash bases) and matches the pre-existing quirk in the fetch-handler path, so no regression — but worth a comment if `stripBase` is ever reused elsewhere.
- In the mount middleware, `stripBase`'s internal boundary `if` re-checks what the outer `startsWith`/segment guard already proved, so it's always-true there — negligible, and keeps the helper self-contained. The per-request `replace(/^\/+/, "")` is a cached literal regex on matched routes only; no meaningful hot-path cost.

Nice defense-in-depth aligning all base-stripping sites with the `withoutBase` hardening. LGTM.

pi0 (approved):

pi0 (approved):

PR #1475, opened 2026-07-17
perf(middleware): precompose middleware chains

pi0x:
### 🔗 Linked issue

_none_

### 📚 Description

Middleware dispatch currently pays a per-request cost that only depends on registration-time state: `~getMiddleware` re-builds the middleware array on every request (an array spread whenever the route has its own middleware), and `callMiddleware` walks it recursively with index bookkeeping through a single megamorphic `fn(event, next)` call site.

Since both the global middleware list and each route's middleware are fixed between registrations, this PR precomposes the chains once and caches them:

- **`composeMiddleware(middleware)`** builds the chain right-to-left into a single `(event, handler) => result` function. Each layer is its own closure, so middleware call sites become monomorphic. The terminal handler is threaded through as an argument (rather than baked in or read from `event.context.matchedRoute`, which a nested app can overwrite mid-chain), so one composed global chain serves every route.
- **Per-route cache**: route middleware + handler are composed on first match and cached on the route object (`~composed`). Only route-local state is baked in, so the cache stays valid when `mount` copies route objects.
- **Per-app cache**: the composed global chain is cached on the app (`~composed`) and explicitly invalidated by `use()` and `mount()`, so middleware added after the first request is picked up. The `mount` wrapper reuses the mounted app's own `~composed` cache, so the child's `use()` invalidates it too.
- `defineHandler({ middleware })` now composes once at definition time instead of re-dispatching per call.

#### Public API

`composeMiddleware` (and the `ComposedMiddleware` type) is exported from `h3` so downstream frameworks (e.g. nitro) can adopt precomposition for their own middleware chains instead of `callMiddleware`. `callMiddleware` remains exported and unchanged.

#### Backward compatibility (`~getMiddleware`)

Existing users extending `H3Core` (e.g. [nitro](https://github.com/nitrojs/nitro/blob/main/src/build/virtual/app.ts#L151-L154)) override `~getMiddleware` — including as an instance-level assignment — to provide **per-event** middleware, which cannot be precomposed. The dispatch strategy is resolved once per app (`~dispatch`, built on first request and invalidated by `use()`/`mount()`): if `~getMiddleware` is overridden (method identity check, run only at dispatcher build time — no per-request cost), dispatch falls back to the previous `callMiddleware` behavior unchanged. Overrides must be in place before the first request (nitro assigns at build time). A regression test covers the nitro pattern.

`next()` memoization semantics (`undefined`/`kNotFound` fall-through, thenable handling) are preserved byte-for-byte; the shared per-layer logic lives in one `callLayer` helper used by both the composed chain and the public `callMiddleware`.

#### Numbers

Isolated dispatch bench (mitata, node 24):

| depth | vs `callMiddleware` | vs spread + `callMiddleware` | allocations |
| ----- | ------------------- | ---------------------------- | ----------- |
| 3     | 1.09x faster        | 1.13x faster                 | −25%        |
| 10    | 1.13x faster        | 1.18x faster                 | −21%        |

App-level `pnpm bench:node` is unchanged (dispatch is sub-µs of a ~31µs iteration dominated by Request/URL work); the win scales with middleware depth and route-middleware usage.

Bundle cost of shipping both the composer and the compat path: H3 +599 B (+191 B gzip), H3Core +539 B (+183 B gzip), defineHandler +103 B (+50 B gzip); budgets in `bundle.test.ts` bumped to match (kept tight).

Known edge (outside documented API): mutating a middleware array in place (a route's `middleware` after its first match, or `~middleware` directly without `use()`) is not picked up live — invalidation is explicit via `use()`/`mount()`. Custom `~getMiddleware` overrides are unaffected.

Possible follow-up: in the `h3-middleware` bench scenario the remaining overhead is dominated by `normalizeMiddleware`'s per-request route regexp matching and the `middlewareParams` spread — baking matchers into composition would be the next lever.

---

## Summary by CodeRabbit

* **New Features**
  * Added middleware composition utilities (`composeMiddleware` / `composeHandler`) and re-exported `ComposedMiddleware`.
* **Performance Improvements**
  * Reduced per-request middleware/handler overhead by caching precomposed route pipelines and composing mounted middleware once.
  * Updated handler middleware wrapping to use the new composition model.
* **Compatibility**
  * Preserved dynamic, per-event middleware behavior when overriding middleware resolution, with proper cache invalidation after `use()` and `mount()`.
* **Tests**
  * Added coverage for composition invalidation and dynamic middleware overrides.
  * Tightened bundle size benchmark limits and updated export snapshots.

### DIFF
--- a/src/h3.ts
+++ b/src/h3.ts
@@ -46,7 +46,8 @@ export class H3Core implements H3CoreType {
   "~middleware": Middleware[];
   "~routes": H3Route[] = [];
 
-  // Cached composition of `~middleware` (invalidated by `use()` and `mount()`)
+  // Cached dispatch and `~middleware` composition (invalidated by `use()` and `mount()`)
+  "~dispatch"?: (event: H3Event, route: MatchedRoute<H3Route> | void) => unknown | Promise<unknown>;
   "~composed"?: ComposedMiddleware;
 
   constructor(config: H3CoreConfig = {}) {
@@ -66,29 +67,7 @@ export class H3Core implements H3CoreType {
       event.context.params = route.params;
       event.context.matchedRoute = route.data;
     }
-    if (this["~getMiddleware"] !== H3Core.prototype["~getMiddleware"]) {
-      // Compat: a custom `~getMiddleware` (subclass or instance override, e.g. nitro)
-      // can return per-event middleware, which cannot be precomposed.
-      return callMiddleware(
-        event,
-        this["~getMiddleware"](event, route as unknown as undefined),
-        route?.data.handler || NoHandler,
-      );
-    }
-    let routeHandler: EventHandler = NoHandler;
-    if (route) {
-      const data = route.data;
-      // Route middleware and handler are fixed at registration: compose them once.
-      // The composed pair travels with the route object when copied (mount).
-      routeHandler = data.middleware?.length
-        ? (data["~composed"] ??= composeHandler(data.middleware, data.handler))
-        : data.handler;
-    }
-    const middleware = this["~middleware"];
-    if (middleware.length === 0) {
-      return routeHandler(event);
-    }
-    return (this["~composed"] ??= composeMiddleware(middleware))(event, routeHandler);
+    return (this["~dispatch"] ??= createDispatcher(this))(event, route);
   }
 
   "~request"(request: ServerRequest, context?: H3EventContext): Response | Promise<Response> {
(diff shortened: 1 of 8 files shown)


pi0x:
Follow-up for downstream adoption: nitrojs/nitro#4443 (drop the `~getMiddleware` override in favor of `composeMiddleware` + h3's precomposed fast path).

pi0 (approved):

Issue #1478, opened 2026-07-18
createEventStream: onClosed/autoclose never fire on client disconnect in non-Node runtimes


pi0x:
Client-disconnect cleanup for `createEventStream` only works on Node — on web-standard runtimes (Deno, Bun, Workerd, service workers, plain `app.fetch()` consumers) neither `autoclose` nor `onClosed` ever fires when the client disconnects. Found during a pre-stable validation pass, reproduced against current `main` (75ac47d).

## `autoclose` is wired exclusively to the Node runtime

`src/utils/internal/event-stream.ts:31-33`:

```ts
if (opts.autoclose !== false) {
  this._event.runtime?.node?.res?.once("close", () => this.close());
}
```

On every non-Node runtime `runtime?.node?.res` is undefined, so a client disconnect never auto-closes the stream.

## `onClosed` callback is swallowed on client disconnect

`src/utils/internal/event-stream.ts:165-167`:

```ts
onClosed(cb: () => any): void {
  this._writer.closed.then(cb).catch(_noop);
}
```

When the client disconnects, the readable side of the `TransformStream` is **cancelled**, which puts the writable side into an **errored** state — so `writer.closed` **rejects**, the callback is routed into `.catch(_noop)`, and never runs. It only fires on a graceful `close()`.

## Net effect

The exact cleanup pattern the `createEventStream` docs recommend silently does nothing on client disconnect on non-Node runtimes:

```ts
const eventStream = createEventStream(event);
const interval = setInterval(() => eventStream.push("tick"), 1000);
eventStream.onClosed(() => {
  clearInterval(interval); // never runs on web/Deno/Bun/Workerd disconnect
});
return eventStream.send();
```

Verified on the web path: after `reader.cancel()` on the response body, `onClosed` never fired and the interval kept ticking (15+ pushes after disconnect). Timers, DB handles, and generator loops leak per connection.

Event framing itself (multi-line `data:`, `id`/`event`/`retry`, comments, sanitization) checked out fine — this is purely the disconnect/cleanup path.

A portable fix likely means detecting cancellation of the readable side (e.g. observing `writer.closed` rejection as a close signal, and/or a `cancel` hook on the readable) rather than relying on `node.res` `close`.

PR #1488, opened 2026-07-18
feat: add `onDispose` hook

### LINKED ISSUE #1487: Add `event.onDispose()`: a portable end-of-event hook for resource cleanup
Spun out of #1478 / #1484. That bug was SSE-specific, but the underlying gap is general — though it's narrower than "no lifecycle signal at all":

- **Aborted** (client vanished mid-response) is already portable today via **`event.req.signal`**: srvx lazily wires a Node `AbortController` to `res` `"close"` (aborting only on premature close), and on Deno/Bun/Workers it's the native request signal. `proxy` already relies on this.
- **Finished** (h3 is done; nothing more will be written) has **no signal on any runtime except Node** (`res` `"close"`), and there is no single place where both halves converge.

So every util needing cleanup-on-end either goes Node-only, or grows its own detection plus its own callback registry — and gets both wrong (see #1478).

## The trap: "after `toResponse`" is not "end of event"

The obvious implementation — a `.finally()` on the response promise in `src/h3.ts` — is wrong. Measured on an SSE route via `app.fetch()` (reproduced independently: 3ms vs 154ms on a shorter stream):

```
   3ms  app.fetch() promise RESOLVED  <-- a naive dispose hook would fire here
 304ms  body fully consumed
 305ms  stream actually finished (onClosed)
```

Response production happens ~100x too early on any streaming response. A hook wired there would fire *while the stream is live*, breaking precisely the utilities it is meant to serve.

## Achievable fidelity

If core observes the body stream:

| case | Node | Web runtimes |
|---|---|---|
| streaming body, normal end | accurate (`res` close) | accurate (body close) |
| streaming body, client disconnect | accurate | accurate (body cancel) |
| non-streaming body | accurate | **approximate** — fires at response production |
| body never consumed by the caller | accurate | never fires |

The last two cells are irreducible: once the `Response` is handed to the runtime, bytes-landed is not observable on web. The API should therefore be documented as *"the handler is done with this event"*, not *"the client received it"*.

## Proposal

```ts
event.onDispose((reason?: unknown) => {
  clearInterval(interval);
});
```

- **Opt-in and lazily armed.** Core checks one symbol on the event; body observation happens only if something registered. A plain JSON route pays nothing.
- **`reason` argument:** `undefined` on normal completion, the cancel/abort reason otherwise. Costs nothing now; tracing/metrics consumers (delivered vs. aborted) will need it, and retrofitting an argument later is painful.
- **Contract:**
  - Fires exactly once, after global `onResponse`, on every termination path: normal end, client cancel, **and body error** (a source stream throwing mid-body must not skip disposal).
  - Callbacks run in registration order; sync throws and async rejections are both absorbed, but reported via `console.error` unless `config.silent` — a cleanup callback that always throws should be discoverable in dev.
  - The aggregate callback promise is passed to `event.waitUntil()` — on Workers, execution after the body finishes isn't otherwise guaranteed.
  - Registering after disposal fires immediately. Registrations are only *guaranteed* when made during request handling (handler / middleware / `onResponse`): on web runtimes, a first registration made mid-stream (e.g. from a timer after the response was produced) can never arm the observer. Node is immune.

### Implementation notes

- **Observe via the `pipeTo` promise, not transformer hooks.** `flush` misses errors, `cancel` misses normal end, and the transformer `cancel` hook is a newer spec addition with uneven runtime support. One construct covers all three termination paths portably:
  ```ts
  const { readable, writable } = new TransformStream();
  body.pipeTo(writable).then(fire, fire); // normal end, consumer cancel, source error
  ```
  One queue hop per chunk, no copy. Worth benchmarking before committing, but it's opt-in per event.
- **Decide whether to wrap at body-production time, not on the finished `Response`.** On Node, srvx's `FastResponse` keeps the raw body and fast-paths string/`Uint8Array` in `_toNodeResponse`; touching `.body` to ask "is this a stream?" materializes a native `Response` and kills that fast path. (On web the check is useless anyway — `new Response("hi").body` is always a `ReadableStream`.) Core already knows the real body type inside `prepareResponseBody`: stream → wrap; string/buffer/null → fire post-`onResponse`; **Node → skip wrapping entirely** and use `res` `"close"`, which also keeps the non-streaming row accurate.

### Error-handling hardening is half the point

#1484 (still open) initially had a `try { cb() } catch {}` that silently failed to catch async rejections, turning a user's rejected cleanup into a process-fatal unhandled rejection. Every util that grows its own callback registry will hit that same rake. One hardened implementation in core is the argument for this feature as much as portability is.

### Why a callback rather than a new `event.signal`

The abort signal already exists as `event.req.signal` — that stays the way to react to a disconnect *mid-flight* (e.g. aborting an upstream fetch). `onDispose` is the converged end-of-event boundary: finish *or* abort. Overloading an `AbortSignal` to also mean "completed successfully" misleads more than the composability buys, and a signal (or callback, or anything else) is derivable from the core hook in a few lines.

An alternative shape worth weighing: **`event.disposed: Promise<unknown>`** (lazy, never-rejecting, resolves with the reason). Late registration, exactly-once, and hardened error absorption all fall out of promise semantics for free; `onDispose` sugar is then one line. Either shape works — the contract above is the substance.

(A note on `Symbol.asyncDispose`/`using`: scope-based disposal fires at *handler return*, which is exactly the too-early timing measured above — so no meaningful alignment there.)

## Consumers

- **`createEventStream`** — the motivating consumer. Replaces the Node-only `res.once("close")` with `event.onDispose(() => this.close())`, and fixes the web-only leak where a stream that is created but never `send()`-ed never runs its `onClosed` callbacks (pre-existing; verified present on `app.fetch()`, absent on Node; not covered by #1484). Caveat per the fidelity table: a response nobody ever reads still leaks on web.
- **`proxy` / `serveStatic` / tracing** — possible later adopters, case-by-case. Proxy in particular already gets teardown mostly for free (it forwards `req.signal` upstream, and runtime cancellation of the returned body propagates up the stream chain), so land `createEventStream` first rather than promising these.

## Notes

- This is an additive public API on `H3Event`, so it is cheapest to land **before v2 stable**.

Happy to prototype this if the shape looks right.

pi0x:
Resolves #1487 (prototype of the refined proposal). Related: #1478 / #1484.

## What

You start a timer, subscribe to something, or open an upstream connection inside a request handler. Where do you clean it up? Today the answer depends on where your app runs: on Node.js you can listen to `res.on("close")`, on Bun / Deno / Cloudflare Workers there is no reliable place at all — so timers keep ticking after clients are long gone.

`onDispose(event, cb)` is one answer that works on every runtime:

```ts
import { onDispose } from "h3";

app.get("/sse", (event) => {
  const interval = setInterval(() => sse.push("tick"), 1000);

  onDispose(event, () => clearInterval(interval));

  // ... return a streaming response
});
```

The callback runs **exactly once**, when the request is truly over — the response finished streaming, the client disconnected, or something errored mid-stream. On disconnect/error, it receives the reason.

Like other h3 utils (`getQuery(event)`, `readBody(event)`), it's just an importable function — apps that never use it don't ship a single byte of it.

## How

- **Node** (`event.runtime.node.res` present): observes `res` `"close"` — accurate for streaming *and* non-streaming bodies, and the body is never wrapped, so srvx's `FastResponse` `_toNodeResponse` fast paths stay intact. Premature close disposes with an `AbortError` `DOMException` (or `res.errored`).
- **Web runtimes**: a streaming body is piped through an identity `TransformStream`; the `pipeTo` promise settles on all three termination paths (normal end → resolve, consumer cancel → reject with the cancel reason, which also propagates upstream, source error → reject). Bodyless responses dispose at response production — once the `Response` is handed to the runtime, delivery is not observable on web (per the fidelity table in #1487).
- **Opt-in, lazily armed, and tree-shakable**: the first `onDispose` registration installs an `observe` function on the event; `toResponse` only keeps a symbol check + `observe` call at its single final exit (after `onResponse` resolves, so the hook ordering holds on every path). Because registration carries the machinery, apps that never import `onDispose` don't bundle any of it — this is why it is a util rather than an `H3Event` method (a method would anchor ~1 kB min in every core bundle).
- **Hardened once, in core**: callbacks run in registration order; sync throws and async rejections are both absorbed (reported via `console.error` unless `silent`) — the exact rake found in #1484. Pending async callbacks are handed to `event.waitUntil()` for serverless runtimes. Registering after disposal fires immediately with the recorded reason.

## Consumer: `createEventStream`

The Node-only `res.once("close", () => this.close())` autoclose is replaced with `onDispose(event, () => this.close())`. This fixes two pre-existing web-runtime leaks (verified failing on `main`, matrix `web` target):

- a stream that is **created but never `send()`-ed** never ran its `onClosed` callbacks (not covered by #1484);
- a client disconnect never triggered autoclose, so pushes kept running per connection.

Note: `onClosed`'s own disconnect-path bug (`writer.closed` rejection swallowed) remains #1484's scope — this PR intentionally doesn't touch it.

## Costs

- Bundle: **+~105 B min / +~42 B gzip** on the tracked core bundles (the symbol + arm check in `toResponse`); the ~1 kB machinery is bundled only where `onDispose` (or `createEventStream`) is imported. Limits ratcheted with the same tight headroom as before (`main` sat 32 B under).
- Runtime: one symbol check per request when unused; the per-chunk queue hop exists only for events that registered a callback *and* have a streaming body on a non-Node runtime.

## Tests

`test/dispose.test.ts` (matrix, 8 cases): non-streaming fire, mid-stream non-fire + fire on full consumption, disconnect reason propagation (web: exact cancel reason; node: abort error), `onResponse` ordering, callback order + sync/async error absorption, late registration, unmodified fast path, status/header preservation on wrapped responses. Plus the two SSE regression tests above. Full suite green (1543 passed), lint/typecheck/build clean.

## Caveats / open questions

- Registration is only guaranteed to observe the end of the event when made during request handling — on web runtimes a *first* registration made mid-stream can never attach the observer (documented; Node is immune).
- `EventStream` autoclose now depends on the event flowing through `toResponse` (true for any H3 app). A consumer invoking handlers directly without `toResponse` loses the old Node-only fallback.
- Body-wrapping perf on the streaming path is unbenchmarked beyond "opt-in only"; happy to add a mitata case if wanted.
- `event.disposed: Promise<unknown>` remains a possible alternative/additional shape (see #1487) — trivial to derive from this registry if preferred (though as an `H3Event` member it would re-anchor the machinery in core bundles).
- Naming: `onDispose(event, cb)` sits next to the `onRequest`/`onResponse`/`onError` *middleware factories* in the public namespace despite having a different shape (per-event util). Open to a different name if that reads as a footgun.

## Summary by CodeRabbit

* **New Features**
  * Added `onDispose` to register one-time per-event cleanup callbacks that run when the event/response lifecycle fully ends, including cancellation and errors.
  * Disposal is now integrated with the response flow to ensure the returned response participates in end-of-event cleanup.
* **Documentation**
  * Added `onDispose` docs covering argument semantics, callback timing relative to `onResponse`, late registration, and failure handling.
* **Bug Fixes**
  * Improved end-of-stream handling for event streams and SSE, including streams that are never sent and automatic cleanup on client disconnect.
* **Tests**
  * Added disposal and SSE lifecycle test coverage, updated export snapshots, and tightened bundle-size benchmark thresholds.

### DIFF
--- a/src/response.ts
+++ b/src/response.ts
@@ -1,6 +1,7 @@
 import { FastResponse } from "srvx";
 import { HTTPError } from "./error.ts";
 import { isJSONSerializable } from "./utils/internal/object.ts";
+import { armDispose, kEventDispose } from "./utils/internal/dispose.ts";
 
 import type { H3Config } from "./types/h3.ts";
 import { kEventRes, kEventResHeaders, kEventResErrHeaders, type H3Event } from "./event.ts";
@@ -26,9 +27,16 @@ export function toResponse(
   }
 
   const { onResponse } = config;
-  return onResponse
-    ? Promise.resolve(onResponse(response as Response, event)).then(() => response)
-    : response;
+  if (onResponse) {
+    return Promise.resolve(onResponse(response as Response, event)).then(() =>
+      (event as any)[kEventDispose]
+        ? armDispose(response as Response, event)
+        : (response as Response),
+    );
+  }
+  return (event as any)[kEventDispose]
+    ? armDispose(response as Response, event)
+    : (response as Response);
 }
 
 export class HTTPResponse {
--- a/src/response.ts
+++ b/src/response.ts
@@ -29,18 +29,19 @@ export function toResponse(
   const { onResponse } = config;
   if (onResponse) {
     return Promise.resolve(onResponse(response as Response, event)).then(() =>
-      observeEventDispose(event, response as Response),
+      observeEventDispose(event, response as Response, val),
     );
   }
-  return observeEventDispose(event, response as Response);
+  return observeEventDispose(event, response as Response, val);
 }
 
 // End-of-event observation (`onDispose` util). The first registration installs
 // the observer setup on the event — core only pays this symbol check, and the
-// machinery tree-shakes out of apps that never import `onDispose`.
-function observeEventDispose(event: H3Event, response: Response): Response {
+// machinery tree-shakes out of apps that never import `onDispose`. The raw
+// handler value is passed along so buffered bodies can skip stream wrapping.
+function observeEventDispose(event: H3Event, response: Response, val: unknown): Response {
   const state = (event as any)[kEventDispose] as DisposeState | undefined;
-  return state ? state.observe(response) : response;
+  return state ? state.observe(response, val) : response;
 }
 
 export class HTTPResponse {
--- a/src/response.ts
+++ b/src/response.ts
@@ -28,16 +28,20 @@ export function toResponse(
 
   const { onResponse } = config;
   if (onResponse) {
-    return Promise.resolve(onResponse(response as Response, event)).then(() =>
-      observeEventDispose(event, response as Response, val),
+    return Promise.resolve(onResponse(response as Response, event)).then(
+      () =>
+        ((event as any)[kEventDispose] as DisposeState | undefined)?.observe(
+          response as Response,
+          val,
+        ) ?? (response as Response),
     );
   }
-  return observeEventDispose(event, response as Response, val);
-}
-
-function observeEventDispose(event: H3Event, response: Response, val: unknown): Response {
-  const state = (event as any)[kEventDispose] as DisposeState | undefined;
-  return state ? state.observe(response, val) : response;
+  return (
+    ((event as any)[kEventDispose] as DisposeState | undefined)?.observe(
+      response as Response,
+      val,
+    ) ?? (response as Response)
+  );
 }
 
 export class HTTPResponse {
--- a/src/response.ts
+++ b/src/response.ts
@@ -1,7 +1,7 @@
 import { FastResponse } from "srvx";
 import { HTTPError } from "./error.ts";
 import { isJSONSerializable } from "./utils/internal/object.ts";
-import { armDispose, kEventDispose } from "./utils/internal/dispose.ts";
+import { kEventDispose, type DisposeState } from "./utils/internal/dispose.ts";
 
 import type { H3Config } from "./types/h3.ts";
 import { kEventRes, kEventResHeaders, kEventResErrHeaders, type H3Event } from "./event.ts";
@@ -29,14 +29,18 @@ export function toResponse(
   const { onResponse } = config;
   if (onResponse) {
     return Promise.resolve(onResponse(response as Response, event)).then(() =>
-      (event as any)[kEventDispose]
-        ? armDispose(response as Response, event)
-        : (response as Response),
+      armEventDispose(event, response as Response),
     );
   }
-  return (event as any)[kEventDispose]
-    ? armDispose(response as Response, event)
-    : (response as Response);
+  return armEventDispose(event, response as Response);
+}
+
+// End-of-event observation (`onDispose` util). The first registration installs
+// the arming logic on the event — core only pays this symbol check, and the
+// machinery tree-shakes out of apps that never import `onDispose`.
+function armEventDispose(event: H3Event, response: Response): Response {
+  const state = (event as any)[kEventDispose] as DisposeState | undefined;
+  return state ? state.arm(response) : response;
 }
 
 export class HTTPResponse {
--- a/src/response.ts
+++ b/src/response.ts
@@ -29,18 +29,18 @@ export function toResponse(
   const { onResponse } = config;
   if (onResponse) {
     return Promise.resolve(onResponse(response as Response, event)).then(() =>
-      armEventDispose(event, response as Response),
+      observeEventDispose(event, response as Response),
     );
   }
-  return armEventDispose(event, response as Response);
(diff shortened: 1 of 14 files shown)


pi0x (commented):
.

src/response.ts
```diff
@@ -26,9 +27,21 @@ export function toResponse(
...
+      observeEventDispose(event, response as Response, val),
+    );
+  }
+  return observeEventDispose(event, response as Response, val);
+}
+
+// End-of-event observation (`onDispose` util). The first registration installs
+// the observer setup on the event — core only pays this symbol check, and the
+// machinery tree-shakes out of apps that never import `onDispose`. The raw
+// handler value is passed along so buffered bodies can skip stream wrapping.
```
pi0:
```suggestion
```

pi0 (approved):

Issue #1437, opened 2026-07-05
RFC: evolving `defineValidatedHandler` / `defineRoute` into a richer validated-routing core and types


sandros94:
### Describe the feature

I've been building [`h3-route-tools`](https://github.com/sandros94/h3-typed-routes) on top of h3 v2 — essentially a playground for a fully-typed, validated-routing experience. A good chunk of it grew straight out of the Standard Schema work (#963/#964, #1146) as well as @productdevbook implementations (#1143, #1237 and #1339), and by now a few pieces feel like they might belong upstream rather than living in a side package. I'd like to open the discussion before we're out of the rc window, so there's still room for the breaking changes some of this would imply.

Scope up front, so nobody panics at the diff: I'm **not** proposing to move the typed-fetch client, the codegen, or the Nitro module into core, and OpenAPI I'd specifically want to _stay_ an opt-in plugin (happy for it to become a first-party one down the line). What I think is worth discussing for core is the **validated-handler surface itself**: `params` + `response` validation, per-method dispatch, uniform async validation and the contract sidecar (definitions and types).

Core already ships `defineValidatedHandler` (experimental), `defineRoute`, and the `getValidated*` helpers. The h3-route-tools versions are supersets of those same APIs — which is exactly why the diff looks scary at a glance. This issue is my attempt to make it legible.

### On async validation

The one concrete gap I keep bumping into: the declarative `defineValidatedHandler` path can't run async schemas for `query`/`headers`. `syncValidate` throws `Asynchronous validation is not supported for ${source}` the moment a Standard Schema returns a `Promise` — only `body` is awaited (lazily, via `.json()`), and `params`/`response` aren't validated there at all. The standalone `getValidated*` helpers _do_ await, so the capability is already in core — it just isn't wired through the handler.

In h3-route-tools everything runs through a single async `validateData`, so async refinements work uniformly across `params`, `query`, `headers`, `body` **and** `response` — e.g. an async uniqueness/remote lookup on a query param, or a DB-backed check inside a response schema. That uniformity is the piece I'd most like to see land upstream.

```ts
// works in h3-route-tools; the query path throws in core's defineValidatedHandler today
defineRouteHandler({
  params: v.object({ id: v.pipe(v.string(), v.transform(Number)) }),
  get: {
    validate: {
      query: v.pipeAsync(v.object({ slug: v.string() }), v.checkAsync(isUnique)),
      response: User, // validated too — 500 on contract breach, not 400
    },
[...]

PR #1491, opened 2026-07-20
feat(validate): support async validation within `defineValidatedHandler`

sandros94:
Discussed in #1437

This is the first PR of hopefully a few related to improving validation strictness and related type support within h3's toolset.

In this change I solely focus on adding async support for query and header validaiton within `defineValidatedHandler` that were previously sync-only.

## Summary by CodeRabbit

* **New Features**
  * Added support for asynchronous validation of request bodies, headers, and query parameters.
  * Async validation results are now correctly awaited before handlers run.

* **Bug Fixes**
  * Invalid asynchronously validated inputs now return the expected validation errors and HTTP 400 responses.
  * Preserved existing custom validation error behavior.

### DIFF
--- a/src/utils/internal/validate.ts
+++ b/src/utils/internal/validate.ts
@@ -75,7 +75,7 @@ export async function validateData<T>(
 // prettier-ignore
 const reqBodyKeys = /* @__PURE__ */ new Set(["body", "text", "formData", "arrayBuffer"]);
 
-export function validatedRequest<
+export async function validatedRequest<
   RequestBody extends StandardSchemaV1,
   RequestHeaders extends StandardSchemaV1,
 >(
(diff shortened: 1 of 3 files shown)


pi0 (approved):

pi0:
Only can you double check it is all compliant with standard-schema?

sandros94:
> Only can you double check it is all compliant with standard-schema?

yes, `~standard.validate()` can both be sync and async, depending on how the schema validation was constructed/used.

I've added some new zod tests that do have a async refines, which forces `~standard.validate()` to be a promise

EDIT: the [upstream standard-schema example](https://standardschema.dev/schema#how-to-only-allow-synchronous-validation) for sync-only usage is for libs that might not support async operations, like some sort of hooks, but we have no such limitation

src/utils/internal/validate.ts
```diff
@@ -169,16 +169,13 @@ export function validatedURL(
...
 
-function syncValidate<Source extends "headers" | "query", T = unknown>(
+async function validateSource<Source extends "headers" | "query", T = unknown>(
   source: Source,
   data: unknown,
   fn: StandardSchemaV1<T>,
   onError?: OnValidateError,
-): T {
-  const result = fn["~standard"].validate(data);
-  if (result instanceof Promise) {
```
pi0, line 179: We could do later but Promise chaining should be prefered over forced `async` keyword (it forces code to resume in next tick even if not needed to!) part of why H3Core is fast is that we avoid async/await

Issue #1482, opened 2026-07-18
Pre-stable validation: smaller findings (response edge cases, type/export drift, deprecated typings)


pi0x:
Umbrella for the remaining smaller findings from the pre-stable validation pass (all verified against `main` @ 75ac47d). Larger items are tracked separately: #1477 (response preparation), #1478 (SSE disconnect), #1479 (non-Error throws), #1481 (`!val.ok` header drop). The double-`readBody` Node hang is being addressed upstream in srvx.

### Response handling

- [x] **HEAD body stripping misses the in-place merge path** — `src/response.ts:129-131`: only the fast path (`:120`) and the immutable-headers catch path (`:134`) strip HEAD bodies. When prepared headers merge in place into a mutable 2xx `Response`, the body is returned intact — `app.fetch`/web-mode/service-worker consumers see a HEAD response with a body (Node strips it at the socket), contradicting the "keep self-consistent" intent at `:119`.
- [x] ~~**`kHandled` ignores `event.res` state** — `src/response.ts:62-63`: returns a bare `FastResponse(null)` before `preparedRes` is read or `kEventRes` cleared. A handler that sets `event.res.status = 201` + headers then returns `kHandled` produces a 200 with no headers, and `onResponse` hooks observe a synthetic 200 that doesn't match what was actually sent.~~ we can't alter node.js responses in reality
- [x] **Throwing `onResponse` hook fails the request raw** — `src/response.ts:30`: `Promise.resolve(onResponse(...)).then(() => response)` has no rejection handling, so a hook that throws turns the request into an unhandled rejection and `onError` is never invoked. If intentional (avoiding hook recursion), it's undocumented.

### Exports / types drift

- [x] **`kNotFound` / `kHandled` not exported from `src/index.ts`** ~~they're part of the documented handler return contract (AGENTS.md), but users can only reach them via `Symbol.for("h3.notFound")`. Either export them or.~~ fix the agents.md docs
- [x] **Declared `H3Core` type is missing `"~dispatch"` and the static `"~h3"` brand** — `src/types/h3.ts:102` gained `"~composed"` in #1475 but not `"~dispatch"` (`src/h3.ts:50`), and the exported `H3` is cast to the declared type, so TS consumers (e.g. nitro) can't see or invalidate the dispatch cache the way `use()`/`mount()` do. The static `"~h3"` brand (`src/h3.ts:42`, used by `src/handler.ts` for instance detection) is also absent from the declared class.
[...]

PR #1492, opened 2026-07-21
fix(deprecated): correct v1 signatures in the compat shim

n0liu:
Addresses two of the deprecated-typing items from #1482.

`appendResponseHeaders` (and its `appendHeaders` alias) declares `headers: string`, but the body iterates the argument with `Object.entries`. As typed the export is unusable: passing the record it actually wants is a type error, and passing the declared `string` yields index-keyed garbage headers. Its sibling `setResponseHeaders`/`setHeaders` already takes `Record<string, string>`, so this brings the append pair in line with it.

`sendRedirect` required `code`, but v1 defaulted to 302 and the delegated `redirect()` still has `status = 302`. `sendNoContent` directly above it already declares `code?: number`, so the two are consistent again.

Both changes are type-only — runtime behaviour is untouched and there is no bundle impact.

Type tests are added to `test/unit/types.test-d.ts`. They fail on `main` (`TS2345` on the record argument, twice, and `TS2554` for the missing `code`) and pass with the change. Locally `vitest --run` reports 55 files / 1561 passed with `Type Errors  no errors`, and `oxlint` plus `oxfmt --check` are clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated deprecated redirect handling to allow the status code to be omitted.
  * Improved deprecated header helpers to accept multiple headers as key-value pairs.
  * Updated type definitions and validation to reflect the revised deprecated API signatures.

### DIFF
--- a/src/_deprecated.ts
+++ b/src/_deprecated.ts
@@ -130,7 +130,7 @@ export const sendNoContent: (event: H3Event, code?: number) => HTTPResponse = (_
   noContent(code);
 
 /** @deprecated Please use `return redirect(event, code)` */
-export const sendRedirect: (event: H3Event, location: string, code: number) => HTTPResponse = (
+export const sendRedirect: (event: H3Event, location: string, code?: number) => HTTPResponse = (
   _,
   loc,
   code,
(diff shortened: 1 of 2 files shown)


pi0 (approved):

n0liu:
Thanks for the quick review and merge! I'll pick up more items from #1482.

PR #1500, opened 2026-07-22
perf(body-limit): stream enforcement instead of pre-buffering

### LINKED ISSUE #1498: bodyLimit/assertBodySize: stream enforcement instead of pre-buffering the body
## Summary

`bodyLimit()` / `assertBodySize()` currently verify request body size by **pre-buffering** the whole body (up to `limit`) before the handler runs. This adds latency and memory overhead, and blocks streaming handlers. We can keep the exact same byte-accurate security guarantee while removing the pre-buffering, by counting bytes **inline as they flow** instead of ahead of time.

## Current behavior

`isBodySizeWithin()` (`src/utils/body.ts:204-237`):

1. If an honest `Content-Length` exceeds the limit → fail-fast (rejected before the handler runs). ✅
2. `Content-Length` + `Transfer-Encoding` together → `400` (RFC 7230 anti-smuggling). ✅
3. **Otherwise it always drains a `req.clone()` stream to completion (up to `limit`) before returning.**

Step 3 is the problem. Because `bodyLimit` is middleware that awaits `assertBodySize` **before** calling `next()`:

```ts
// src/utils/middleware.ts
return async (event, next) => {
  await assertBodySize(event, limit);
  return next();
};
```

...a streaming handler behind `bodyLimit()` cannot start until the entire body has arrived and been buffered, and up to `limit` bytes are held in memory per in-flight request.

> Enforcement itself is correct — no chunked-encoding bypass; CL+TE rejected with `400`. This issue is purely about the latency/memory characteristics.

## Why we can't just trust `Content-Length` (fast-path rejected)

An obvious idea is: if `Content-Length <= limit`, skip the read. **This is unsafe.** `assertBodySize` operates on a web-standard `Request`, where the `content-length` header and the body stream are independent fields — nothing guarantees they agree. While a conformant HTTP/1.1 or HTTP/2 server bounds the body to the declared length via transport framing, h3 also receives Requests from adapters, proxies, `mockEvent`, service workers, and synthetic construction, where the header can claim `5` while the stream delivers gigabytes. Trusting the header would open a bypass in exactly those cases. The guard must stay **byte-accurate**, not header-based.

## Proposed improvement — streaming enforcement (no pre-buffer)

Byte-accuracy does **not** require pre-buffering — count actual bytes as they flow, not ahead of time:

- Keep the cheap fail-fast `413` for an honest oversized `Content-Length`, and keep the CL+TE `400`.
- For everything else, wrap `event.req.body` in a counting `TransformStream` that errors the moment the running total exceeds `limit`, and swap in a new `Request` carrying the wrapped stream. Bytes are counted as the handler reads them.

Result: no `clone()`, no pre-buffer, memory bounded by what the handler itself holds, and streaming handlers start immediately. Same security guarantee (real bytes verified; a lying-small `Content-Length` is still caught mid-stream).

## Tradeoffs to accept / document

1. **Overflow surfaces mid-stream, not as a pre-handler `413`.** For chunked / unknown-length bodies you can't know the size is too big until you've read too much, so an overflow becomes a stream error mid-read (aborted stream) rather than a tidy `413`. The fail-fast `413` is preserved wherever an honest `Content-Length` is present.
2. **Enforcement becomes tied to body consumption.** Today `bodyLimit` reads the clone itself, so it enforces even if the handler ignores the body; streaming enforcement only counts what the handler reads. This is unavoidable — "enforce independently of the handler" *is* the buffering we're removing. It's benign (an unread body consumes no app memory and is bounded by transport backpressure) but should be documented.

## Scope

- [ ] Counting `TransformStream` wrapper + `event.req` replacement in `assertBodySize` / `bodyLimit`
- [ ] Preserve fail-fast `413` (honest oversized `Content-Length`) and CL+TE `400`
- [ ] Tests: content-length fast-fail, chunked overflow (mid-stream error), CL+TE `400`, and a streaming handler that starts before the full body arrives
- [ ] Docs note (the current buffering caveat in `docs/2.utils/1.request.md` / `9.more.md` was added as an interim measure and would be updated)

Happy to open a PR for review.

pi0x:
Closes #1498.

## What

`bodyLimit()` / `assertBodySize()` verified request body size by **draining a cloned request stream** before the handler ran — adding latency, holding up to `limit` bytes in memory per in-flight request, and blocking streaming handlers. This switches to **streaming enforcement** using srvx's pull-based limiter (now publicly exported at `srvx/body-limit`), keeping the exact same byte-accurate guarantee with no pre-buffering.

## How

- **`assertBodySize`** wraps `event.req.body` with `limitBodyStream(...)` and swaps in a rebuilt `Request`, so bytes are counted **as the handler reads them** — no `clone()`, no pre-buffer, streaming handlers start immediately.
  - srvx's `limitRequestBody` can't be used directly: its internal `new Request(request, ...)` throws on srvx's Node `ServerRequest` (`Cannot read private member #state`). Instead we wrap `req.body` with `limitBodyStream` and rebuild a native `Request`, re-attaching srvx's runtime augmentation (`runtime` / `waitUntil` / `ip` / `context`) — mirroring `createSubRequest` in `proxy.ts`. Verified `event.runtime` and `getRequestIP` keep working after the swap.
- **Preserved & still synchronous:** fail-fast `413` for an honest oversized `Content-Length`, and `400` for `Content-Length` + `Transfer-Encoding` (RFC 7230 smuggling).
- **`assertBodySize` is now synchronous** (`void`) — no draining left to await.
- **Error mapping:** a deferred overflow surfaces as srvx's `ERR_BODY_TOO_LARGE`. It's mapped to a **handled `413`** in `toResponse` (so it isn't logged as an unhandled server error), and `readBody`'s `formData()` catch re-throws it instead of masking it as `400`.

## Trade-offs (documented in JSDoc)

1. Overflow on a chunked / unknown-length body surfaces **mid-read** rather than as a pre-handler `413` (the fail-fast `413` is kept wherever an honest `Content-Length` is present).
2. Enforcement is tied to **consumption** — a body the handler never reads is not counted (benign; bounded by transport backpressure).

## Notes / things to review

- **srvx bumped to `^0.12.2`** (adds the `srvx/body-limit` entry point).
- **H3Core bundle guardrail** nudged `7550 → 7600` bytes (`3000 → 3050` gzip): the `ERR_BODY_TOO_LARGE` check sits in the `toResponse` hot path (~13B). Flagging for a conscious ok.
- The oversized-`formData()` integration test is **node-target only**: an in-process `FormData` body is a lazy undici stream that emits an unhandled "already closed" rejection when the multipart parse is aborted mid-flight — an artifact of the in-memory source, not of real socket-backed requests (verified clean over a real server).

## Tests

New `test/body-limit.test.ts` (matrix) + rewritten `test/unit/body-limit.test.ts` cover: fail-fast `413`, chunked overflow surfacing mid-stream, lying-small `Content-Length`, CL+TE `400`, `formData` → `413` mapping, and an unread body not being counted. Full suite: 1592 passed, lint + typecheck clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Enforced request body limits while streaming/consuming the body, with oversized payloads consistently returning **`413 Request Entity Too Large`**.
  - Preserved existing limit errors as-is (instead of converting them to generic parsing/invalid-body errors) across form, JSON, validation, and JSON-RPC flows.
  - Added fail-fast handling for conflicting **`Content-Length`** and **`Transfer-Encoding`** scenarios to avoid request smuggling.
- **Documentation**
  - Updated body-limit and body-size enforcement docs to precisely describe error timing and streaming behavior.
- **Tests**
  - Added/expanded matrix and unit tests to cover buffered, streaming, chunked, multipart, and edge header cases.

### DIFF
--- a/docs/2.utils/1.request.md
+++ b/docs/2.utils/1.request.md
@@ -14,7 +14,7 @@ icon: material-symbols-light:input
 
 Asserts that the request body size is within the specified limit.
 
-The limit is enforced **as the body is read**, not by pre-buffering: the request body stream is wrapped in a byte counter (via {@link limitRequestBody}) that aborts with a `413` Request Entity Too Large error the moment the running total exceeds `limit`. This preserves the byte-accurate guarantee (a lying-small `Content-Length` is still caught mid-stream) without holding the body in memory or blocking streaming handlers.
+The limit is enforced **as the body is read**, not by pre-buffering: the request is wrapped by srvx's `limitRequestBody`, which counts bytes as they flow and aborts with a `413` (`ERR_BODY_TOO_LARGE`) the moment the running total exceeds `limit`. This preserves the byte-accurate guarantee (a lying-small `Content-Length` is still caught mid-stream) without holding the body in memory or blocking streaming handlers. h3's body readers map that overflow to a `413` response; a handler reading `event.req` directly sees the `ERR_BODY_TOO_LARGE` error itself.
 
 An honest `Content-Length` that already exceeds the limit is rejected up-front with a `413`, and a request carrying both `Content-Length` and `Transfer-Encoding` is rejected with a `400` (request smuggling, RFC 7230).
 
--- a/docs/2.utils/1.request.md
+++ b/docs/2.utils/1.request.md
@@ -12,15 +12,19 @@ icon: material-symbols-light:input
 
 ### `assertBodySize(event, limit)`
 
-Asserts that request body size is within the specified limit.
+Asserts that the request body size is within the specified limit.
 
-If body size exceeds the limit, throws a `413` Request Entity Too Large response error.
+The limit is enforced **as the body is read**, not by pre-buffering: the request body stream is wrapped in a byte counter (via {@link limitBodyStream}) that aborts with a `413` Request Entity Too Large error the moment the running total exceeds `limit`. This preserves the byte-accurate guarantee (a lying-small `Content-Length` is still caught mid-stream) without holding the body in memory or blocking streaming handlers.
+
+An honest `Content-Length` that already exceeds the limit is rejected up-front with a `413`, and a request carrying both `Content-Length` and `Transfer-Encoding` is rejected with a `400` (request smuggling, RFC 7230).
+
+Because enforcement is tied to consumption, an overflow on a chunked / unknown-length body surfaces when the handler reads the body rather than as a pre-handler `413`, and a body the handler never reads is never counted.
 
 **Example:**
 
 ```ts
-app.get("/", async (event) => {
-  await assertBodySize(event, 10 * 1024 * 1024); // 10MB
+app.post("/", async (event) => {
+  assertBodySize(event, 10 * 1024 * 1024); // 10MB
   const data = await event.req.formData();
 });
 ```
--- a/docs/2.utils/1.request.md
+++ b/docs/2.utils/1.request.md
@@ -14,7 +14,7 @@ icon: material-symbols-light:input
 
 Asserts that the request body size is within the specified limit.
 
-The limit is enforced **as the body is read**, not by pre-buffering: the request is wrapped by srvx's `limitRequestBody`, which counts bytes as they flow and aborts with a `413` (`ERR_BODY_TOO_LARGE`) the moment the running total exceeds `limit`. This preserves the byte-accurate guarantee (a lying-small `Content-Length` is still caught mid-stream) without holding the body in memory or blocking streaming handlers. h3's body readers map that overflow to a `413` response; a handler reading `event.req` directly sees the `ERR_BODY_TOO_LARGE` error itself.
+The limit is enforced **as the body is read**, not by pre-buffering: the request is wrapped by srvx's `limitRequestBody`, which counts bytes as they flow and aborts with a `413` {@link HTTPError} the moment the running total exceeds `limit` (the error is injected via `createError`). This preserves the byte-accurate guarantee (a lying-small `Content-Length` is still caught mid-stream) without holding the body in memory or blocking streaming handlers.
 
 An honest `Content-Length` that already exceeds the limit is rejected up-front with a `413`, and a request carrying both `Content-Length` and `Transfer-Encoding` is rejected with a `400` (request smuggling, RFC 7230).
 
--- a/docs/2.utils/1.request.md
+++ b/docs/2.utils/1.request.md
@@ -14,7 +14,7 @@ icon: material-symbols-light:input
 
 Asserts that the request body size is within the specified limit.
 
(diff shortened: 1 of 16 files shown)


pi0x:
### Follow-up: independent validation pass

Ran an adversarial review of the change. Results:

- ✅ lint + typecheck + full suite (1596 passed, no unhandled rejections)
- ✅ byte-accurate guarantee intact (lying-small `Content-Length` still caught mid-stream; non-numeric/negative CL safely falls through to the stream limiter — no bypass)
- ✅ `event.url`, `getRequestIP`, `event.runtime`/`waitUntil` all verified working after the `event.req` swap

**Fixed a real defect it surfaced** (2nd commit): the deferred `ERR_BODY_TOO_LARGE` was being masked as `400`/parse-error in two other consumption paths besides `readBody`'s `formData()`:
- `defineValidatedHandler` JSON path (`internal/validate.ts`) → was `400 Invalid JSON body`
- `defineJsonRpcHandler` (`json-rpc.ts`) → was a JSON-RPC parse-error envelope (HTTP 200)

Both now re-throw the overflow so it maps to `413`, via a shared `isBodyLimitError` helper. Regression tests added for both.

**One item left for a maintainer call:** the rebuilt `Request` re-attaches `runtime`/`waitUntil`/`ip`/`context` but not `tls` (srvx mtls plugin). No h3 util reads `event.req.tls`, so a handler reading it *after* `bodyLimit`/`assertBodySize` would see `undefined`. Niche — happy to add `limited.tls = req.tls` if you'd like it covered.

pi0x:
### Update: inlined the limiter, dropped the srvx/body-limit dependency

Two rounds of cleanup on the error handling:

1. Instead of letting srvx's raw `ERR_BODY_TOO_LARGE` propagate and sniffing for it in `response.ts` / `readBody` / validated handlers / JSON-RPC, the limiter now aborts the body stream with a real `413` `HTTPError` at the source. Consumers just re-throw a pre-existing `HTTPError` (general `HTTPError.isError` guard) rather than masking it — no error-code string scattered around, and `response.ts` needs no special case.

2. That first pass wrapped `srvx/body-limit`'s `limitBodyStream` in a *second* stream purely to translate the error — a wasteful double-wrap. Replaced with a single ~15-line pull-based limiter that counts + throws the `HTTPError` directly. This is the **"mirror the helper in h3"** option you listed in the issue comment, and it lets us keep the `HTTPError` shape consistent.

Net effect vs. the earlier commits: **no srvx version bump** (reverted — `srvx/body-limit` is no longer imported), and the **H3Core bundle guardrail is back to its original `7550`/`3000`** (the limiter now lives outside the core `toResponse` path).

Still green: 1596 passed, lint + typecheck clean, no unhandled rejections.

pi0 (approved):

PR #1502, opened 2026-07-22
feat(context): per-event typed H3EventContext

sandros94:
Draft based on exploration of #1501 (includes those commits, as I still have to git good). Also proposed in https://github.com/h3js/h3/issues/1437#issuecomment-5025050049

Commit ad5fcba essentially adds a second param to H3Event that consists in a typed H3EventContext, as well as adds a type argument to the latter for customizing internals.

One such use case is properly typing `params` in something like `defineValidatedHandler`, which might have also coerced its data (eg: an id being number instead) keeping the parity between runtime and types.

This also adds room of improvements for other validated data for `defineValidatedHandler` such as headers and queries which they currently simply gets mutaded in place, going back to their string form instead. A clear followup for this would be to introduce also `event.context.{headers,query}` with the coerced data and type output of the validation schemas (also reducing the number of operations performed by the forward and back mutation).

Leaving as draft till #1501 is merged and this concept is approved, @pi0 no further work is needed in this PR other than your own changes or if you want me to also add the `event.context.{headers,query}` logic or defer to next PR

### DIFF
--- a/src/event.ts
+++ b/src/event.ts
@@ -1,4 +1,4 @@
-import type { ServerRequest, ServerRuntimeContext } from "srvx";
+import type { ServerRequest, ServerRequestContext, ServerRuntimeContext } from "srvx";
 import type { H3EventContext } from "./types/context.ts";
 
 import { EmptyObject } from "./utils/internal/obj.ts";
@@ -29,6 +29,7 @@ export interface HTTPEvent<_RequestT extends EventHandlerRequest = EventHandlerR
 
 export class H3Event<
   _RequestT extends EventHandlerRequest = EventHandlerRequest,
+  _ContextT extends ServerRequestContext = H3EventContext,
 > implements HTTPEvent<_RequestT> {
   /**
    * Access to the H3 application instance.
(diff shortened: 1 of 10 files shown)


sandros94:
ok I'm experimenting a bit more with this and it might be the wrong approach, or worst just partial

sandros94:
Rewrote the typing approach in bc756c5 after a review pass with Fable (credit where credit is due). Also rebased on latest #1501 (including the `getRouterParams` decode fix).

The gist:

1. `H3EventContext` goes back to being a plain interface — `declare module` augmentation keeps working exactly as today (there's now a type test pinning this), no generic on the interface itself
2. the typed view is a separate `TypedH3EventContext<Overrides>` mapped type: overridden keys become _required_, everything else (augmentations included) passes through untouched
3. `EventHandler` gains a trailing, defaulted context generic — non-breaking, so `defineValidatedHandler` keeps using the canonical handler type instead of an inline signature (with potential for downstream custom handlers or the defineRoute changes I proposed)

The part I care most about is coercion parity: with `params: z.object({ id: z.coerce.number() })` the handler sees `id` as a `number` in `event.context.params`, `getRouterParams` and `getRouterParam` (required, no `?.`) and that is actually what's there at runtime. And since the mapped view works for any context key (no per-field type machinery), the same mechanics extend naturally to `event.context.{headers,query}` as the followup, dropping the forward-and-back string mutation entirely.

An `event.validated.{params,query,headers}` view (option 4 from https://github.com/h3js/h3/issues/1437#issuecomment-5025050049) would still be compatible on top of this if we ever want a dedicated namespace, the two aren't mutually exclusive

gulshan:
My (failed) armature attempt to typed context in H3 #1379
May or may not help.

sandros94:
> My (failed) armature attempt to typed context in H3 #1379 May or may not help.

Yes we have many PRs touching or experimenting with it, this one is no different and still mainly an experimentation.

What I do value is to not edit runtime logic and focus only on type DX. srvx is already permissive enough on that aspect, which means that h3 should get an opt-in that acts more on a per-handler basis other than global.

typing H3EventContext globally (for all middlewares and events) is more on the application logic terriroty IMO (which you can already do via `declare module 'h3'` override and a custom runtime plugin).

I also consider #1237 a bit limiting, as it focuses only on typing the event.context by itself. Which is lost upon using the utils such as `getRouterParams` and `getRouterParam` (and query ones once this approach gets approved and I implement them).

Issue #1477, opened 2026-07-18
Response preparation bugs: non-Error throws leak as 200, sync prepare throws bypass hooks, Uint8Array loses content-length


pi0x:
Three related correctness bugs in `src/response.ts` (`toResponse` / `prepareResponse`), found during a pre-stable validation pass and reproduced against current `main` (75ac47d).

## 1. Sync throws inside `prepareResponse` bypass the entire hook pipeline

`src/response.ts:23` — `prepareResponse` is called without try/catch. Anything it throws synchronously (e.g. `JSON.stringify` on a circular return value at `src/response.ts:206`, or circular `error.data` in `errorResponse`) propagates as a raw exception/rejection out of `toResponse` and `~request`:

```ts
app.get("/circular", () => {
  const a: any = {};
  a.self = a;
  return a;
});
// → app.fetch() rejects with raw TypeError
// → onError, onResponse, and silent/console.error logic are all bypassed
// → runtime sees an unhandled rejection instead of a 500
```

## 2. `Uint8Array` return value loses `content-length` (and re-populates `event.res` after the clear)

`src/response.ts:102` clears the prepared response *before* `prepareResponseBody` runs:

```ts
(event as any)[kEventRes] = undefined; // Clear prepared response to avoid duplication
```

The `Uint8Array` branch at `src/response.ts:194` then does `event.res.headers.set("content-length", ...)`, which lazily creates a brand-new `H3EventResponse` that nothing reads — `preparedHeaders` was already captured pre-clear:

```ts
app.get("/bin", () => new Uint8Array([1, 2, 3]));
// → response has no content-length header
```

Side effect: `event[kEventRes]` is left populated again after preparation, partially defeating the duplication fix from #1259 for this branch (a subsequent `toResponse` pass or `onResponse` hook reading `event.res` sees stale state).

PR #1503, opened 2026-07-22
fix(response): route synchronous prepareResponse throws through the error pipeline

g1mn:
Part of #1477 (1 of 2 — split per @pi0's request; companion Uint8Array fix in https://github.com/h3js/h3/pull/1504).

Wraps the `prepareResponse` call in `toResponse` so a **synchronous throw during response preparation** — e.g. `JSON.stringify` on a circular return value, or circular `error.data` inside `errorResponse` — is routed through the normal error pipeline (`toError` → `errorResponse`) instead of escaping as a raw exception/rejection. `onError` / `onResponse` and logging now run, and a `500` is returned. A non-`Error` throw is handled the same way.

```ts
app.get("/circular", () => { const a: any = {}; a.self = a; return a; });
// before: app.fetch() rejects with a raw TypeError, hooks never run
// after:  toError() → errorResponse → 500, onError/onResponse fire
```

Tests (inside `describeMatrix`, via `t.app` / `t.fetch` per the coding guidelines): a circular return value and a thrown `HTTPError` with circular `data` both render as `500` without escaping, and the hook pipeline still fires. `pnpm test` (lint + format + typecheck + full vitest) is green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Synchronous failures during response preparation (including circular/invalid handler return values) are now routed through the standard error pipeline, producing a consistent HTTP 500 response.
  * Synchronous failures inside the error hook are handled safely, avoiding escaped exceptions or infinite recursion and returning the usual normalized error output.
  * Request/error/response hooks continue to run once each even when response preparation fails.
* **Tests**
  * Added regression coverage for circular handler returns and circular error data.
  * Added a regression case for synchronously throwing error hooks.
  * Updated bundle size benchmark limits to allow slightly larger outputs.

### DIFF
--- a/src/response.ts
+++ b/src/response.ts
@@ -139,7 +139,8 @@ function prepareResponse(
     const { onError } = config;
     const errHeaders: Headers | undefined = (event as any)[kEventRes]?.[kEventResErrHeaders];
     return onError && !nested
-      ? Promise.resolve(onError(error, event))
+      ? Promise.resolve()
+          .then(() => onError(error, event))
           .catch((error) => error)
           .then((newVal) => prepareResponse(newVal ?? val, event, config, true))
       : errorResponse(error, config.debug, errHeaders);
--- a/src/response.ts
+++ b/src/response.ts
@@ -25,9 +25,6 @@ export function toResponse(
   try {
     response = prepareResponse(val, event, config);
   } catch (error) {
-    // A synchronous throw while preparing the response (e.g. `JSON.stringify` on a
-    // circular value) must not escape as a raw exception/rejection: route it through
-    // the same error pipeline as a thrown/rejected handler value (onError, logging).
     return toResponse(toError(error), event, config);
   }
   if (typeof (response as PromiseLike<Response>)?.then === "function") {
--- a/src/response.ts
+++ b/src/response.ts
@@ -21,7 +21,15 @@ export function toResponse(
     ) as Promise<Response>;
   }
 
-  const response = prepareResponse(val, event, config);
+  let response: Response | Promise<Response>;
+  try {
+    response = prepareResponse(val, event, config);
+  } catch (error) {
+    // A synchronous throw while preparing the response (e.g. `JSON.stringify` on a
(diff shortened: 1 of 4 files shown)


pi0:
Thanks. If you like to contribute, can you please split per issue to two prs?

g1mn:
Thanks for the quick review! Split as requested:
- #1503 (this PR) — now scoped to just routing synchronous `prepareResponse` throws through the error pipeline.
- https://github.com/h3js/h3/pull/1504 — the `Uint8Array` `content-length` fix.

Also moved the new tests into `describeMatrix` with `t.fetch` per the coding guidelines (thanks @coderabbitai). 🙏

g1mn:
Addressed the review:

- 🔴 **Sync-throwing `onError` recursion** — fixed in 5089638. Wrapped the hook as `Promise.resolve().then(() => onError(error, event))` (matching the `onResponse` pattern), so a synchronous throw becomes a rejection handled by the existing `.catch` and the retry terminates after one attempt instead of re-entering `toResponse`. Added a regression test (*a synchronously-throwing onError does not recurse infinitely*) that stack-overflows / times out on the pre-fix code and returns a normalized 500 after.
- **`describeMatrix` note on `test/errors.test.ts`** — these are already inside the `describeMatrix("errors", …)` block using `t.app` / `t.fetch`; that comment looks like it carried over from the pre-split revision.

Thanks @coderabbitai! 🙏

src/response.ts
```diff
@@ -21,7 +21,15 @@ export function toResponse(
...
   }
 
-  const response = prepareResponse(val, event, config);
+  let response: Response | Promise<Response>;
+  try {
+    response = prepareResponse(val, event, config);
+  } catch (error) {
+    // A synchronous throw while preparing the response (e.g. `JSON.stringify` on a
+    // circular value) must not escape as a raw exception/rejection: route it through
+    // the same error pipeline as a thrown/rejected handler value (onError, logging).
```
pi0:
```suggestion
```

pi0 (approved):

PR #1505, opened 2026-07-22
fix(response): route synchronous prepareResponse throws through the error pipeline

g1mn:
Part of #1477 (1 of 2 — split per @pi0's request; companion Uint8Array fix in https://github.com/h3js/h3/pull/1504).

Wraps the `prepareResponse` call in `toResponse` so a **synchronous throw during response preparation** — e.g. `JSON.stringify` on a circular return value, or circular `error.data` inside `errorResponse` — is routed through the normal error pipeline (`toError` → `errorResponse`) instead of escaping as a raw exception/rejection. `onError` / `onResponse` and logging now run, and a `500` is returned. A non-`Error` throw is handled the same way.

```ts
app.get("/circular", () => { const a: any = {}; a.self = a; return a; });
// before: app.fetch() rejects with a raw TypeError, hooks never run
// after:  toError() → errorResponse → 500, onError/onResponse fire
```

Tests (inside `describeMatrix`, via `t.app` / `t.fetch` per the coding guidelines): a circular return value and a thrown `HTTPError` with circular `data` both render as `500` without escaping, and the hook pipeline still fires. `pnpm test` (lint + format + typecheck + full vitest) is green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling when response preparation fails, including circular data and HTTP errors.
  * Synchronously throwing error hooks are now handled safely without escaping or causing recursion.
  * Requests now consistently return a normalized 500 error response when unrecoverable errors occur.
  * Ensured request, error, and response hooks are invoked correctly during failure scenarios.

* **Tests**
  * Added regression coverage for response preparation and hook error handling.
  * Updated bundle-size benchmark thresholds.

### DIFF
--- a/src/response.ts
+++ b/src/response.ts
@@ -139,7 +139,8 @@ function prepareResponse(
     const { onError } = config;
     const errHeaders: Headers | undefined = (event as any)[kEventRes]?.[kEventResErrHeaders];
     return onError && !nested
-      ? Promise.resolve(onError(error, event))
+      ? Promise.resolve()
+          .then(() => onError(error, event))
           .catch((error) => error)
           .then((newVal) => prepareResponse(newVal ?? val, event, config, true))
       : errorResponse(error, config.debug, errHeaders);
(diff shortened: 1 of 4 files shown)


g1mn:
Duplicate of #1503, which was already merged. Closing.

PR #1506, opened 2026-07-22
feat(static): support HTTP Range requests in serveStatic

### LINKED ISSUE #1267: Feature request: HTTP Range support for serveStatic
### Describe the feature

Describe the feature

serveStatic currently ignores HTTP Range requests. When a client sends a Range header, the server still responds with the full response body and does not include partial-content headers. This prevents:

Download resumption

Media seeking in browsers and video/audio players

Proper CDN and proxy behavior

This feature request proposes adding byte range support to serveStatic for GET and HEAD requests when the asset size is known.

Proposed behavior

Always advertise Accept-Ranges: bytes

Support only a single byte range

Ignore malformed or multi-range requests and serve the full response

For valid ranges:

Respond with 206 Partial Content

Include Content-Range and correct Content-Length

For HEAD, send headers only

For unsatisfiable ranges:

Respond with 416 Range Not Satisfiable

Include Content-Range: bytes */<size>

Content-Length: 0

Important constraints

Preserve existing caching, encoding selection, fallthrough behavior, and 404/405 handling

Do not introduce new public exports

Keep all helper logic private within the serveStatic module

Use cases

Video/audio streaming

Resumable file downloads

Browser compatibility

Better support for media players and HTTP clients

Additional information

This behavior aligns with RFC 7233 and matches what users expect from static file servers. Most modern browsers automatically send range requests for media, and lack of support breaks playback and seeking.

No API changes are required—this is a backward-compatible internal enhancement to serveStatic.

### Additional information

g1mn:
Closes #1267.

Adds single-range byte support to `serveStatic` for `GET`/`HEAD` requests when the asset size is known (`meta.size`), following the behavior proposed in the issue:

- Always advertises `accept-ranges: bytes` when size is known
- Supports only a single byte range (`bytes=start-end`, `bytes=start-`, `bytes=-suffix`); malformed or multi-range headers are ignored and the full response is served
- A satisfiable range returns `206 Partial Content` with `content-range`/`content-length`; `HEAD` returns headers only
- An out-of-bounds range returns `416 Range Not Satisfiable` with `content-range: bytes */<size>` and `content-length: 0`
- No interaction with existing caching (`if-none-match`/`if-modified-since`), encoding selection, `fallthrough`, or 404/405 handling — all untouched
- No new public exports; range parsing/slicing lives entirely inside `src/utils/static.ts`

```ts
app.get("/video.mp4", (event) => serveStatic(event, { getMeta, getContents }));
// Range: bytes=0-1023  -> 206, content-range: bytes 0-1023/<size>
// Range: bytes=999999- -> 416, content-range: bytes */<size>  (if out of bounds)
```

Content is sliced for `string`/`Uint8Array`/`ArrayBuffer`/`Blob` return values from `getContents`. Any other body type (e.g. a `ReadableStream`) can't be sliced without fully buffering it, so it falls back to a full `200` response rather than silently mis-serving a range.

Tests (`describeMatrix`, via `t.app`/`t.fetch`): start-end/open-ended/suffix ranges, clamping past the end and oversized suffix ranges, 416 for out-of-bounds/zero-length-suffix/inverted ranges, malformed and multi-range headers ignored, no Range support when size is unknown, and per-body-type slicing (string/Uint8Array/ArrayBuffer/Blob) plus the stream fallback. `pnpm test` (lint + format + typecheck + full vitest + coverage) is green; the only failure locally (`getRequestFingerprint` IP match) is a pre-existing, unrelated sandbox-networking flake reproducible on `main` without this change.

## Summary by CodeRabbit

* **New Features**
  * Added support for single-byte range requests when serving static assets.
  * Valid range requests now return partial content with appropriate range and length headers.
  * Added support for open-ended and suffix ranges, including `HEAD` requests.
  * Static assets advertise byte-range support when their size is known.

* **Bug Fixes**
  * Unsatisfiable ranges now return `416` responses.
  * Unsupported or unsliceable content safely falls back to a complete response.

### DIFF
--- a/src/utils/static.ts
+++ b/src/utils/static.ts
@@ -194,19 +194,118 @@ export async function serveStatic(
     event.res.headers.set("content-encoding", meta.encoding);
   }
 
-  if (meta.size !== undefined && meta.size > 0 && !event.res.headers.get("content-length")) {
-    event.res.headers.set("content-length", meta.size + "");
+  // Range requests need a known, whole-asset size to validate/compute against.
+  let range: { start: number; end: number } | undefined;
+  const size = meta.size !== undefined && meta.size > 0 ? meta.size : undefined;
+  if (size !== undefined) {
+    event.res.headers.set("accept-ranges", "bytes");
+    const rangeHeader = event.req.headers.get("range");
+    if (rangeHeader) {
+      const parsed = parseRange(rangeHeader, size);
+      if (parsed === "unsatisfiable") {
+        return new HTTPResponse(null, {
+          status: 416,
+          headers: { "content-range": `bytes */${size}`, "content-length": "0" },
+        });
+      }
+      range = parsed;
+    }
+  }
+
+  if (range) {
+    event.res.headers.set("content-range", `bytes ${range.start}-${range.end}/${size}`);
+    event.res.headers.set("content-length", range.end - range.start + 1 + "");
+  } else if (size !== undefined && !event.res.headers.get("content-length")) {
+    event.res.headers.set("content-length", size + "");
   }
 
   if (event.req.method === "HEAD") {
-    return new HTTPResponse(null, { status: 200 });
+    return new HTTPResponse(null, { status: range ? 206 : 200 });
   }
 
   const contents = await options.getContents(id);
+  if (range) {
+    const sliced = sliceBody(contents, range.start, range.end);
+    if (sliced !== undefined) {
+      return new HTTPResponse(sliced, { status: 206 });
+    }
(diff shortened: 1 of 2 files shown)


pi0:
Thanks for PR. But i prefer to defer and do this one myself with more care. in the meanwhile, srvx 0.12 landed full range support for serving `public/` assets

g1mn:
Totally understand, thanks for the quick look! Good to know srvx 0.12 already covers range support for `public/` assets — makes sense to keep this consolidated on your end rather than layering another implementation in `serveStatic`. Happy to help elsewhere if useful.

PR #1511, opened 2026-07-24
fix(response): defer stream metadata until first chunk

### LINKED ISSUE #1510: [v2] Regression: Response sended before first stream payload
### Environment

- h3: 2.0.1-rc.25
- node: 24.18.0

## v2.0.1-rc.25 (Broken status/headers)

```tsx
import { H3, serve } from "h3";

const app = new H3().get("/", async (event) => {
  return ReadableStream.from((async function* () {
    event.res.status = 201;
    event.res.headers.set("hello", "world");

    yield "Hello World";
  })())
});

serve(app, { port: 3000 });
```

## v1.15.11 (Working status/headers)

```ts
import { setResponseHeader, setResponseStatus } from "h3";
import { createApp, defineEventHandler, sendStream, toNodeListener } from "h3";
import { createServer } from "http";

const app = createApp();
app.use("/", defineEventHandler(async (event) => {
  return ReadableStream.from((async function* () {
    setResponseStatus(event, 201);
    setResponseHeader(event, "hello", "world");

    yield "Hello World";
  })())
}));

createServer(toNodeListener(app)).listen(3000)
```

### Describe the bug

h3 v1 waited with sending the response until the first async stream payload. This allowed setting response status codes and headers from within a stream (as long as it was part of the first payload). In h3 v2 you must set response status codes and headers before returning the stream. Response changes made while async rendering the first stream payload are now ignored. This makes it impossible to set response status/headers depending on async data within stream based frameworks such as SolidStart.

### Additional context

Related: solidjs/solid-start#2229

### Logs

```sh

```

MFA-G:
### 🔗 Linked issue

Fixes #1510

### 📚 Description

Wait for a returned `ReadableStream` to produce its first chunk before preparing the HTTP response. This restores the v1 behavior needed by async streaming renderers: status and headers set while producing the first chunk are included before the runtime commits the response.

The stream is teed so the probed chunk remains available to the response consumer, while cancellation continues to propagate to the source. A matrix regression test covers both web and Node adapters.

### 🧪 Validation

- `pnpm vitest run test/stream.test.ts test/dispose.test.ts --typecheck.enabled=false` (31 passed, 3 skipped)
- `pnpm exec oxlint src/response.ts test/stream.test.ts`
- `pnpm exec oxfmt --check src/response.ts test/stream.test.ts`
- `pnpm typecheck`
- Full `pnpm test`: 1,610 passed, 44 skipped; one environment-specific IP assertion failed (`::ffff:127.0.0.1` vs expected loopback forms), plus the three bundle-size budgets increased by ~306–320 raw bytes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved streaming responses so status codes and headers are finalized after the first content chunk is ready.
  * Ensures clients receive accurate response metadata alongside streamed content.

* **Tests**
  * Added coverage verifying delayed stream initialization, response metadata, and streamed body content.

### DIFF
--- a/src/response.ts
+++ b/src/response.ts
@@ -108,6 +108,7 @@ function prepareResponse(
   event: H3Event,
   config: H3Config,
   nested?: boolean,
+  primed?: boolean,
 ): Response | Promise<Response> {
   if (val === kHandled) {
     return new FastResponse(null);
@@ -141,6 +142,12 @@ function prepareResponse(
           .catch((error) => error)
           .then((newVal) => prepareResponse(newVal ?? val, event, config, true))
       : errorResponse(error, config.debug, errHeaders);
+  }
+
+  // Delay preparing the response until the stream produces its first chunk. This
+  // lets async renderers set status and headers before the runtime commits them.
+  if (!primed && val instanceof ReadableStream) {
+    return primeStream(val).then((stream) => prepareResponse(stream, event, config, nested, true));
   }
 
   // Only set if event.res.headers is accessed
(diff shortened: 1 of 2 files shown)


pi0:
AI

PR #1515, opened 2026-08-04
fix(session): encode large payloads outside node-compatible runtimes

### LINKED ISSUE #1514: base64Encode throws RangeError for large payloads on runtimes without a global Buffer
### Environment

h3 `2.0.1-rc.26` (`main` @ 502c59f), reproduced on Node 25.8.2. Affects any runtime where `globalThis.Buffer` is absent — the `h3/deno`, `h3/service-worker` and browser entries, and workerd without `nodejs_compat`.

### Reproduction

`src/utils/internal/encoding.ts` has two `base64Encode` implementations: `Buffer.from(buff).toString("base64url")` when a global `Buffer` exists, and a hand-rolled fallback otherwise. The fallback ends with:

```js
return String.fromCharCode(...bytes);
```

`bytes` holds one element per output character, so the spread passes ~1.33 arguments per input byte and overflows the engine's argument limit:

```js
// node --input-type=module
const realBuffer = globalThis.Buffer;
globalThis.Buffer = undefined; // emulate deno / service-worker / browser
const { base64Encode } = await import("h3/dist/…/encoding.js"); // internal helper

base64Encode(new Uint8Array(1024));      // ok
base64Encode(new Uint8Array(128 * 1024)); // RangeError: Maximum call stack size exceeded
globalThis.Buffer = realBuffer;
```

Bisecting the boundary on Node 25.8.2 gives the last working input size as **90,185 bytes** (≈120k output characters). The limit is a function of the available stack, so a runtime with a smaller stack fails well below that.

### Describe the bug

`sealSession()` base64-encodes the entire sealed session in one call, so this is reachable from `useSession()` / `updateSession()` — not just from the internal helper — and it triggers inside the size range h3 itself permits:

- a chunked session cookie may hold up to `MAX_CHUNKED_COOKIE_COUNT` (100) chunks of ~4KB ≈ 400KB, and
- a session read from the `x-{name}-session` header has no size cap at all.

So on Deno, a service worker, or a browser build, storing a session larger than roughly 64–90KB throws `RangeError: Maximum call stack size exceeded` from inside `updateSession` instead of writing the cookie. On Node and Bun the `Buffer` path is taken and the bug is invisible, which is why it has not surfaced: every existing case in `test/unit/encoding.test.ts` runs on Node, so the fallback branch is never executed (it sits at ~46% line coverage), and the one test that does stub `Buffer` — `test/unit/iron-crypto.test.ts` — is `it.skip`ped with a note that mocking `Buffer` destabilises vite internals.

Encoding and decoding otherwise agree between the two implementations: I swept every input length 0–40 with random bytes and both round-trip identically, and tokens sealed under one path unseal correctly under the other. Size is the only divergence.

Happy to send a PR — building the string in fixed-size chunks fixes it and leaves the `Buffer` path untouched.

Jaybhade:
resolves #1514

### What

`base64Encode` falls back to a hand-rolled encoder when there is no global `Buffer` — the `h3/deno`, `h3/service-worker` and browser entries. That fallback ended with `String.fromCharCode(...bytes)`, spreading one argument per output character, so it throws `RangeError: Maximum call stack size exceeded` once the payload passes the engine's argument limit. Bisected on Node 25.8.2, the last input size that works is **90,185 bytes**; the threshold depends on the available stack, so a runtime with a smaller one fails lower.

This is reachable from public API, not just the internal helper: `sealSession()` base64-encodes the whole sealed session, and the sizes h3 itself allows go well past the limit — a chunked session cookie holds up to `MAX_CHUNKED_COOKIE_COUNT` (100) × ~4KB ≈ 400KB, and sessions read from the `x-{name}-session` header are uncapped. On Deno or a service worker, `updateSession` throws instead of writing the cookie.

### How

Build the string in fixed-size chunks (8192 characters per `String.fromCharCode` call). The `Buffer` path is untouched, so Node and Bun keep the exact same code path and output.

### Tests

Three cases added to `test/unit/encoding.test.ts`, under a `describe` that swaps `globalThis.Buffer` out **synchronously** — no `await` runs while the global is missing, which is what destabilises vite internals in the `it.skip`ped `iron-crypto` case, so these run reliably:

| test | against unpatched `main` |
| --- | --- |
| encodes payloads larger than the engine argument limit | **fails** — `RangeError: Maximum call stack size exceeded` at `encoding.ts:47` |
| agrees with the Buffer implementation for every input length | passes (regression guard for the two implementations) |
| decodes base64url produced by the Buffer implementation | passes (regression guard) |

The failing one is the regression test for this bug; I ran all three against unpatched `lib` first to confirm only that case is red.

The two guard cases pin behaviour this touches but does not change — I separately swept every input length 0–40 with random bytes and confirmed the fallback and `Buffer` paths produce byte-identical output, and that a token sealed under either path unseals under the other.

### Verification

```
oxlint .                  clean
oxfmt --check .           All matched files use the correct format
tsc --noEmit              no errors
vitest --run --coverage   56 files, 1627 passed | 44 skipped   (baseline: 1624 passed)
```

`src/utils/internal/encoding.ts` line coverage goes from 48% to 100%.

I left `CHANGELOG.md` and the version alone.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Base64 encoding reliability for large inputs.
  * Preserved consistent encoding and decoding behavior when native buffering support is unavailable.
  * Maintained compatibility with standard Base64URL output.

* **Tests**
  * Added coverage for large payloads, varied input sizes, fallback encoding, round-trip conversions, and Base64URL compatibility.

### DIFF
--- a/src/utils/internal/encoding.ts
+++ b/src/utils/internal/encoding.ts
@@ -7,15 +7,7 @@ https://github.com/denoland/std/blob/main/LICENSE
 export const textEncoder: TextEncoder = /* @__PURE__ */ new TextEncoder();
 export const textDecoder: TextDecoder = /* @__PURE__ */ new TextDecoder();
 
-// ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_
-const base64Code = [
-  65, 66, 67, 68, 69, 70, 71, 72, 73, 74, 75, 76, 77, 78, 79, 80, 81, 82, 83, 84, 85, 86, 87, 88,
-  89, 90, 97, 98, 99, 100, 101, 102, 103, 104, 105, 106, 107, 108, 109, 110, 111, 112, 113, 114,
-  115, 116, 117, 118, 119, 120, 121, 122, 48, 49, 50, 51, 52, 53, 54, 55, 56, 57, 45, 95,
-];
-
-// Number of characters emitted per `String.fromCharCode` call.
-const ENCODE_CHUNK_SIZE = 8192;
+const base64Chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_";
 
 export function base64Encode(data: ArrayBuffer | Uint8Array | string): string {
   const buff = validateBinaryLike(data);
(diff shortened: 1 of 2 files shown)


Jaybhade:
Thanks — one of the two was a real hole in the test, fixed in d6c4f3c.

**Fixture built after `Buffer` was removed** — correct, and the test was lying about what it covered. `base64Encode` was being called *inside* `withoutBuffer`, so the fixture came from the fallback encoder and the case only round-tripped the fallback against itself. The encode now happens while `Buffer` is still present:

```ts
const input = new Uint8Array([251, 239, 190, 0, 1, 2]);
const encoded = base64Encode(input);
expect(withoutBuffer(() => base64Decode(encoded))).toEqual(input);
```

Those bytes encode to `----AAEC`, so the fallback decoder's `-`/`_` → `+`/`/` substitution is actually exercised. Confirmed it's load-bearing by mutating that line out of `base64Decode` — the case now fails, where before the fix it passed.

**`describeMatrix` for the Buffer-free block** — I don't think this one applies. `describeMatrix` boots an `H3` app plus a Node HTTP server per context to run request handling against the web and node targets; it isn't a general cross-runtime harness. `base64Encode`/`base64Decode` are pure functions imported straight from `src/utils/internal/encoding.ts`, with no event, app or server involved, and the branch under test is selected by `globalThis.Buffer` rather than by the target — which the suite already controls directly. It's also not the convention here: all 22 files in `test/unit/` use plain `describe`, and `describeMatrix` is used only by the server-level tests in `test/`. Wrapping this block would double the runtime and add a server per run without changing what's covered. Happy to switch if you'd rather it be uniform.

`vitest run` is green (56 files, 1627 passing) and `pnpm lint` is clean.

pi0 (approved):

pi0 (approved):
Thanks

PR #1550, opened 2026-09-15
types(auth): basicAuth context fields are always set

hxperl:
`H3EventContext["basicAuth"]` declares all three inner fields optional, but there is exactly one write site — `src/utils/auth.ts:104` — and every successful path reaches it:

```ts
context.basicAuth = { username, password, realm };
```

Two `slice` results and a `??` fallback. All three are always present, always strings; every other path throws. So the optional markers describe a state the runtime never produces. The two `@example` blocks in the same file disagree with each other about it, which is what led me here — line 45 uses `basicAuth.username`, line 115 uses `basicAuth?.username`.

Against `main` at bce4ae8, reading the credentials back:

```
repro.ts(11,9): error TS2322: Type 'string | undefined' is not assignable to type 'string'.
repro.ts(12,9): error TS2322: ...
repro.ts(13,20): error TS2345: ...
```

All three are false. Driving the real handler — including the `validate` paths that permit empty credentials per RFC 7617 — always yields three strings; empty is still a `string`, never `undefined`.

The three inner fields become required. The property itself stays optional, because it genuinely is absent until a request authenticates. `test/unit/types.test-d.ts` gets a block covering both halves; it fails on `main` and passes here.

**The judgement call, and it is not free.** Writing a *partial* object into the context stops compiling:

```
compat.ts(6,1): error TS2739: Type '{ username: string; }' is missing password, realm
```

I think rejecting it is the point — nothing in h3 writes one — but it could appear in a test that mocks an authenticated event. If you would rather keep that working, the alternative is to leave the type alone and fix the `requireBasicAuth` example to use `?.` instead. Your call. I considered separate read/write types and rejected it: a partial write would really sit in the context, so the narrow read type would be a lie.

**Adjacent, not bundled:** even with this, `requireBasicAuth`'s own `@example` still does not compile, because `basicAuth` is optional and TS cannot narrow it from an `await`ed call. Fixing that is an API decision that is yours — happy to file an issue. Also, open PR #1502 touches `src/types/context.ts`; different concern, but whichever lands second may need a trivial rebase.

Verified on Node 24.21.0 / macOS arm64 / TS 7.0.2: `pnpm lint` and `pnpm typecheck` clean, `pnpm test` 2811 passed with one failure — `test/body-limit.test.ts`, a 5s timeout that reproduces on unmodified `main` and names a different test each run.

I did not open an issue first as the template asks; the reproduction is inline so the PR stands alone, but say if you would still like one.

## Summary by CodeRabbit

* **Documentation**
  * Clarified the authentication context fields available after successful basic authentication.
  * Username and realm are now documented as required strings, while password remains optional.

* **Tests**
  * Added type coverage to verify authentication context behavior before and after authentication.


src/types/context.ts
```diff
@@ -26,11 +26,11 @@ export interface H3EventContext extends ServerRequestContext {
...
 
-  /* Basic authentication data */
+  /* Basic authentication data, set by `requireBasicAuth` once credentials are accepted */
   basicAuth?: {
-    username?: string;
-    password?: string;
-    realm?: string;
+    username: string;
+    password: string;
+    realm: string;
```
pi0, line 33: not all 3 are required

PR #1551, opened 2026-09-15
docs: document raw query access via event.url.search

### LINKED ISSUE #1255: Support nested query or get raw query or both...
### Describe the feature

I have a nested query like `filter[color]=red&filter[size]=large&user[product][id]=385&sort=-createdAt&include=author,comments&active=true`. I want to get the raw query to transform the query into object with `qs` lib or I would like to see this on the getQuery.

An example how I use `qs` lib:
```js
const query = "?filter[color]=red&filter[size]=large&user[product][id]=385&sort=-createdAt&include=author,comments&active=true";
      
      const queryObject = qs.parse(query, {
        ignoreQueryPrefix: true,
        decoder: (str: string, defaultDecoder: qs.defaultDecoder, charset: string, type: "key" | "value") => {    
          if (type === "value" && !!Number(str)) {
            return Number(str);
          }

          if (type === "value" && str === "true") {
            return true;
          }
          
          if (type === "value" && str === "false") {
            return false;
          }

          if (type === "value" && str.includes(",")) {
            return str.split(",").map((s: string) => s.trim());
          }

          return defaultDecoder(str);
        }
      });
```

Edit: One way to get the raw query is: `const rawQuery = getRequestURL(event).search;`
If I had the `getRawQuery` it will be better...

### Additional information

OskarLebuda:
Resolves https://github.com/h3js/h3/issues/1255

## Summary by CodeRabbit

- **Breaking Changes**
  - Removed the exported `getRawQuery` helper.
  - Access raw, unparsed query strings through `event.url.search`.

- **Documentation**
  - Updated `getQuery` guidance to clarify raw query access and support custom parsing of nested queries.
  - Updated examples to show the raw query string, including its leading `?`.


src/utils/request.ts
```diff
@@ -98,6 +98,19 @@ export function toRequest(
...
+ * Get the raw query string from the request URL.
+ *
+ * @example
+ * app.get("/", (event) => {
+ *   const query = getRawQuery(event); // "key=value&key2=value1&key2=value2"
+ * });
+ */
+export function getRawQuery(event: HTTPEvent): string {
+  const url = (event as H3Event).url || new URL(event.req.url);
+  return url.search.slice(1);
```
pi0: I think we should just document using `event.url.search` for accessing raw query it is unnecessary wrapper util.

OskarLebuda: You're right. I've updated it to just a doc.

pi0 (approved):

PR #1553, opened 2026-09-16
fix(cookie): include partitioned in the distinct-cookie key

Kaue-TecsaGroup:
## Summary

`_getDistinctCookieKey()` (in `src/utils/cookie.ts`) decides whether a `setCookie`/`deleteCookie` call replaces an existing queued `Set-Cookie` header for the same response, keyed by `[name, domain, path]` only.

A [partitioned (CHIPS)](https://developer.mozilla.org/en-US/docs/Web/Privacy/Guides/Privacy_sandbox/Partitioned_cookies) cookie is stored in a **separate browser cookie jar** from an unpartitioned cookie sharing the same name/domain/path. Treating them as the same key is incorrect: whichever `setCookie`/`deleteCookie` call happens last in a given response silently evicts the other's header from `event.res.headers`, even though a real browser would keep both as two independent cookies.

### Where this actually bites

The common real-world trigger is migrating an existing cookie to add `partitioned: true` (e.g. adopting CHIPS for a cross-site/embedded context) while also needing to clear the old, pre-migration unpartitioned cookie still sitting in returning users' browsers:

```ts
// Clear the legacy, pre-CHIPS cookie...
deleteCookie(event, "session", { sameSite: "lax" })
// ...and set today's partitioned one.
setCookie(event, "session", token, { sameSite: "none", secure: true, partitioned: true })
```

Today, only **one** of these two `Set-Cookie` headers survives in the response — `_getDistinctCookieKey` sees them as the same `session;;/ ` key and the second call's `event.res.headers.delete("set-cookie")` + rebuild silently drops the first. Which one wins depends purely on call order, not on the two cookies actually being distinct in the browser. We hit this directly: a read-only request that only issued the legacy-clearing `deleteCookie` (no `setCookie` in that same response) sent that as the *sole* `Set-Cookie` header — on a client that doesn't fully isolate `Partitioned` cookies, this can end up clearing the real, working session cookie outright.

### Fix

Add `partitioned` to `_getDistinctCookieKey`'s key, alongside `name`/`domain`/`path`. Two cookies with the same name/domain/path but different `partitioned` values are now correctly treated as distinct, non-colliding `Set-Cookie` headers. Cookies with the same `partitioned` value (the overwhelmingly common case, `undefined`/`false` on both sides) keep deduplicating exactly as before.

## Test plan

- [x] Added `does not deduplicate cookies that differ only by \`partitioned\`` — asserts both headers are present, with `Partitioned` on the new-cookie header only.
- [x] Added `still deduplicates cookies with the same \`partitioned\` value` — asserts the existing merge/replace behavior is unchanged.
- [x] `pnpm vitest run test/cookies.test.ts` — 120/120 passing (was 118 before this PR's 2 new cases).
- [x] `pnpm oxlint` / `pnpm oxfmt --check` on both changed files — clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed cookie handling so partitioned and unpartitioned cookies with the same name, domain, and path are preserved as separate entries.
  - Ensured cookies with matching partition settings continue to be deduplicated correctly.


pi0 (approved):
Thanks!

PR #1555, opened 2026-09-17
fix(sse): preserve buffered events during overlapping flushes

### LINKED ISSUE #1554: Buffered SSE events can be duplicated or lost during overlapping flushes
### Environment

H3 main at `a8a803beb98613677317a2bc5fe0a3017fd98db8` (`2.0.1-rc.32` in package.json), Node.js 26.7.0, macOS arm64.

### Reproduction

From an installed checkout of that revision, run:

```sh
node --input-type=module <<'JS'
import { EventStream, mockEvent } from './src/index.ts';

const stream = new EventStream(mockEvent('/'));
stream.pause();
await stream.push('first');

// Neither write can finish until the readable side starts being consumed.
const firstFlush = stream.flush();
const secondFlush = stream.flush();
const text = new Response(await stream.send()).text();
await Promise.all([firstFlush, secondFlush]);
await stream.close();
console.log(JSON.stringify(await text));
JS
```

Expected: `"data: first\n\n"`. Actual: `"data: first\n\ndata: first\n\n"`.

There is also a data-loss case: replace `const secondFlush = stream.flush()` with `await stream.push('second')` and only await `firstFlush` before closing. Expected output contains `first` and `second`, but actual output only contains `first`. The second event was accepted while paused, then erased when the first flush completed.

### Describe the bug

`EventStream.flush()` keeps `_unsentData` in the shared buffer while awaiting `writer.write()`. When a client has not yet consumed the response, another flush can write that same buffer again. A paused producer can also append data during the wait, which the first flush's final assignment discards.

This also happens when `resume()` overlaps `close()`, because both call `flush()`. The regression reproduces through both H3's web response path and its Node HTTP adapter.

I have prepared a small fix that detaches the current batch before awaiting its write, with regression tests, and will submit the PR. The writer continues to provide write ordering and backpressure.

### Additional context

This is separate from the teardown behavior fixed in #1484. Four new regression cases fail on the pinned base (duplicate flush, lost buffered data, and resume/close through web and Node); they pass with the fix. Cancellation remains covered without mocked stream writes.

AI assistance was used to investigate and prepare the fix and tests.

maximilliangrand:
Resolves #1554.

When a buffered SSE write is waiting for a reader, a second `flush()` can send the same events again. A paused `push()` during that wait can instead be erased when the first flush finishes. `resume()` overlapping `close()` also duplicates buffered events through both the web and Node adapters.

Detach the batch from `_unsentData` before awaiting its write. This gives later pushes their own buffer and leaves write ordering/backpressure to the existing `WritableStream`. Closing drains any newly buffered batch through a single shared close operation. Unconsumed streams are cancelled on disposal so pending writes do not prevent cleanup.

Regression coverage uses real streams for overlapping flushes, buffered single/batch/comment pushes, cancellation, and resume/close through the web/Node test matrix. The four original regression failures were confirmed on the base revision before changing the implementation. The concurrent-close data-loss regression also fails on `d32e9fa` before the follow-up fix. Cancellation controls remain covered, including overlapping close calls; HEAD response disposal is covered as well.

Validation on Node.js 26.7.0, macOS arm64, pnpm 12.3.4:

- SSE suites: 51 cases passed as part of the full suite.
- `pnpm test`: lint, formatting, typecheck, and all 2,838 tests passed; 48 skipped.
- `pnpm build`: passed.
- Additional native-stream stress checks preserved exact ordering over 100 overlapping flush/resume/pause/push/close sequences and settled cancellation with one close callback.

An initial full run (and the first follow-up full run) timed out in the existing, unrelated `body-limit` test for an unread chunked request. That file passed in isolation and the complete `pnpm test` rerun passed unchanged.

AI assistance was used for implementation, regression tests, and an independent review of ordering and shutdown behavior.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed streaming issues that could duplicate or lose buffered server-sent events during concurrent resume, close, or flush operations.
  - Ensured events added while a stream is paused or closing are preserved and delivered in order.
  - Improved handling when readers cancel during concurrent flushing.
  - Improved stream cleanup when response bodies are dropped or writes remain pending.

- **Tests**
  - Added coverage for concurrent resume, close, flushing, backpressure, and cancellation scenarios.


maximilliangrand:
Addressed the concurrent-close finding in `b4e6524`, preserving pi0's draining and disposal changes. One shared close promise now owns the drain loop and the writer close.

The new native-stream regression reproduces silent loss of the second event on `d32e9fa` when two closes overlap a late paused push. It passes with the fix; a cancellation control verifies both callers settle and `onClosed` runs once. `pnpm test` passes all 2,838 tests (48 skipped), lint/format/type checks, and `pnpm build` passes. The same pre-existing body-limit timeout occurred in the first full run; the unchanged rerun passed.

pi0 (approved):

